PR #11, opened 2017-04-05
Update dokan.js

MamunHoque:
There vendor profile update (social setting ) form does not show validation error message. So this comment will fix this issue.


alamgircsebd:
No longer response on this request and the changes implemented on main file, which one need to do on source file `assets/src/js/script.js`

PR #33, opened 2017-06-23
Spanish translation

Sophivorus:
Come on guys, it's the third time I send this patch. Now it's easier to review than ever. Get it merged already.


nayemDevs:
@Sophivorus, I am really sorry to say that we are unable to merge this pull request because in Dokan plugin we only provide the pot file and you have to translate the POT file to your own language and then save the translation file with your own language and country code like as dokan-lite-bn_BD.po/mo. 

Now, if you want to translate then you can add the translation here- https://translate.wordpress.org/projects/wp-plugins/dokan-lite and that would be more easier.

Sophivorus:
Ok. I've added my translations to https://translate.wordpress.org/projects/wp-plugins/dokan-lite, but I think they need approval.

PR #73, opened 2017-10-30
Fix the order status issue.

yipchunmun:
- Fix the issue where the completed sub-order status does not update the parent order status.


yipchunmun:
- Prompt a confirmation dialog box to seller to confirm the order status. 
- To refresh the page automatically after seller updated the status to prevent seller from changing the order status again (refunded & cancelled). For example, if a seller update the status from processing to cancelled and click edit again to update the order status to completed, it will cause the calculation of the sales order to be incorrect.

PR #118, opened 2018-01-14
Feature: Store opening time/closing time widget is fixed #113

### LINKED ISSUE #113: Store opening time/closing time widget

saimonh3:


ManigandanRaamanathan:
May I know the conflict that this widget has after last fix?

ManigandanRaamanathan:
@saimonh3  Can you change the exsiating time picker (on comit 
895d4fb) for both start time and close time to actual time picker ui with time formats instead of drop down options? 
Current drop down timer option produces a fixed format which is hh:mm tt. This does not provide a complete (timestamp) format. There will be an issue if the user wants the system to perform calculation to identify whether the store is currently open or not. Can you please update the fixed option to timestamp format option?

ManigandanRaamanathan:
@saimonh3   need validation between start time and end time. The user should not set the end time below or same as the start time. Can you please add the validation part

PR #156, opened 2018-02-05
Balance is greater than actual sales is fixed #155

### LINKED ISSUE #155: Balance is greater than actual sales

saimonh3:


includes/functions.php
```diff
@@ -390,20 +390,25 @@ function dokan_generate_sync_table() {
...
+     *
+     * @param WC_ORDER $order
+     *
+     * @param int $seller_id
+     *
+     * @return int $earned
+     */
+    function dokan_get_seller_earnings_by_order( $order, $seller_id ) {   
+        $commission_recipient = dokan_get_option( 'extra_fee_recipient', 'dokan_general', 'seller' );
+        $earned = $order->get_total() - ( $order->get_total_shipping() + $order->get_total_tax() + dokan_get_admin_commission_by( $order, $seller_id ) );
```
tareq1988: Move this line to an else block below, otherwise if the commission recipient is admin, we are calculating some unnecessary things.

PR #197, opened 2018-03-11
Dokan Vendor Store URL Availability fixed #196

### LINKED ISSUE #196: Dokan Vendor Store Name Availability
the vendor store name always show "available" even when the store name has taken, but generate some number in link (ie. if **dummy** already taken and someone else use **dummy** as store name, the link will become **dummy-2**) can you fix the availability of store name so it won't regenerate number on link

Sk-Shaikat:


classes/admin-user-profile.php
```diff
@@ -127,7 +127,7 @@ function add_meta_fields( $user ) {
                 <tr>
                     <th><?php esc_html_e( 'Store URL', 'dokan-lite' ); ?></th>
                     <td>
-                        <input type="text" name="dokan_store_url" class="regular-text" id="seller-url" value="<?php echo esc_attr( $user->data->user_nicename ); ?>"><strong id="url-alart-mgs"></strong>
+                        <input type="text" name="dokan_store_url" vendor="<?php echo $user->ID; ?>" class="regular-text" id="seller-url" value="<?php echo esc_attr( $user->data->user_nicename ); ?>"><strong id="url-alart-mgs"></strong>
```
tareq1988: You are using a custom attribute for vendor which is non-standard. Use a data property instead.

classes/ajax.php
```diff
@@ -131,10 +131,19 @@ function shop_url_check() {
             }
         }
 
-        wp_send_json_success( array(
-            'message' => __( 'Available', 'dokan-lite' ),
-            'url'     => sanitize_user( $url_slug )
-        ) );
+        if( is_admin() && isset( $_POST['vendor_id'] ) ) {
```
tareq1988: A space is needed before if

PR #211, opened 2018-03-29
added individual seller withdraw treshhold and fixed dokan_get_current_user_id function

saimonh3:


classes/template-withdraw.php
```diff
@@ -128,7 +128,12 @@ public function show_seller_balance() {
 
         $balance        = dokan_get_seller_balance( dokan_get_current_user_id(), true );
         $withdraw_limit = dokan_get_option( 'withdraw_limit', 'dokan_withdraw', -1 );
-        $threshold      = dokan_get_option( 'withdraw_date_limit', 'dokan_withdraw', -1 );
+
+        if ( get_user_meta( dokan_get_current_user_id(), 'withdraw_date_limit', true ) !== '' ) {
```
tareq1988: Move this to a function: `dokan_get_withdraw_threshold( $vendor_id )` and call it here. Same to other areas.

alamgircsebd:
Here have some conflicting errors but this one needed, we will do this changes later make own.

Issue #242, opened 2018-04-26
dokan-admin.js is not loaded on settings page so Select "Store Header Template" not working


iamdharmesh:
After update to version 2.8.0 dokan-admin.js is not loaded on settings page so Select "Store Header Template" not working

Please put 
`// Load common styles and scripts`
`wp_enqueue_style( 'dokan-admin-css' );`
`wp_enqueue_script( 'dokan-tooltip' );`
`wp_enqueue_script( 'dokan-admin' );`

Before or outside of
`// load vue app inside the parent menu only`
`if ( 'toplevel_page_dokan' == $hook ) {`

This will fix the issue.

![pasted image at 2018_04_26 12_28 pm](https://user-images.githubusercontent.com/10613171/39290517-5cfd8eba-491f-11e8-8392-e2863b40e888.png)

Thanks.

PR #244, opened 2018-04-27
Fixed 'dokan-admin.js is not loaded on settings page so Select "Store Header Template" not working' #242

iamdharmesh:
Fixed 'dokan-admin.js is not loaded on settings page so Select "Store Header Template" not working' #242


sabbir1991:
Thanks for your pull request. It has been fixed by our end already, so we are unable to accept your pull request right now. And also with your given solution, those JS files will be loaded in the whole WordPress backend thus it might conflict with other plugins. Then again thanks for your interest 👍

PR #348, opened 2018-08-16
refactor: use vendor_id from dokan_order table instead of post_author column of a shop order

saimonh3:


includes/order-functions.php
```diff
@@ -913,3 +906,29 @@ function dokan_order_csv_export( $orders, $file = null ) {
...
+ * @param  int order_id
+ *
+ * @return int
+ */
+function dokan_get_seller_id_by_order_id( $id ) {
+    $order = wc_get_order( $id );
+    $items = $order->get_items( 'line_item' );
+
+    // if we have multiple sellers then return 0
+    if ( count( $items ) > 1 ) {
```
sabbir1991: If a vendor has multiple line items or products then its always return 0

sabbir1991: If a vendor has one items then you just get author ID from first product id. If has multiple Items then you can only pick first one then get author id from this product id

includes/admin-functions.php
```diff
@@ -73,7 +73,7 @@ function dokan_admin_shop_order_edit_columns( $existing_columns ) {
 function dokan_shop_order_custom_columns( $col ) {
     global $post, $the_order;
 
-    if ( empty( $the_order ) || dokan_get_prop( $the_order, 'id') != $post->ID ) {
```
sabbir1991: Remove `dokan_get_props` since we don't provide support below WC 3.0 version. Just call directly `$the_order->get_id()`

includes/order-functions.php
```diff
@@ -383,12 +381,7 @@ function dokan_sync_insert_order( $order_id ) {
 function dokan_get_seller_id_by_order( $order_id ) {
     global $wpdb;
```
sabbir1991: Need persistence caching system. Use `wp_cache_set` and `wp_cache_get`

PR #351, opened 2018-08-19
refactor: code formatting

itowhid06:


includes/product-functions.php
```diff
@@ -441,14 +441,14 @@ function dokan_products_array_filter_editable( $product ) {
  */
 function dokan_product_get_row_action( $post ) {
 
-    if ( empty( $post->ID ) ) {
-        return $row_action;
```
ediamin: This is more preferable condition than your proposed changes, since there is no point of continuing ahead if we do not have `$post->ID`. Meaning, settings `$row_action`, `$row_action_html` and `$product_id` is pointless if `$post->ID` is empty.

templates/products/products-listing.php
```diff
@@ -290,7 +290,6 @@
 
                         </table>
                     </form>
-                </div>
```
ediamin: Why would you remove this closing `div` is unclear. Since, it closes the opening `div.dokan-dashboard-product-listing-wrapper`

includes/product-functions.php
ediamin: I think it is better to write `return array();` at this line instead of moving the code after assigning three variables.

PR #358, opened 2018-08-29
translation string in vendor dashboard and changing more product into…

shabnam611:
… more products

### DIFF
--- a/includes/template-tags.php
+++ b/includes/template-tags.php
@@ -330,7 +330,7 @@ function dokan_get_dashboard_nav() {
     );
 
     $settings = array(
-        'title' => sprintf(("%s <i class='fa fa-angle-right pull-right'></i>"), __( 'Settings', 'dokan-lite')),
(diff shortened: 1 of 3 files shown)


includes/template-tags.php
```diff
@@ -330,7 +330,7 @@ function dokan_get_dashboard_nav() {
     );
 
     $settings = array(
-        'title' => __( 'Settings <i class="fa fa-angle-right pull-right"></i>', 'dokan-lite'),
+        'title' => sprintf(__("%s <i class='fa fa-angle-right pull-right'></i>"), __( 'Settings', 'dokan-lite')),
```
sabbir1991: Don't need to add `__()` function inside `sprintf` function.

shabnam611: Thanks for the feedback, I corrected it. Please check.

ediamin: Use single quote outside and double quote for attributes. like 
`'%s <i class="fa fa-angle-right pull-right"></i>'`

sabbir1991: Please also remove extra `()` inside `sprintf`

PR #374, opened 2018-09-14
fix: commission issue when there is multiple category for a product

saimonh3:


saimonh3:
This commit is not needed.

PR #378, opened 2018-09-18
Refactor Dokan Setup Wizard

ediamin:
Update UI and add new Insight class.


classes/tracker.php
```diff
@@ -4,69 +4,37 @@
...
-    protected function get_order_count() {
-        global $wpdb;
+    public function appsero_init_tracker_dokan() {
+        if ( ! class_exists( 'AppSero\Insights' ) ) {
+            require_once DOKAN_LIB_DIR . '/appsero/src/insights.php';
+        }
 
-        return (int) $wpdb->get_var( "SELECT count(ID) FROM $wpdb->posts WHERE post_type = 'shop_order' and post_status IN ('wc-completed', 'wc-processing', 'wc-on-hold', 'wc-refunded');");
-    }
+        $this->insights = new AppSero\Insights( '559bcc0d-21b4-4b34-8317-3e072badf46d', 'Dokan Multivendor Marketplace', __FILE__ );
```
tareq1988: The `__FILE__` has to be the main file of Dokan. Which is `DOKAN_FILE`

tareq1988: Also, please send those extra data as before as well.

```$data = array(
    'products' => $this->get_post_count( 'product' ),
    'orders'   => $this->get_order_count(),
    'is_pro'   => $this->is_pro_exists() ? 'yes' : 'no'
);```

File: classes/tracker.php [hunk 2]
```diff
@@ -33,8 +33,25 @@ public function appsero_init_tracker_dokan() {
             require_once DOKAN_LIB_DIR . '/appsero/src/insights.php';
         }
 
-        $this->insights = new AppSero\Insights( '559bcc0d-21b4-4b34-8317-3e072badf46d', 'Dokan Multivendor Marketplace', __FILE__ );
+        $this->insights = new AppSero\Insights( '559bcc0d-21b4-4b34-8317-3e072badf46d', 'Dokan Multivendor Marketplace', DOKAN_FILE );
```
ediamin: @tareq1988 please review the changes.

lib/appsero/src/insights.php
```diff
@@ -483,7 +483,7 @@ public function optout() {
      *
      * @return integer
      */
-    protected function get_post_count( $post_type ) {
+    public function get_post_count( $post_type ) {
```
ediamin: @tareq1988 I had to make it public. I think it will be better to use this class as abstract and try to send tracker data for multiple plugin in a single post request.

tareq1988: Yeah the function can be public, but making the class abstract forces everyone to extend the class.

PR #381, opened 2018-09-19
use wp_add_inline_script()

QWp6t:
[wp_add_inline_script()](https://developer.wordpress.org/reference/functions/wp_add_inline_script/) is made specifically for this purpose. It is preferred here because otherwise it's possible that your script will be appended prior to jQuery being loaded.

This is untested. I made this edit in the GitHub editor. I'll give you guys write permission to my PR if you want to commit any changes yourselves.


alamgircsebd:
This one no longer needed, we have refactor the structure after v3.0.0

PR #425, opened 2018-11-10
calculate commission for item by quantity

hamedmoody:
Now by selling each item by greater than 1 quantity, admin commission multiply with item quantity


UVLabs:
this should be merged

UVLabs:
@hamedmoody this fixes the issue where the admin only gets the commission once if the product has more than one of itself in the cart right?

Like if someone purchases 4 dog toys costing $2 each(from the same vendor) and the admin has $1 commission, **currently the admin would only get $1 commission and vendor would get $7**, _instead of_ the admin getting $4 commission and vendor getting $4

saimonh3:
Thanks for the PR. This feature has already been implemented in dokan.

PR #438, opened 2018-11-19
Added new action and filter hooks

softound:
Hello,

We have added new filter and action hooks those are very useful in our project,
Please check and implement them.

Thanks

### DIFF
--- a/includes/class-order-manager.php
+++ b/includes/class-order-manager.php
@@ -299,6 +299,9 @@ class Dokan_Order_Manager {
         }
 
         $order->save();
+        
+        //New action hook to add custom line items or trigger function after line items are added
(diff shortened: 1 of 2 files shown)


includes/class-order-manager.php
```diff
@@ -299,6 +299,9 @@ public function create_line_items( $order, $products ) {
         }
 
         $order->save();
+        
+        //New action hook to add custom line items or trigger function after line items are added
+        do_action('after_create_line_items', $order);
```
sabbir1991: Please add `dokan` prefix before hook name and add proper spacing
```suggestion
        do_action( 'dokan_after_create_line_items', $order );
```

softound:
thank you :)

PR #441, opened 2018-11-27
feat: add dokan_dashboard_wrap_start and end hook for theme compatibility

saimonh3:


templates/products/products-listing.php
```diff
@@ -224,3 +226,5 @@
         ?>
 
     </div><!-- .dokan-dashboard-wrap -->
+
+    <?php do_action( 'dokan_dashboard_wrap_end' ); ?>
```
sabbir1991: Please fix line intending @saimonh3

PR #448, opened 2018-12-03
refactor: remove offline payment method options

saimonh3:


UVLabs:
@saimonh3 why are you removing this?? This was a great feature addition to the plugin why do you want to remove it??

saimonh3:
@UVLabs There are some things which is yet to be implemented regarding offline payment. So we are taking off this feature for now. We will be adding this with some other features in near future. Thanks

UVLabs:
@saimonh3 do you have any roadmap or idea when? This is kind of a needed feature maybe you can add a filter/action or something to let us control this for now if we want to until you implement the full feature?

PR #522, opened 2019-01-10
Update order-functions.php

softound:
Add "dokan_order_total" filter hook to modify the order total. Thanks


tareq1988:
I know it's an old pull request and wasn't reviewed in time. Now it has conflicts, so it won't get merged. Also, I don't think it's a good idea to add filter on a certain item. Instead, the whole array should have a filter before inserting the data into the DB, if needed.

PR #535, opened 2019-01-22
refactor(Dashboard): add action to dashboard nav

aft3rlife:
add action to dashboard nav to control nav links


sabbir1991:
I think this action is not necessary for changing nav items. You can use filters for changing any urls or nav item values. for general nav menus you can use `dokan_get_dashboard_nav` and for common links just use `dokan_dashboard_nav_common_link`

PR #574, opened 2019-02-26
refactor: set default threshold day to avoid invalid date creation

saimonh3:


includes/order-functions.php
```diff
@@ -312,7 +312,7 @@ function dokan_sync_insert_order( $order_id ) {
     $admin_commission   = dokan_get_admin_commission_by( $order, $seller_id );
     $net_amount         = $order_total - $admin_commission;
     $net_amount         = apply_filters( 'dokan_order_net_amount', $net_amount, $order );
-    $threshold_day      = dokan_get_option( 'withdraw_date_limit', 'dokan_withdraw', 0 );
+    $threshold_day      = dokan_get_option( 'withdraw_date_limit', 'dokan_withdraw' ) ? dokan_get_option( 'withdraw_date_limit', 'dokan_withdraw' ) : 0;
```
sabbir1991: No need to use `dokan_get_option` twice. Use a variable for holding this value before.

PR #641, opened 2019-06-20
feat: add stock notification email for vendor

saimonh3:


alamgircsebd:
Here have conflicting errors, right now we can't marge this PR because have refactor full plugin structure after v3.0.0
This features needed, we will do this later.

PR #643, opened 2019-06-22
Split orders in admin dashboard.

mi5t4n:
Split the order when the order containing products from multiple vendors is created from the admin dashboard.

### DIFF
--- a/includes/class-order-manager.php
+++ b/includes/class-order-manager.php
@@ -25,6 +25,10 @@ class Dokan_Order_Manager {
 
         // restore order stock if it's been reduced by twice
         add_action( 'woocommerce_reduce_order_stock', array( $this, 'restore_reduced_order_stock' ) );
+
+        // Split orders when order containing producuts from multiple vendor is created from admin dashboard.
+        add_action( 'save_post_shop_order', array( $this, 'split_order_admin_dashbaord' ), 10, 3 );
+
     }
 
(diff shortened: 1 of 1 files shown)


mi5t4n:
@tareq1988 can you check the pull request?

saimonh3:
@mi5t4n Thanks for the PR. It does split the order but order status on the sub orders doesn't seems working accordingly. Would you check?

mi5t4n:
ok i will check it.

mi5t4n:
@saimonh3 I have checked it. The status of the suborders can also be changed from the admin dashboard as well as the seller dashboard. The order status of the suborders also changes to the parent order status. Can you verify it on your end, it is working on my end.

saimonh3:
@mi5t4n When the parent order status is anything else other than 'pending payment' it doesn't work. For example, if the parent order status is 'processing' while creating the order, all the sub orders status should be 'processing' as well.

mi5t4n:
@saimonh3 Thank you for your response. As you said, indeed the status of the suborder is not set to the status of the parent order when the order is created with a status other than processing. I have changed the code and committed. Please verify on your end.

saimonh3:
@mi5t4n It works now. Thanks for PR.

mi5t4n:
@saimonh3 Thank You too.

PR #677, opened 2019-10-03
Update product-functions.php

kas5986:
have made some customization for dokan and there was no hook to get into this function so added a new filter to modify it on runtime hope this will be added in new version


alamgircsebd:
This one already implemented on plugin

PR #695, opened 2019-11-12
Fix vendor payment settings page available withdraw methods display i…

siananda:
How to reproduce this issue:

1. Uncheck **Withdraw Methods** from admin Withdraw Options settings page.

admin withdraw method settings

2. Login using vendor account information and check vendor **Payment Settings** page.

Payment Settings page


sabbir1991:
Thanks for your contribution 👍

PR #706, opened 2019-11-29
filter is required here to do custom modification

kas5986:
filter or action is required to do changes by hooks if someone is creating new plugin or customizing core functionalities !

### DIFF
--- a/templates/orders/details.php
+++ b/templates/orders/details.php
@@ -238,7 +238,7 @@ $hide_customer_info = dokan_get_option( 'hide_customer_info', 'dokan_selling', '
                                 <span><?php esc_html_e( 'Customer IP:', 'dokan-lite' ); ?></span>
                                 <?php echo esc_html( get_post_meta( dokan_get_prop( $order, 'id' ), '_customer_ip_address', true ) ); ?>
                             </li>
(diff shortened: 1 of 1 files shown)


templates/orders/details.php
```diff
@@ -238,6 +238,7 @@
                                 <span><?php esc_html_e( 'Customer IP:', 'dokan-lite' ); ?></span>
                                 <?php echo esc_html( get_post_meta( dokan_get_prop( $order, 'id' ), '_customer_ip_address', true ) ); ?>
                             </li>
+                            <?php apply_filters( 'dokan_manage_order_details_after', $order, $order_id, $order_items);?>
```
sabbir1991: Please change this filter name to `dokan_order_details_after_customer_info` instead of `dokan_manage_order_details_after`

kas5986: resolved

saimonh3:
@kas5986 Thanks for the PR, Does do_action serve your needs? If so please make it do_action instead of apply_filters.

kas5986:
> @kas5986 Thanks for the PR, Does do_action serve your needs? If so please make it do_action instead of apply_filters.

sure it works with do action but my need was only filter however made it aciton and changing my code hook into action as well :)

PR #709, opened 2019-12-09
Feature Request: new filter dokan_withdraw_requests_limit to override the withdraw request list page limit value

siananda:
Feature Request: please add a new filter dokan_withdraw_requests_limit to override the withdraw request list page limit value.

### DIFF
--- a/classes/withdraw.php
+++ b/classes/withdraw.php
@@ -126,6 +126,8 @@ class Dokan_Withdraw {
     function get_withdraw_requests( $user_id = '', $status = 0, $limit = 10, $offset = 0 ) {
         global $wpdb;
 
+        $limit = apply_filters( 'dokan_withdraw_requests_limit', $limit, $status );
+
         if ( empty( $user_id ) ) {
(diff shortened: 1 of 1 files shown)


iammahmudur:
@Sk-Shaikat vai

Could you please look into this soon?

saimonh3:
@siananda Thanks for the pull request. Unfortunately we can't merge this right now, as it's not a good idea to filter function's parameter for our use cases. 

`get_withdraw_requests( $user_id = '', $status = 0, $limit = 10, $offset = 0 )`
This should be done in the REST API end. Currently a plugin wide refactor is in progress and this issue will be addressed there.

saimonh3:
@siananda Thanks for the pull request. Unfortunately we can't merge this right now, as it's not a good idea to filter function's parameter for our use cases.

```php
get_withdraw_requests( $user_id = '', $status = 0, $limit = 10, $offset = 0 )
```
This should be done in the REST API end. Currently a plugin wide refactor is in progress and this issue will be addressed there.

Thanks again for your concern, it's highly appreciated.

siananda:
@saimonh3 thanks for your comment. Can you suggest me, how can I change admin **Withdraw Requests** page perPage value? It's default 10. I want to 50 per page. 

withdraw requests

PR #713, opened 2019-12-10
added new action hook to to add new element

kas5986:
adding new element for my custom project after products this was required and added hope it will be added in coming version


alamgircsebd:
This one no longer needed

PR #716, opened 2019-12-12
Filter hook added for store and registration validation

shuvorroy:
As per our client's requirements, we needed to validate some fields from vendor registration and store settings form. For this purpose, we have added 2 filter hooks


alamgircsebd:
Here have conflicting errors, right now we can't marge this PR because we have refactor after v3.0.0 where changed the structure where not capable this request

PR #718, opened 2019-12-19
Limit Query to "shop_order" Post Type Only

cyberwani:
Added fix to limit Order IDs from "shop_order" post type only.


alamgircsebd:
Here have conflicting errors, right now we can't marge this PR because we have refactor after v3.0.0 where changed the structure where not capable this request.

This changes needed, we will add this.

PR #765, opened 2020-02-10
Fix: When no product show empty table issue fixed #758

### LINKED ISSUE #758: When there are no products, instead of showing empty table maybe we can mimic how WooCommerce does this
Screen Shot 2020-02-05 at 12 53 40 PM

Screen Shot 2020-02-05 at 12 55 43 PM

alamgircsebd:


sabbir1991:
Need to improve the code logic and a better design

PR #780, opened 2020-02-19
feat: add product and order based commission calculation

saimonh3:
Screen Shot 2020-02-19 at 4 19 56 PM


alamgircsebd:
Here have conflicting errors, right now we can't marge this PR because we have refactor full plugin structure after v3.0.0, where not capable this request.

This features not requested any more. we will add this features later.

PR #800, opened 2020-03-26
fix - Vendor is not receiving order email issue fixed #799

### LINKED ISSUE #799: Vendor is not receiving order email
Description: 

If a product is purchased the vendor is not receiving email 

How to reproduce :

1. Purchase product from a vendor 
2. The email is sent to admin and customer 
3. The vendor is not receiving the order email 

Expected behaviours : 

Vendors should receive email

alamgircsebd:


tareq1988:
@sabbir1991 is it a good idea to check by `get_post_meta` directly?

alamgircsebd:
> 
> 
> @sabbir1991 is it a good idea to check by `get_post_meta` directly?

No, this is not a good idea. Here we need to create proper function which one can check sub order, here we used dokan_is_sub_order one but it not working properly. Let me try to do this properly.

marine7494:
Hi,
I have a marketplace with Dokan.
Right now, I centralize all the order in my mailbox and forward it to the right vendor, because the plugin didn't work very well in the vendors notifications.
Do you confirm this change is working well to fix this problem ?
I am beginner in coding, could you detail the exact procedure, please ?
very sorry to ask that ...
Many thanks
Marine

alamgircsebd:
> Hi,
> I have a marketplace with Dokan.
> Right now, I centralize all the order in my mailbox and forward it to the right vendor, because the plugin didn't work very well in the vendors notifications.
> Do you confirm this change is working well to fix this problem ?
> I am beginner in coding, could you detail the exact procedure, please ?
> very sorry to ask that ...
> Many thanks
> Marine

Hello Marine

Thanks for your comments. We solved this issue and released V3.0.3 where you will get this updates. Please update your plugin hope you will get changes. 

https://wordpress.org/plugins/dokan-lite/ 

Let us know if you get any more issue with vendor order email.

Thanks

PR #821, opened 2020-04-23
Fix order manager caching

kvasbo:
The current order manager cache implementation is not working, as it does not take many of the arguments into account. I have replaced it with a simpler and non-manual cache key generator that will work for any set of parameters.

### DIFF
--- a/includes/Order/Manager.php
+++ b/includes/Order/Manager.php
@@ -34,7 +34,7 @@ class Manager {
         $cache_group = 'dokan_seller_data_'.$args['seller_id'];
         
         // Use all arguments to create a hash used as cache key
-        $cache_key = md5(json_encode($args));
+        $cache_key = 'dokan_seller_orders_'.md5(json_encode($args));
         
         $orders      = wp_cache_get( $cache_key, $cache_group );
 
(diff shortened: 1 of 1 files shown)


includes/Order/Manager.php
```diff
@@ -30,9 +30,12 @@ public function all( $args = [] ) {
 
         $args = wp_parse_args( $args, $default );
 
-        $offset       = ( $args['paged'] - 1 ) * $args['limit'];
+        $offset      = ( $args['paged'] - 1 ) * $args['limit'];
         $cache_group = 'dokan_seller_data_'.$args['seller_id'];
-        $cache_key   = 'dokan-seller-orders-' . $args['status'] . '-' . $args['seller_id'];
+        
+        // Use all arguments to create a hash used as cache key
+        $cache_key   = 'dokan_seller_orders-' . md5(json_encode($args));
```
saimonh3, line 37: @kvasbo Thanks for the PR. Seems good to me, just a minor change is needed. Can you please insert spaces as needed? Something like this.
```php
        $cache_key = 'dokan_seller_orders-' . md5( json_encode( $args ) );
```

kvasbo: Too late now I guess :)

PR #827, opened 2020-04-30
Updated order functions.php in includes

Thisura98:
fix(Order): filter out empty seller ids

Fixes a bug where empty Seller IDs were returned by dokan_get_sellers_by.

### DIFF
--- a/includes/Order/functions.php
+++ b/includes/Order/functions.php
@@ -608,8 +608,8 @@ function dokan_get_sellers_by( $order ) {
 
         //New filter hook to modify the seller id at run time.
         $seller_id = apply_filters( 'dokan_get_sellers_by', $seller_id, $item );
-        if(!empty($seller_id)){
-            $sellers[$seller_id][] = $item;
+        if ( ! empty( $seller_id ) ) {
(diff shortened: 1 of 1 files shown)


includes/Order/functions.php
```diff
@@ -608,7 +608,9 @@ function dokan_get_sellers_by( $order ) {
 
         //New filter hook to modify the seller id at run time.
         $seller_id = apply_filters( 'dokan_get_sellers_by', $seller_id, $item );
-        $sellers[$seller_id][] = $item;
+        if(!empty($seller_id)){
```
saimonh3, line 611: Thanks for the PR. Can you please insert space inside the if condition? Something like this.

```php
        if ( ! empty( $seller_id ) ) {
            $sellers[ $seller_id ][] = $item;
        }
```

Thisura98: Hello! I added some spacing inside the if condition and made a new commit [#cc2e136 ](https://github.com/weDevsOfficial/dokan/pull/827/commits/cc2e13671cb1f81b0489cb36c15aea9cf447ce88) as requested.

PR #834, opened 2020-05-15
fix: On RESTful order creation, only single store is added into the response even if there are multiple stores is fixed #833

### LINKED ISSUE #833: On RESTful order creation, only single store is added into the response even if there are multiple stores

saimonh3:


saimonh3:
**Response when there are multiple stores.**

Screen Shot 2020-05-15 at 11 20 12 AM

**When there is single store.**

Screen Shot 2020-05-15 at 11 20 12 AM

saimonh3:
@Mossalli1 Please have a look.

PR #835, opened 2020-05-17
Logo problem on mobile device

kas5986:
Logo width is the disturbing overall layout and making mobile device user confused 100% width fix this issue

### DIFF
--- a/includes/Vendor/SetupWizard.php
+++ b/includes/Vendor/SetupWizard.php
@@ -123,7 +123,7 @@ class SetupWizard extends DokanSetupWizard {
         </head>
         <body class="wc-setup wp-core-ui">
             <?php if ( ! empty( $this->custom_logo ) ) { ?>
-                <h1 id="wc-logo"><a href="<?php echo esc_url( home_url() ) ?>"><img src="<?php echo esc_url( $this->custom_logo ); ?>" alt="<?php echo esc_attr( get_bloginfo( 'name' ) ); ?>" style="width : 100%;" /></a></h1>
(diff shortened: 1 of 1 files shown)


includes/Vendor/SetupWizard.php
```diff
@@ -123,7 +123,7 @@ public function setup_wizard_header() {
         </head>
         <body class="wc-setup wp-core-ui">
             <?php if ( ! empty( $this->custom_logo ) ) { ?>
-                <h1 id="wc-logo"><a href="<?php echo esc_url( home_url() ) ?>"><img src="<?php echo esc_url( $this->custom_logo ); ?>" alt="<?php echo esc_attr( get_bloginfo( 'name' ) ); ?>" /></a></h1>
+                <h1 id="wc-logo"><a href="<?php echo esc_url( home_url() ) ?>"><img src="<?php echo esc_url( $this->custom_logo ); ?>" alt="<?php echo esc_attr( get_bloginfo( 'name' ) ); ?>" style="width : 100%;" /></a></h1>
```
alamgircsebd, line 126: This is wrong thinking, if you use 100% width then it will be problem when use big image, also we don't allow inline CSS, try to use on source files for css/js

kas5986: do you want me to remove this line and add this to css?

kas5986:
https://paste.pics/fdf5194b10fa6189a98e84638464c56f without 100%
https://paste.pics/031d955b7c797f5f56ade959d62b5ca1

https://paste.pics/5857b227c9617a947955ccd07b37ead6 with 100%

includes/Vendor/SetupWizard.php
alamgircsebd: Right, please remove the inline css and use the css codes on this file `assets\css\setup.css` with `CSS @media Rules` because you want to use for only mobile devices

alamgircsebd:
Already have another request here https://github.com/weDevsOfficial/dokan/pull/1007

PR #844, opened 2020-06-01
Store settings for multiple shifting. functionality seems fine. desig…

zamandevs:
…n break when add more shift

### DIFF
--- a/includes/Dashboard/Templates/Settings.php
+++ b/includes/Dashboard/Templates/Settings.php
@@ -17,9 +17,9 @@ class Settings {
     /**
      * Loading autometically when class initiate
      *
+     * @return void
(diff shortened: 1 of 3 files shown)


includes/Dashboard/Templates/Settings.php
```diff
@@ -17,9 +17,9 @@ class Settings {
     /**
      * Loading autometically when class initiate
      *
+     * @return void
```
ediamin, line 20: @zamandevs why did you move this line for every method?

zamandevs: didn't notice that. I maintain phpstorm  plugin for Wordpress. when reformatted whole code it occurred.

PR #861, opened 2020-07-02
fix - remove admin role vendors from store listing page

alamgircsebd:


sabbir1991:
Invalid. Not recommended this way

PR #872, opened 2020-07-15
Fix: admin vendor total count fixed #850

### LINKED ISSUE #850: Wrong filter count in the backend Vendor listing page.
**Description**: The count filter counter only counts the seller role and the admin role is being ignored.

**How to reproduce:**
1. Go to Dokan> Vendors tab.
2. Now in the screenshot notice that All(8) filter says I have 8 stores but in reality I have 9 stores.
image

3. It is the same for the Active and Pending tab as well. The admin store is being ignored.

zamandevs:


ediamin:
@zamandevs PR title should be like this `Fix admin vendor total count`. Then put `Fixes #850` either in commit description or PR description. Commit description is better of course.

PR #877, opened 2020-07-23
fix namespace WC_Order

aqjw:


ediamin:
Good catch! Thanks for the contribution 👍

PR #882, opened 2020-08-05
fix: Bug fix for "Withdrawal current balance is incorrect"

ceefour:
Reference: Dokan Pro support ticket #338760.

This PR fixes the bug that occurs because Dokan did not use proper cache behavior :

1. used wrong cache group & cache key
2. did not flush the relevant cache entries after balance was updated

### DIFF
--- a/includes/Withdraw/Hooks.php
+++ b/includes/Withdraw/Hooks.php
@@ -28,7 +28,9 @@ class Hooks {
      * @return void
      */
     public static function delete_seller_balance_cache( $status, $user_id, $id ) {
-        wp_cache_delete( 'dokan_seller_balance_' . $user_id );
+        $cache_group = 'dokan_seller_data_' . $user_id;
+        $cache_key = 'dokan_seller_balance_' . $user_id;
+        wp_cache_delete( $cache_key, $cache_group );
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ceefour:
Please merge

yasqihafizhana:
Please merge, because we need to fix this bug immediately

kapilpaul (commented):
this is working fine.

zamandevs:
@ceefour Did you check Reports? It's not working properly. When caching enables our current implementation for **Withdrew current balance** is wrong & reports is fine. but this PR cause wrong **Reports**. Please look into this scenario.

zamandevs (commented):
Its not working properly. produce wrong **Report**.

ceefour:
@zamandevs Thank you for confirming that "When caching enabled our current implementation for **Withdraw current balance is wrong** & reports is fine"

Note that I am not a weDevs developer, I am developer with an active Dokan Professional 3 Sites license. However, previously weDevs have refused to acknowledge this problem since more than a month ago, and blamed the caching plugin. That's why I attempted, with very limited knowledge of Dokan plugin internals, to make my own fix and contribute the solution back.

Can you confirm that weDevs will provide a proper fix to this problem and can you give a time estimate ?

zamandevs:
@ceefour Thanks for your contribution. hopefully, we will fix this on the next release.

ceefour:
@tareq1988 @kapilpaul I've removed the requested lines. Please re-review for merge.

PR #893, opened 2020-08-19
fix: Ensure sub-orders also get the correct payment gateway fee (if any)

ceefour:
Enable `dokan_get_processing_fee` filter hook callback to calculate payment gateway fee for sub-orders as well.

Required to fix Dokan Pro support ticket #124441.


yasqihafizhana:
please merge

PR #903, opened 2020-08-31
fix: guest checkout name in vendor order details fixed #894

### LINKED ISSUE #894: Guest Checkout name on the vendor order
Description : 

If you enable guest check checkout Dokan vendor order shows guest while the WooCommerce shows the user name

Expected behavior

Dokan should show the user name the same as WooComerce does 

Screenshots : 

https://nimb.ws/hEWenJ
https://nimb.ws/szZl0Q

kapilpaul:


templates/orders/listing.php
```diff
@@ -103,7 +103,7 @@
                                 }
 
                             } else {
-                                $user = __( 'Guest', 'dokan-lite' );
+                                $user = __( $order->get_formatted_billing_full_name(), 'dokan-lite' );
```
sabbir1991, line 106: Remove translation function from vaiable

PR #916, opened 2020-09-09
fix: Order fetching issues with REST API fixed #818

### LINKED ISSUE #818: Order fetching issues with REST API
**These doesn't work as expected:**

```
wp-json/dokan/v1/orders/?status=processing
wp-json/dokan/v1/orders/?order=asc&orderby=id
wp-json/dokan/v1/orders/?exclude=3589
```

zamandevs:


kapilpaul (commented):
this commits looks fine

PR #919, opened 2020-09-09
Fix/order status refunded issue

alamgircsebd:


tareq1988:
@alamgircsebd try running phpcs locally: `./vendor/bin/phpcs includes/Order/Hooks.php`. If phpcs isn't installed, run `composer install`.

PR #926, opened 2020-09-21
fix protect to Csrf bug with wp_verify_nonce

lenonleite:
change condition AND to OR and isset operator

### DIFF
--- a/includes/Admin/UserProfile.php
+++ b/includes/Admin/UserProfile.php
@@ -538,7 +538,7 @@ class UserProfile {
 
         $post_data = wp_unslash( $_POST );
 
-        if ( isset( $post_data['dokan_update_user_profile_info_nonce'] ) && ! wp_verify_nonce( $post_data['dokan_update_user_profile_info_nonce'], 'dokan_update_user_profile_info' ) ) {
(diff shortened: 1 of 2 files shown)


lenonleite:
haha No problem, I can do it, one minute. =)

lenonleite:
PHPCS is set many points but I changed just one line. What's happen? =/

includes/Admin/UserProfile.php
```diff
@@ -538,7 +538,7 @@ function save_meta_fields( $user_id ) {
 
         $post_data = wp_unslash( $_POST );
 
-        if ( isset( $post_data['dokan_update_user_profile_info_nonce'] ) && ! wp_verify_nonce( $post_data['dokan_update_user_profile_info_nonce'], 'dokan_update_user_profile_info' ) ) {
+        if ( ! isset( $post_data['dokan_update_user_profile_info_nonce'] ) || ( isset( $post_data['dokan_update_user_profile_info_nonce'] ) && ! wp_verify_nonce( $post_data['dokan_update_user_profile_info_nonce'], 'dokan_update_user_profile_info' ) ) ) {
```
kapilpaul, line 541: No need to check the isset again in the second condition, because you are already cheking it in the first condition.

you can write it as
`        if ( ! isset( $post_data['dokan_update_user_profile_info_nonce'] ) || ! wp_verify_nonce( $post_data['dokan_update_user_profile_info_nonce'], 'dokan_update_user_profile_info' ) ) {
`

nurul-umbhiya:
Already fixed in dokan lite.

PR #927, opened 2020-09-24
fix: redirect to 404 if vendor do not exist for toc template

### LINKED ISSUE #922: Vendor's Terms and Conditions page is not being deleted after deleting a vendor
**Description:** 
If a vendor is permanently deleted, the vendor's terms and conditions page is not being deleted. In fact, it does not display any terms and conditions content of the vendor. However, the URL still works.

**How to reproduce the issue:**
- Create a new vendor
- Add vendor terms and conditions content.
- Permanently delete the vendor from the Users panel
- Then check the Seller Terms and Conditions page.
- URL works. However, the terms and conditions of that store do not contain any content. (https://prnt.sc/uhaslp)

**Expected behavior:**
The page should not be found.

kapilpaul:
fixed #922

### DIFF
--- a/includes/Rewrites.php
+++ b/includes/Rewrites.php
@@ -9,7 +9,7 @@ namespace WeDevs\Dokan;
  */
 class Rewrites {
 
-    public $query_vars = array();
+    public $query_vars = [];
     public $custom_store_url = '';
 
     /**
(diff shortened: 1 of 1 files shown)


includes/Rewrites.php
```diff
@@ -211,12 +211,24 @@ function store_template( $template ) {
...
+    public function store_toc_template( $template ) {
+        $store_name = get_query_var( $this->custom_store_url );
+
         if ( ! $this->is_woo_installed() ) {
             return $template;
         }
-        if ( get_query_var( 'toc' ) ) {
-            return dokan_locate_template( 'store-toc.php' );
+
+        if ( ! empty( $store_name ) ) {
```
tareq1988, line 221: We should check the existence of query var `toc`, then we should fetch the seller. That's a performance improvement I guess. If the query var wasn't there, still the user will be fetched. So we can defer the fetching later once we know the query var we are expecting was the right one.

kapilpaul: Yes, that's a better idea. I will update the PR.

PR #931, opened 2020-10-12
Feat: dashboard banner with upgrade popup, vendor capabilities page

xaviranik:


tareq1988 (changes requested):
All texts should be translateable. Lot texts doesn't have textdomains. Add textdomains on all the strings.

kapilpaul:
also please fix the phpcs errors.

xaviranik:
> All texts should be translateable. Lot texts doesn't have textdomains. Add textdomains on all the strings.

working on it bhai

src/admin/components/UpgradePopup.vue
```diff
@@ -0,0 +1,195 @@
+<template>
+    <div id="dokan-upgrade-wrapper">
+        <div id="dokan-upgrade-popper">
+            <p>{{ __("You're using", 'dokan-lite' ) }} <span>{{ __( 'Dokan Lite', 'dokan-lite' ) }}</span>. {{ __( 'To unlock more features, consider', 'dokan-lite' ) }}</p>
+            <a href="https://wedevs.com/dokan-lite-upgrade-to-pro" class="button">{{ __( 'Upgrade to Pro', 'dokan-lite' ) }}</a>
+        </div>
+
+        <div v-if="showPopup" id="dokan-upgrade-popup">
```
kapilpaul, line 8: Do we need this modal popup? if yes then how it will trigger? if no then we can remove the codes.

src/admin/components/UpgradePopup.vue
```diff
@@ -0,0 +1,195 @@
+<template>
+    <div id="dokan-upgrade-wrapper">
+        <div id="dokan-upgrade-popper">
+            <p>{{ __("You're using", 'dokan-lite' ) }} <span>{{ __( 'Dokan Lite', 'dokan-lite' ) }}</span>. {{ __( 'To unlock more features, consider', 'dokan-lite' ) }}</p>
+            <a href="https://wedevs.com/dokan-lite-upgrade-to-pro" class="button">{{ __( 'Upgrade to Pro', 'dokan-lite' ) }}</a>
```
kapilpaul, line 5: Can we get the URL dynamically from the backend? We can bind this with dokan variable.

src/admin/pages/Settings.vue
```diff
@@ -49,33 +50,38 @@
...
+                    </template>
+                </div>
+
+                <div class="loading" v-if="showLoading">
+                    <loading></loading>
+                </div>
             </div>
         </div>
+
+        <SettingsBanner if=" ! hasPro"></SettingsBanner>
```
kapilpaul, line 67: Please remove the white space before ! eg: `<SettingsBanner if="! hasPro"></SettingsBanner>`

src/admin/components/SettingsBanner.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+<div id="dokan-settings-banner">
+    <img :src="bannerImage" :alt="__( 'Dokan Settings Banner', 'dokan-lite' )">
+    <div class="content">
+        <h1>{{ __( 'Looking for Coupons, Variable Products, SEO or Shipping?', 'dokan-lite' ) }}</h1>
+        <p>{{ __( 'Unlock these awesome frontend features with Dokan PRO', 'dokan-lite' ) }}</p>
+        <a class="button" href="https://wedevs.com/dokan-lite-upgrade-to-pro">{{ __( 'Upgrade to Pro', 'dokan-lite' ) }}</a>
```
kapilpaul, line 7: Can we get the URL dynamically from the backend? We can bind this with dokan variable.

src/admin/pages/Premium.vue
```diff
@@ -13,6 +13,20 @@
         </div>
         <!-- End of Header Section -->
 
+        <!--   Vendor Capabilities Banner     -->
+        <div class="vendor-capabilities-banner" :style="bannerBg">
```
kapilpaul, line 17: This section is almost the same as VendorCapabilities.vue's banner section. Can we do one thing to put this in a component and add props for the banner image. Also, we can add a slot so that if we need we can add buttons too.

PR #933, opened 2020-10-14
order list - fix seller

aqjw:


aqjw:
@ceefour @tareq1988 @kvasbo @CasperBraske could you check this shit?

tareq1988:
@aqjw is this the correct language for opening a PR? Give me a reason why should I even bother to think about reviewing it?

aqjw:
> @aqjw is this the correct language for opening a PR? Give me a reason why should I even bother to think about reviewing it?

Sorry about that. I just wanted to get your attention. It took me 3 hours to find a bug on the site. But the bug ended up in your plugin

tareq1988:
There could be bug anywhere, but that's a very wrong a of getting the attention my friend. Please fix the language.

ceefour:
Factually @aqjw did get attention although in a terrible way, you only waited 3 minutes between posting and the (initial) foul language. Thanks for fixing the comment language. :)

For comparison, PR #882 needed almost one month to get a review (and we're on a Dokan Pro license), but thankfully it was marked "fixed on next release", although it's still open now..

aqjw:
Could you also check this?  https://github.com/weDevsOfficial/wp-user-frontend/pull/969

tareq1988:
@aqjw in this PR, the `$post->ID` and the new `$order_id` you passed are the same. I don't see any benefit from passing the other parameter. If the `$post->ID` wasn't correct, the rest of the function wouldn't have worked.

aqjw:
> @aqjw in this PR, the `$post->ID` and the new `$order_id` you passed are the same. I don't see any benefit from passing the other parameter. If the `$post->ID` wasn't correct, the rest of the function wouldn't have worked.

@tareq1988 they are the same in most cases, but if there are orders and suborders, then everything is confused.
Try to create suborders, please.

tareq1988:
In either case, the ID's are the same. I don't get it how that changes anything? Dokan by default hides the suborders only, which can be toggled. Still they are individual rows in the table and the ID's are unchanged. 

image

aqjw:
order list 
image

order page
image

product page
image


___________________________________________________________________

order list after the changes https://github.com/weDevsOfficial/dokan/pull/933/commits/53239be7b822843444cf286f6eb28883c4f1d9c4

image

tareq1988:
According to your screenshots, I can see the problem. But maybe the problem is happening because of plugin conflicts, some other plugin is modifying the global `$post` object, that's why maybe you're seeing a different result. But if that was the case, other columns added by Dokan would be different as well. 

So I'm closing this PR, we need a more reliable solution for getting the order ID. Maybe by passing the 2nd argument, in that case the code in other places needs to be modified as well.

PR #945, opened 2020-11-06
fix: on vacation product listing not showing on vendor dashboard issu…

alamgircsebd:
…e fixed

### DIFF
--- a/templates/products/products-listing.php
+++ b/templates/products/products-listing.php
@@ -102,13 +102,14 @@
                             <tbody>
                                 <?php
                                 $pagenum       = isset( $_GET['pagenum'] ) ? absint( $_GET['pagenum'] ) : 1;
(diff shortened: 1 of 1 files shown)


templates/products/products-listing.php
```diff
@@ -102,13 +102,14 @@
                             <tbody>
                                 <?php
                                 $pagenum       = isset( $_GET['pagenum'] ) ? absint( $_GET['pagenum'] ) : 1;
-                                $post_statuses = array( 'publish', 'draft', 'pending', 'future' );
+                                $post_statuses = array( 'publish', 'draft', 'pending', 'future', 'vacation' );
```
kapilpaul, line 105: Here you may also apply a filter to handle this from PRO version.

alamgircsebd:
Already implement on Dokan pro seller vacation module

PR #949, opened 2020-11-08
fix: vendor created by admin shows in search result

### LINKED ISSUE #936: Search result does not show store/vendor when added by Admin
**Description of the bug:**
If an admin adds a vendor from Dokan > Vendors > Add new the store is not found in the store-list search until the vendor saves his store settings from the dashboard.

**How to reproduce:**
1. Add a vendor from wp-admin > Dokan > Vendors section.
2. Search the store from the store listing page (it will show no vendor found)
3. login to the vendor dashboard and save store settings.
4. Search again on the store list page, you should get the store as expected

**Expected behavior:**
The store should be available on the search without the vendor needing to save the store settings.

First reported: https://wordpress.org/support/topic/store-listing-search/
Reproduced by me and the issue exists on latest Dokan Lite

WordPress: 5.5.1
Dokan-Lite: 3.0.9
Dokan-PRO: 3.0.8
WooCommerce: 4.5.2

kapilpaul:
fixed #936

### DIFF
--- a/includes/Vendor/Manager.php
+++ b/includes/Vendor/Manager.php
@@ -196,6 +196,7 @@ class Manager {
         }
 
         $vendor->update_meta( 'dokan_profile_settings', $store_data );
+        $vendor->update_meta( 'dokan_store_name', $store_data['store_name'] );
         $vendor->set_store_name( $store_data['store_name'] );
         $vendor->save();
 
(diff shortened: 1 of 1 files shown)


alamgircsebd:
Code seems good, issue tested and resolved with this fix

PR #950, opened 2020-11-09
update: Apply new filter on product listing status

alamgircsebd:


kapilpaul (commented):
This commit seems ok. We should use the shorthand array declaration. This is also a rule in our phpcs.

PR #951, opened 2020-11-09
update: Add new filter on dokan_is_product_author

alamgircsebd:


tareq1988:
@alamgircsebd Why is this filter needed here?

alamgircsebd:
> @alamgircsebd Why is this filter needed here?

When a vendor staff try to manage a product that time need check correct product author. Example main vendor created a product and when a staff edit that product then return false, now we checking with this filter parent vendor id by using staff id

PR #963, opened 2020-11-25
Feat/custom deactivation reason

xaviranik:


includes/Tracker.php
```diff
@@ -63,4 +65,62 @@ protected function get_order_count() {
 
         return (int) $wpdb->get_var( "SELECT count(ID) FROM $wpdb->posts WHERE post_type = 'shop_order' and post_status IN ('wc-completed', 'wc-processing', 'wc-on-hold', 'wc-refunded');" );
     }
+
+    /**
+     * Gets custom deactivation reasons
+     *
+     * @since 3.0.15
```
kapilpaul, line 72: please use `@since DOKAN_LITE_SINCE` so it will automatically be replaced with the respected version when released.

kapilpaul (changes requested):

PR #964, opened 2020-11-25
Feature: store customizer and better theme support

tareq1988:
- **New**: Added store customizer
- **Fix**: Duplicate sidebar from theme was showing when storefront was enabled
- **New**: Added twenty twenty theme support for store page
- **New**: Added Astra theme support for store page and dashboard page, it'll look good with `no-sidebar` layout.
- **New**: Added layout support for store page

##  Screenshot

image


tareq1988:
Besides the new feature in the description, some things changed in the CSS/styling of the store page. A new wrapper div added in store page `.dokan-store-wrap`. I've taken a mobile-first approach, what it does is we design a layout that works by default for mobile screens, then we proceed by supporting bigger screen devices. For both left sidebar and right sidebar layout, the sidebar should always display after the content area in mobile view. The sidebar is **26%** wide and the content area is the rest, we should test if the sidebar is the right size for all screen sizes (widescreen and standard laptops).

attn: @shanjidah

FaraziF:
If enable **use theme sidebar**, the store sidebar style is breaking. [image](https://prnt.sc/vr9ed3)

tareq1988:
Due to the changes in the layout from the plugin, the `.col-md-3` class needs to be removed from the [theme store sidebar](https://github.com/weDevsOfficial/dokan-theme/blob/develop/sidebar-store.php#L7).

PR #967, opened 2020-11-27
fix - fixed dokan_save_product() function sanitize issue for post_con…

nurul-umbhiya:
…tent and post_excerpt data #968


nurul-umbhiya:
@shanjidah I think we can skip the testing part for this PR, we are not actually adding anything new here.

PR #972, opened 2020-12-01
fix: Vendor listing and withdraw page not loading from admin area whe…

alamgircsebd:
…n use other langauges issue fixed


xaviranik (commented):
Code looks alright

PR #974, opened 2020-12-01
fix: stock level wrong calculation on order note

### LINKED ISSUE #978: Erroneous order note about reducing stock
When creating an order with two or more vendors, an order note is added both to the main order and the suborders saying that the stock of the product has been reduced. This note is wrong for one of the orders, as the reduction is calculated twice (e.g. 2->1, 1->0 for a product with two in stock). The actual stock reduction is performed correctly.

kapilpaul:
Fixed #978


alamgircsebd (approved):
Codes looks fine

PR #977, opened 2020-12-02
fix: Dokan reports not updated after modification of order amount from admin

kapilpaul:
fix: Reports not updated after modification of order amount from admin


xaviranik (commented):
it looks fine.

FaraziF:
@kapilpaul bhai, is it the right scenario?
If admin updates order amount from admin dashboard like 100 to 120 then updated amount 20 will apply only for vendor earning, not admin commission.

kapilpaul:
@FaraziF bhai, if the admin updates the order amount like 100 to 120, then the order total will be 120. admin will get his commission from the order total and the rest is for the vendor.

FaraziF:
Within the scenario, Admin can not get commissions from the updated amount. please review the [video](https://www.awesomescreenshot.com/video/2013630?key=917d1332e3c726528b4a5c8ebc484148)

FaraziF:
Vendor earning not updated when order amount update from admin.

nurul-umbhiya:
@kapilpaul I'm not merging this pr now. will discuss dependency with payment gateways.

Mohaiminulislam1989:
Invalid.

PR #979, opened 2020-12-03
fix: Upgrade to pro module page overlapping issue with admin notice, css & changed svg

xaviranik:


kapilpaul (commented):
code seems fine.

PR #984, opened 2020-12-06
fix: Store and store TOC template make high priority

alamgircsebd:
It conflicting with Divi theme when use custom Divi builder templates and then happen template_include filter hook priority issue


kapilpaul:
Can you please add some description about the issue and details?

alamgircsebd:
> Can you please add some description about the issue and details?

It conflicting with Divi theme when use custom Divi builder templates and then happen `template_include` filter hook priority issue

kapilpaul (commented):
Code looks fine

PR #985, opened 2020-12-07
fix: Store Settings page url issue when vendor dashboard use as child…

alamgircsebd:
… page


kapilpaul (commented):
everything is fine in code.

PR #989, opened 2020-12-10
fix: Withdraw methods toggle options not working on dokan setup wizard

alamgircsebd:


includes/Admin/SetupWizard.php
```diff
@@ -479,8 +479,8 @@ public function dokan_setup_withdraw() {
                                         ?>
                                     </div>
                                     <div class="dokan-wizard-service-enable">
-                                        <input type="checkbox" name="withdraw_methods[<?php esc_html( $key ); ?>]" id="withdraw_methods[<?php esc_html( $key ); ?>]" class="switch-input" value="<?php esc_html( $key ); ?>" <?php echo ( in_array( $key, array_values( $withdraw_methods ), true ) ) ? 'checked="checked"' : ''; ?>>
-                                        <label for="withdraw_methods[<?php esc_html( $key ); ?>]" class="switch-label"></label>
+                                        <input type="checkbox" name="withdraw_methods[<?php echo esc_html( $key ); ?>]" id="withdraw_methods[<?php echo esc_html( $key ); ?>]" class="switch-input" value="<?php echo esc_html( $key ); ?>" <?php echo ( in_array( $key, array_values( $withdraw_methods ), true ) ) ? 'checked="checked"' : ''; ?>>
```
nurul-umbhiya, line 482: we should use esc_attr instead of esc_html, because we are using this value on a attribute

includes/Admin/SetupWizard.php
```diff
@@ -479,8 +479,8 @@ public function dokan_setup_withdraw() {
...
+                                        <label for="withdraw_methods[<?php echo esc_html( $key ); ?>]" class="switch-label"></label>
```
nurul-umbhiya, line 483: use esc_attr for 'for attribute', secondly why this label value is empty?

PR #994, opened 2020-12-15
New/currency switcher support

alamgircsebd:


includes/wc-functions.php
```diff
@@ -1123,3 +1123,31 @@ function dokan_store_category_delete_transient( $post_id ) {
...
+     * @return bool
+     */
+    function dokan_woocs_recalculate_order_payment_completed( $order_id ) {
+        if ( class_exists( 'WOOCS' ) ) {
+            global $WOOCS;
+            $WOOCS->recalculate_order( $order_id );
+        }
+    }
+    
+    add_action( 'woocommerce_order_status_completed', 'dokan_woocs_recalculate_order_payment_completed', 1 );
```
kapilpaul, line 1144: please move actions outside of the condition.

kapilpaul (changes requested):

FaraziF:
When switch currency, only in new order currency has been changing. Reference: [image](https://prnt.sc/w7rzf6)

FaraziF:
WC Currency Switcher plugin is not working properly even without Dokan. We need to test this plugin first, list the issues and then consider the integration. So we shouldn't merge this pull request now.

PR #998, opened 2020-12-17
fix: withdraw methods are not saving for some users via dokan upgrader

nurul-umbhiya:


alamgircsebd (approved):
Code looks fine

kapilpaul (commented):
code looks fine

PR #1011, opened 2020-12-22
perf: dokan admin settings fast load and optimized

kapilpaul:


includes/Admin/Settings.php
```diff
@@ -162,8 +162,16 @@ public function get_sanitize_callback( $slug = '', $context = 'read' ) {
             return false;
         }
 
+        //settings fields are called every time. so we kept it in cache for a small amount of time. to avoid error and better performance.
+        $settings_fields = get_transient( 'get_settings_fields' );
+
+        if ( ! $settings_fields ) {
+            $settings_fields = $this->get_settings_fields();
+            set_transient( 'get_settings_fields', $settings_fields, 90 );
```
nurul-umbhiya, line 177: can you please change the transient name here? get_settings_fields seems too generic to be used as a key.

PR #1019, opened 2020-12-30
Fix/store product category not showing properly

alamgircsebd:


includes/template-tags.php
```diff
@@ -552,7 +552,14 @@ function dokan_store_category_menu( $seller_id, $title = '' ) {
         $all_terms = [];
         foreach ( $product_ids as $product_id ) {
             $terms = get_the_terms( $product_id, 'product_cat' );
-            array_push( $all_terms, $terms[0] );
+
+            if ( count( $terms ) > 1 ) {
```
kapilpaul, line 588: do we need this check? we can easily loop through the terms if there is one or more. we need to just push the term in the array.

kapilpaul (changes requested):

nurul-umbhiya:
@alamgircsebd we need to check the return type of get_the_terms() before running foreach loop. For now, I'm merging this PR. @kapilpaul kindly fix this issue and make a new pr

`
            foreach ( $terms as $term ) {
                array_push( $all_terms, $term );
            }
`

Issue #997, opened 2020-12-17
The vendor can't change the display name


FaraziF:
**Scenario:** When vendor change display name then display name automatically switches to store name. Please review the [video](https://www.loom.com/share/ac9647935d204f46ba7af324442c2db0)

alamgircsebd:
Did not reproduce this issue correctly. check it https://www.loom.com/share/dfbf5aa98cb8450ebb104d84adbe0d45

FaraziF:
Please try to reproduce as a vendor, not admin

nurul-umbhiya:
@alamgircsebd first change store name of a vendor. then try to change the display name from My Account page for that vendor.

PR #1025, opened 2021-01-05
The vendor can not change the display name

asaquzzaman:


FaraziF:
The vendor can't change the display name fixed #997

Issue #941, opened 2020-10-24
REST API order total value gets rounded


rashedripon:
**Description of the bug:**
When an order data is retrieved using **.../wp-json/dokan/v1/orders/<id>** the "total" order value is rounded. One the other hand, WooCommerce's response shows the correct value.

**How to reproduce:**
1. Make an order with fraction value - https://prnt.sc/v67i7y
2. Retrieve the order using **.../wp-json/dokan/v1/orders/<id>** - https://prnt.sc/v67ika (the value is rounded from 9.55 to 10)
3. Try the same with WooCommerce **/wp-json/wc/v3/orders/<id>** - https://prnt.sc/v67jba

**Expected behavior:**
The total value should not be rounded and show as the WooCommerce shows. The item price value is correct in Dokan.

**Versions:**
Dokan Lite - 3.0.11
WC - 4.6.1

PR #1028, opened 2021-01-07
Fix: REST API order total value gets rounded

hasinur1997:
When an order data is retrieved using `.../wp-json/dokan/v1/orders/` the "total" order value is rounded. One the other hand, WooCommerce's response shows the correct value.

Fix: #941

### DIFF
--- a/includes/REST/OrderController.php
+++ b/includes/REST/OrderController.php
@@ -235,7 +235,7 @@ class OrderController extends DokanRESTController {
 
         // Format decimal values.
         foreach ( $format_decimal as $key ) {
-            $data[ $key ] = wc_format_decimal( $data[ $key ], $this->request['dp'] );
+            $data[ $key ] = wc_format_decimal( $data[ $key ], wc_get_price_decimals() );
         }
 
         // Format date values.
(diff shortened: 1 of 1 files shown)


kapilpaul (commented):
code looks fine

includes/REST/OrderController.php
```diff
@@ -235,7 +235,7 @@ protected function get_formatted_item_data( $object ) {
 
         // Format decimal values.
         foreach ( $format_decimal as $key ) {
-            $data[ $key ] = wc_format_decimal( $data[ $key ], $this->request['dp'] );
+            $data[ $key ] = wc_format_decimal( $data[ $key ], wc_get_price_decimals() );
```
xaviranik, line 238: No need to use `wc_get_price_decimals()`, instead use empty string.

includes/REST/OrderController.php
```diff
@@ -259,7 +259,7 @@ protected function get_formatted_item_data( $object ) {
             $data['refunds'][] = array(
                 'id'     => $refund->get_id(),
                 'refund' => $refund->get_reason() ? $refund->get_reason() : '',
-                'total'  => '-' . wc_format_decimal( $refund->get_amount(), $this->request['dp'] ),
+                'total'  => '-' . wc_format_decimal( $refund->get_amount(), wc_get_price_decimals() ),
```
xaviranik, line 262: No need to use `wc_get_price_decimals()`, instead use empty string.

PR #1031, opened 2021-01-08
fix: spmv redirection for auction and booking type product

xaviranik:


xaviranik:
I am closing this PR because the fixes have been given to specific module.

Issue #1030, opened 2021-01-07
Rounded vendor balance can not be withdrawn


rashedripon:
**Description:**
If the vendor has a rounded value in their balance then you are unable to request for a withdrawal of the full amount as it shows this error - https://prnt.sc/wi7fpr

**How to reproduce:**
1. Under wp-admin > WooCommerce > Settings > General set the "Number of decimals" as 0 - https://prnt.sc/wi7gnw
2. Set your admin commission and product in a way that the earning decimal is rounded. For example -
Commission - 10%, Product price - 1,255, 
Earning - **1,129.50 (real)** || **1,130** (showed on dashboard because of rounding - https://prnt.sc/wi7izl
3. Complete the order and now try to withdraw the whole balance
4. It will not work because the actual balance is $ .50 short because of rounding

**Expected behavior:**
Withdraw balance should show the non-rounded value so that the full balance can be withdrawn and vendors are not confused.

**Versions:**
WC - 4.8.0
WP - 5.6
Dokan Lite - 3.1.0
Dokan Pro - 3.1.3

PR #1037, opened 2021-01-11
Fix: vendor withdraw balance

hasinur1997:
If the vendor has a rounded value in their balance then you are unable to request for a withdrawal of the full amount as it shows an error.

Fix: #1030


kapilpaul (commented):
code is fine

Rink9:
@hasinur1997 QA Approved

PR #1042, opened 2021-01-12
fix: Product REST API is not returning products with "Pending" & "Dra…

### LINKED ISSUE #1041: Product REST API is not returning products with "Pending" & "Drafts"  statuses
hitting this endpoint ('/wp-json/dokan/v1/products/') only returns products with published status. Products with "Pending Review" & "Draft" statuses are only returned while doing a query ('/wp-json/dokan/v1/products?=status') .

iftekkhar:
…fts" statuses REST API

fixed #1041


kapilpaul (commented):
this code is fine.

PR #1045, opened 2021-01-13
fix: Rounded vendor balance can not be withdrawn issue fixed #1030

### LINKED ISSUE #1030: Rounded vendor balance can not be withdrawn

alamgircsebd:


Rink9:
QA Approved.

Rink9:
QA Approved

PR #1051, opened 2021-01-15
fix: vendor pending tab keeps loading

xaviranik:
This PR fixes https://github.com/weDevsOfficial/dokan-pro/issues/1008

### DIFF
--- a/includes/REST/StoreController.php
+++ b/includes/REST/StoreController.php
@@ -249,7 +249,7 @@ class StoreController extends WP_REST_Controller {
             $args['meta_query'] = [
                 [
                     'key'     => 'dokan_store_name',
-                    'value'   => sanitize_text_field( $params['search'] ),
+                    'value'   => isset( $params['search'] ) ? sanitize_text_field( $params['search'] ) : '',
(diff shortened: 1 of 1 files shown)


includes/REST/StoreController.php
```diff
@@ -249,7 +249,7 @@ public function get_stores( $request ) {
             $args['meta_query'] = [
```
nurul-umbhiya, line 249: do we really need this meta query if the search item is empty ? @kapilpaul @xaviranik

nurul-umbhiya (changes requested):

includes/REST/StoreController.php
kapilpaul: If no stores found in then we are searching again with meta value for store name. Here, we need to remove search and search_columns, because with this args meta_query is not working.

nurul-umbhiya: what I meant was, if user didn't search for anything, do we need this bit of code? 
`'value'   => isset( $params['search'] ) ? sanitize_text_field( $params['search'] ) : '',`
if search param is empty we can skip I guess?

PR #1052, opened 2021-01-15
Fix:Site registration as vendor ghost touch fixes #998

shams-sadek1981:
Radio button should work only when user click mouse cursor on the top of the radio btn.

fix https://github.com/weDevsOfficial/dokan-pro/issues/998


templates/global/seller-registration-form.php
```diff
@@ -8,63 +8,75 @@
  */
```
nurul-umbhiya, line 8: Need to change the formatting for this file.

FaraziF:
@shams-sadek1981 Unexpected height

Screenshot 2021-01-21 at 4 56 32 PM

PR #1059, opened 2021-01-21
update: Add new filter hook for store per page number

alamgircsebd:


alamgircsebd:
Add a setting where admin can set number of products show on store page

Screenshot (3)

PR #1066, opened 2021-01-25
fix: Update Setting button showing all setting page issue fixed

alamgircsebd:


includes/Dashboard/Templates/Settings.php
```diff
@@ -53,16 +53,18 @@ public function show_enable_seller_message() {
      */
     public function render_settings_header() {
         global $wp;
+        $is_main_setting = 0;
```
kapilpaul, line 56: Variable name `is_store_setting` is more relevant. Please use true or false for boolean.

kapilpaul (changes requested):

PR #1079, opened 2021-02-02
fix: product tab's post per page in more products

kapilpaul:


shohag121:
Here is the screenshot of the issue.

Test-product-–-wD

nurul-umbhiya (approved):

PR #1082, opened 2021-02-03
feat: Vendors able to edit product slug from their product edit page

alamgircsebd:
Screenshot(3)


alamgircsebd:
@shanjidah apu you can test this PR we will release this features next version inshallah

includes/Dashboard/Templates/Products.php
```diff
@@ -387,6 +387,7 @@ public function handle_product_update() {
 
         $errors     = array();
         $post_title = sanitize_text_field( $postdata['post_title'] );
+        $post_slug  = isset( $postdata['editable-post-name'] ) && ! empty( $postdata['editable-post-name'] ) ? sanitize_text_field( $postdata['editable-post-name'] ) : '';
```
nurul-umbhiya, line 390: use sanitize_title() instead of sanitize_text_field()

includes/Dashboard/Templates/Products.php
```diff
@@ -426,6 +427,20 @@ public function handle_product_update() {
                 'post_status'    => isset( $postdata['post_status'] ) ? sanitize_text_field( $postdata['post_status'] ) : 'pending',
                 'comment_status' => isset( $postdata['_enable_reviews'] ) ? 'open' : 'closed',
             ) );
+            
+            if ( $post_slug ) {
```
nurul-umbhiya, line 431: use wp_unique_post_slug() function to generate unique slug,

nurul-umbhiya (changes requested):

includes/Dashboard/Templates/Products.php
alamgircsebd: This is normal text field where have slug, this is not title, so we should use sanitize_text_field

nurul-umbhiya: sanitize_title is used to sanitize slug

nurul-umbhiya (approved):

PR #1089, opened 2021-02-12
fix: fixes nonce checking for some functions

nurul-umbhiya:


includes/Dashboard/Templates/Products.php
```diff
@@ -375,16 +371,12 @@ public function handle_product_update() {
             return;
         }
 
-        if ( ! isset( $_POST['dokan_update_product'] ) ) {
+        if ( ! isset( $_POST['dokan_update_product'] ) || ! wp_verify_nonce( sanitize_key( $_POST['dokan_edit_product_nonce'] ), 'dokan_edit_product' ) ) {
```
kapilpaul, line 374: here, we are checking 'dokan_update_product' and in wp_verify_nonce we are checking 'dokan_edit_product_nonce'. I think this is a mismatch.

includes/functions.php
```diff
@@ -2274,7 +2274,7 @@ function dokan_product_search_by_sku( $where ) {
 
     $getdata = wp_unslash( $_GET );
 
-    if ( ! isset( $getdata['product_search_name'] ) || empty( $getdata['product_search_name'] ) || ! isset( $getdata['dokan_product_search_nonce'] ) || ! wp_verify_nonce( wc_clean( $getdata['dokan_product_search_nonce'] ), 'dokan_product_search' ) ) {
+    if ( ! isset( $getdata['product_search_name'] ) || empty( $getdata['product_search_name'] ) || ! isset( $getdata['dokan_product_search_nonce'] ) || ! wp_verify_nonce( sanitize_key( $getdata['dokan_product_search_nonce'] ), 'dokan_product_search' ) ) {
```
kapilpaul, line 2277: Do we need to check 'not isset' and 'empty' both?

PR #1135, opened 2021-03-22
fix: Select2 block the input field in dropdown

shohag121:
When the user trying to search or create a tag on the product page the tag field is getting blocked by the dropdown result.

fix: https://github.com/weDevsOfficial/dokan-pro/issues/1172


nurul-umbhiya (approved):

danielbreves:
This seems to cause a big gap below the dropdown. Not sure what was the original issue it was trying to fix:

Screen Shot 2021-11-13 at 4 59 26 pm

cc @shohag121 @nurul-umbhiya

CasperBraske:
> This seems to cause a big gap below the dropdown. Not sure what was the original issue it was trying to fix:

I have an add-on for Dokan that uses Select2 and I can confirm this.

image

nurul-umbhiya:
@rajankarmakar can you please look into this issue?

rajankarmakar:
The issue will be fixed once the PR will be marged.
https://github.com/weDevsOfficial/dokan/pull/1446

CasperBraske:
@rajankarmakar  It looks like it will fix the issue in the product edit page, only. Not elsewhere.

PR #1177, opened 2021-04-19
Fix: product count exclude booking product

shohag121:
Product count is displaying count number including `product_type = booking` in this PR booking product are excluded.

fix: https://github.com/weDevsOfficial/dokan-pro/issues/1229

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -26,7 +26,7 @@ function dokana_admin_menu_capability() {
  *
  * @since 2.7.3
  *
- * @return void
+ * @return int
  */
 function dokan_get_current_user_id() {
     if ( current_user_can( 'vendor_staff' ) ) {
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -186,7 +186,7 @@ function dokan_redirect_if_not_seller( $redirect = '' ) {
  *
  * @return array
  */
-function dokan_count_posts( $post_type, $user_id ) {
+function dokan_count_posts( $post_type, $user_id, $exclude_product_type = 'booking' ) {
```
nurul-umbhiya, line 189: make exclude_product_type to array and change code accordingly. in the future, we might need to exclude auction products also.

shohag121: Code changed as you instructed.

FaraziF:
We can bring items counting on product listing page on frontend like backend https://prnt.sc/11ad5af

includes/functions.php
```diff
@@ -183,23 +183,35 @@ function dokan_redirect_if_not_seller( $redirect = '' ) {
...
  */
-function dokan_count_posts( $post_type, $user_id ) {
+function dokan_count_posts( $post_type, $user_id, $exclude_product_types = array( 'booking' ) ) {
     global $wpdb;
 
-    $cache_group = 'dokan_seller_product_data_' . $user_id;
-    $cache_key   = 'dokan-count-' . $post_type . '-' . $user_id;
-    $counts      = wp_cache_get( $cache_key, $cache_group );
-
+    $exclude_product_types      = esc_sql( $exclude_product_types );
```
nurul-umbhiya, line 193: if you are using cache, you need to delete cache.

shohag121: Code changed as per your instruction.

nurul-umbhiya (approved):

PR #1203, opened 2021-05-12
feat(api): Add support to send objects to trash

### LINKED ISSUE #1202: DELETE is "forced" on API even with param described as default to false
**Is your feature request related to a problem? Please describe.**

There is no information about "force" param on DELETE in products [on documentation](https://wedevsofficial.github.io/dokan/#delete-a-product). Said that, WooCommerce API Rest provide us this feature and the Dokan REST registration has this parameter too ([here](https://github.com/weDevsOfficial/dokan/blob/develop/includes/REST/ProductController.php#L103)).

But in the abstract class **delete_item** method [forces the deletion](https://github.com/weDevsOfficial/dokan/blob/develop/includes/Abstracts/DokanRESTController.php#L187).

I'm not sure this is a feature request or a bug (as the param is registered).

**Describe the solution you'd like**

The force param should be accepted.
Maybe true as default to backward compatibility.

**Describe alternatives you've considered**

There is no alternative.

**Additional context**

There is no additional context.

mariovalney:
Add support to force parameter on API requests.
Use false as default because backward compatibility.

Closes #1202

### DIFF
--- a/includes/Abstracts/DokanRESTController.php
+++ b/includes/Abstracts/DokanRESTController.php
@@ -185,7 +185,9 @@ abstract class DokanRESTController extends WP_REST_Controller {
         $response = rest_ensure_response( $data );
 
         // If we're forcing, then delete permanently.
-        $object->delete( true );
+        $force = ( false !== $request['force'] );
+        $object->delete( $force );
+
         $result = 0 === $object->get_id();
 
         if ( ! $result ) {
--- a/includes/Abstracts/DokanRESTController.php
+++ b/includes/Abstracts/DokanRESTController.php
@@ -185,8 +185,7 @@ abstract class DokanRESTController extends WP_REST_Controller {
         $response = rest_ensure_response( $data );
 
         // If we're forcing, then delete permanently.
-        $force = ( false !== $request['force'] );
-        $object->delete( $force );
+        $object->delete( $request['force'] );
 
         $result = 0 === $object->get_id();
 
--- a/includes/REST/ProductController.php
+++ b/includes/REST/ProductController.php
@@ -103,7 +103,7 @@ class ProductController extends DokanRESTController {
                     'args'                => [
                         'force' => [
                             'type'        => 'boolean',
-                            'default'     => true,
+                            'default'     => false,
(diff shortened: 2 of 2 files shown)


includes/REST/ProductController.php
```diff
@@ -103,7 +103,7 @@ public function register_routes() {
                     'args'                => [
                         'force' => [
                             'type'        => 'boolean',
-                            'default'     => false,
+                            'default'     => true,
```
nurul-umbhiya, line 106: do not make this value default to true, leave this as false

includes/Abstracts/DokanRESTController.php
```diff
@@ -185,7 +185,9 @@ public function delete_item( $request ) {
         $response = rest_ensure_response( $data );
 
         // If we're forcing, then delete permanently.
-        $object->delete( true );
```
nurul-umbhiya, line 188: instead of checking for true value, send the argument of this method as $request['force']

nurul-umbhiya (changes requested):

includes/REST/ProductController.php
mariovalney: This will break backward compatibility as the current value is not remove.
Are you sure?

includes/Abstracts/DokanRESTController.php
mariovalney: Done!

includes/REST/ProductController.php
nurul-umbhiya: the default value should be false as it was before, if you want to force delete, send this value as true from your API request, set the default value to false as it was before.

nurul-umbhiya: We don't have to worry about backward compatibility, because only places we used force value with Product API.

mariovalney: Yes! My mistake: I was advocating about keeping the compatibility but wrote the opposite...
Fixed!

nurul-umbhiya (approved):

nurul-umbhiya (approved):

PR #1215, opened 2021-05-31
New/two new options added with stores listing short-codes

alamgircsebd:
2 new options added on stores listing short-codes

1. Now we can show the stores which one have at least 1 product, example: `[dokan-stores with_products_only="yes"]`

### DIFF
--- a/includes/Rewrites.php
+++ b/includes/Rewrites.php
@@ -334,7 +334,7 @@ class Rewrites {
     /**
      * Store listing page make order by random
      *
-     * @since DOKAN_PRO_SINCE
+     * @since DOKAN_LITE_SINCE
      *
      * @param obj $class
      *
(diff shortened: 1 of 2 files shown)


includes/Rewrites.php
```diff
@@ -329,4 +330,25 @@ public function store_query_filter( $query ) {
...
+     *
+     * @return obj $class
+     */
+    public function random_store_query( $class ) {
+        if ( ! dokan_is_store_listing() ) {
+            return $class;
+        }
+
+        if ( 'random' === $class->query_vars['orderby'] ) {
+            $class->query_orderby = str_replace( 'user_login', 'RAND()', $class->query_orderby );
```
nurul-umbhiya, line 349: using RAND() will give you random results each time, this will mess up pagination, you can't use it like that, kindly remove this code.

nurul-umbhiya (approved):

PR #1228, opened 2021-06-07
new: Able to stores search with first, last and display name

alamgircsebd:


Tangol:
Sorry, but I would have to go against this PR being merged.
As reported here https://github.com/weDevsOfficial/dokan/issues/1143, using the First and Last name of the store owner violates some privacy rules and it's not a good practice for several reasons.

I'll give you an example of what happened a while back, in my country: 2 vendors were selling mostly the same products in a very competitive way, but one of them was doing a far better job. The other went crazy, and by knowing the 1st vendor name and location, he proceeded in tracking him physically, then he threatened him on the street in front of his wife and child...

What does this PR do?
It gives bad guys another tool to find out a store owner's name.

We as marketplaces, need to prevent this by ensuring the personal information stays private.
If you think about it, it's outrageous to post the full name of the vendor alongside the store name.
It's like you're having Microsoft as a store and Bill Gates as a vendor...

alamgircsebd:
Our product manage will R&D on this PR before final merge

alamgircsebd:
No longer need the PR

PR #1260, opened 2021-06-28
fix: flush rewrite rules after plugin is activated

nurul-umbhiya:
Permalink Issues after Dokan Plugin is activated.
Permalink issue after settings is saved


FaraziF:
All expected scenario is working.

PR #1262, opened 2021-06-29
add a filter before return the orders

CODLOP:
add a filter before return the orders to open the way to add custom features such as assigning a worker to monitor specific orders etc.

### DIFF
--- a/includes/Order/Manager.php
+++ b/includes/Order/Manager.php
@@ -69,6 +69,8 @@ class Manager {
                 }, $orders
             );
 
+            $orders = apply_filters('dokan_get_vendor_orders', $orders, $args);
+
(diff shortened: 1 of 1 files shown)


includes/Order/Manager.php
```diff
@@ -69,6 +69,8 @@ function( $order_data ) {
                 }, $orders
             );
 
+            $orders = apply_filters( 'dokan_get_vendor_orders', $orders, $args );
```
nurul-umbhiya, line 72: remove this filter from here and add this in line 78,

nurul-umbhiya:
fixed here: https://github.com/weDevsOfficial/dokan/commit/52bb9c6ee4e929f9a69e5a345b728baa378da89e

PR #1275, opened 2021-07-13
Order item price

79mplus-admin:
adding a filter for modifying the order item's price when calculating earnings for the ordered item.


iammahmudur:
@nayemDevs vai

- can you check if this is possible to merge this pull request? https://github.com/weDevsOfficial/dokan/pull/1275
- few clients are asking about this and we have to make a change for this on the client site for a client's project

nurul-umbhiya (approved):

PR #1281, opened 2021-07-26
fix: spelling of "picture" across files

twmbx:
changed "picutre" to "picture"


nurul-umbhiya:
hello @twmbx, thanks for contributing to Dokan,
you only need to edit src/admin/pages/VendorAccountFields.vue, skip other files, and make a new commit.

Issue #1288, opened 2021-08-10
No Shipping Address in The Vendor New Order Email When Multiple Vendors' Products in An Order.


shafinahmad01:
Describe the bug:

If multiple vendors' products are ordered from the same cart then there's no shipping address on the Vendor New Customer Order email template, even if both the shipping and billing address was provided during the checkout. (Dokan Lite). The issue is doesn't happen when Dokan Pro is enabled or with only one vendor's products are ordered. 

How To Reproduce: 

Steps to Reproduce the Issue:
1. Enable any email log plugin on the site
2. Then add multiple vendors' products while only Dokan Lite is enabled
3. Complete the checkout with both the Billing & Shipping Address
4. Now check the email log you will see no shipping address is sent to the vendor

Expected behavior:

If Dokan Pro is also enabled then that issue doesn't appear and both shipping and billing address gets sent to the vendor even if there are multiple products on the cart.

Screen Records:

Here's a screen record to demonstrate the issue https://www.loom.com/share/4c477b92b8d149c199be632443af210e

Environment:
 - Dokan Version: v3.2.9
 - Dokan Pro Version: v3.3.3
 - WooCommerce Version: v5.5.1
 - Browser: chrome, firefox
 - Current Theme: Dokani

nurul-umbhiya:
This is the default Dokan behavior. Dokan Lite doesn't support multiple shipping methods for each vendor. This feature is available under Dokan Pro. I'm closing this issue for now.

PR #1297, opened 2021-08-18
fix: shipping is not creating for sub orders, for this shipping address is missing in  customer order details, vendor order email

shohag121:
We were checking for shipping methods associated with the seller during creating a suborder shipping.

Currently, we are not associating any shipping methods with any seller in Dokan lite.

Here we are checking if the pro is activated during the comparison for the shipping method associated with the seller.

Fix: #1288


xaviranik (approved):

alamgircsebd (approved):

Mohaiminulislam1989:
@shohag121 vai what is the last status of this PR?

nurul-umbhiya:
As discussed, we need to move pro shipping split feature to lite,  otherwise  this won't work

PR #1312, opened 2021-08-27
Fix/js deprecated check

shohag121:


Rink9:
Retested. Fixed all given issues.

PR #1327, opened 2021-09-14
fix vendor info issue 1143

### LINKED ISSUE #1143: Vendor Info
Please note that the **Vendor Info** tab shows the actual name of the vendor which is extremely unprofessional, and violates some privacy rules.
Most vendors will be companies/brands, so showing the company/brand name is enough. There's no need for the whole world to see the owner's name.

vdrname

So the professional way would be to just have **Vendor:** `Julia's` with Julia's linking to the vendor's store.

abdulhadicse:
The single product page displays the seller's real name on the vendor info tab. 
https://prnt.sc/1s8w0xd

Fix #1143


templates/global/product-tab.php
```diff
@@ -20,17 +20,17 @@
...
+            <span>
+                <?php esc_html_e( 'Vendor:', 'dokan-lite' ); ?>
+            </span>
 
-    <li class="seller-name">
-        <span>
-            <?php esc_html_e( 'Vendor:', 'dokan-lite' ); ?>
-        </span>
+            <span class="details">
+                <?php printf( '<a href="%s">%s</a>', esc_url( dokan_get_store_url( $author->ID ) ), esc_attr( $store_info['store_name'] ) ); ?>
```
shohag121, line 29: kindly use numbered placeholder like `%1$s`, use the correct escape function for the store name.

templates/global/product-tab.php
```diff
@@ -20,17 +20,17 @@
...
+            </span>
+        </li>
+    <?php } ?>
```
shohag121, line 32: use appropriate php condition syntax for HTML use like `<?php if () : ?>` and `<?php endif; ?>` where necessary.

shohag121 (changes requested):

shohag121:
Kindly remove `package-loc.json` from this PR.

Issue #1283, opened 2021-07-31
window.confirm is going progressively removed from the platform


oliviertassinari:
**Describe the bug**

The Dokan vendor dashboard uses the native UI dialogs like `window.confirm`. However, the web standard working group is pushing toward removing them from the web platform: https://github.com/whatwg/html/issues/6897#issuecomment-885773622.

The usage is also problematic as we can no longer embed the Dokan dashboard in an iframe since https://www.chromestatus.com/feature/5148698084376576.

**To Reproduce**

https://github.com/weDevsOfficial/dokan/blob/0eb5b9a610e704cec35ec271074528b3bfde0a51/assets/src/js/orders.js#L338

**Expected behavior**

Use custom widgets, like https://material-ui.com/components/dialogs/#confirmation-dialogs.

**Environment (please complete the following information):**

 - Dokan Version: v3.2.8

shohag121:
@abdulhadicse @devasadnur

PR #1328, opened 2021-09-14
update: integration of sweetalert2 for confirmation, alert

abdulhadicse:
Previously we used `window.alert` and `window.confirm` for confirmation alert system, which is going to be deprecated in near future.

So, in this branch, we have integrated **sweetalert2** for confirmation alert.

In a collaboration with @devAsadNur 

Fix: #1283


xaviranik (changes requested):

assets/src/js/helper.js
```diff
@@ -40,3 +40,36 @@ function dokan_get_i18n_date_format( format = true ) {
...
   return datepickerFormat;
 }
+
+/**
+ * Dokan Sweet Alert
+ *
+ * @since DOKAN_LITE
+ *
+ * @param {string} message The event message for notification
+ * @param {object} [options] Configuration of sweet alert.
```
xaviranik, line 50: Kindly write {object} [options] to elaborate which parameters are expected in the object.

assets/src/js/helper.js
```diff
@@ -40,3 +40,36 @@ function dokan_get_i18n_date_format( format = true ) {
 
   return datepickerFormat;
 }
+
+/**
+ * Dokan Sweet Alert
+ *
+ * @since DOKAN_LITE
```
xaviranik, line 47: Kindly remove DOKAN_LITE and manually manage the version

xaviranik (changes requested):

assets/src/js/orders.js
```diff
@@ -332,10 +332,12 @@ jQuery(function($) {
 
         refunds: {
 
-            do_refund: function() {
+            do_refund: async function() {
                 dokan_seller_meta_boxes_order_items.block();
+                
+                let isRefund = await dokan_sweet_alert( dokan_refund.i18n_do_refund, { action : 'confirm' }  );
```
ManiruzzamanAkash, line 338: if the variable used only once and not being updated, use `const` instead of `let`

assets/src/js/product-editor.js
```diff
@@ -515,10 +515,12 @@
                 });
             },
 
-            removeAttribute: function(evt) {
+            removeAttribute: async function(evt) {
+                evt.preventDefault()
                 evt.stopPropagation();
 
-                if ( window.confirm( dokan.remove_attribute ) ) {
+                let isRemoved = await dokan_sweet_alert( dokan.remove_attribute, { action : 'confirm' } );
```
ManiruzzamanAkash, line 522: Same here, use `const`, instead of `let` 
since, it doesn't updated anywhere.

PR #1329, opened 2021-09-15
feat: Set limitation for how many product tags that vendor can input

### LINKED ISSUE #1236: set limitation for how many product tags that vendor can input
There is no limits for how many tags that vendor can input, vendor can input endless tags for their product,  this will create mess on the system and no good for the filtering.  setting for how many tags that vendor can input for each of the product is kind  of essential.

Aunshon:
Previously, vendor could add unlimited tags on his/her product in add, edit product.

But now I have added a new filter called `dokan_maximum_tags_select_length` by which maximum tag addition number can be changed. 

Also added validation in frontend and backend for maximum tag add and edit
frontend validation: https://prnt.sc/1s8veg8
Backend validation: https://prnt.sc/1s8vspn

fixes #1236 Similar code fix in dokan pro https://github.com/weDevsOfficial/dokan-pro/pull/1471

### DIFF
--- a/assets/js/dokan.js
+++ b/assets/js/dokan.js
@@ -785,6 +785,7 @@ jQuery(function($) {
                     if ( ! $found ) data.unshift( tag );
                 },
                 minimumInputLength: 2,
+                maximumSelectionLength: dokan.maximum_tags_select_length !== undefined ? dokan.maximum_tags_select_length : -1,
                 ajax: {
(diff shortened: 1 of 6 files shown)


leew999:
Can this be added into the next update? without this, vendors are just creating hundreds tags into the system.

nurul-umbhiya (approved):

PR #1335, opened 2021-09-22
Fix withdraw limit validation

### LINKED ISSUE #1331: withdraw option field is not validate, validation is required
**Describe the bug**
The vendor is able to withdraw the negative amount and the amount is adding the vendor account after the withdraw.
**How to Reproduce:**
1.Go to wp-admin ->dokan ->settings ->withdraw option.
2.admin set minimum withdraw limit at negative amount.
3.Go to vendor dashboard.
3.vendor provide withdraw amount at negative amount.
4.submit request button.

**Attachment:**
**video link**:https://www.loom.com/share/9e7a2c806f724a88b3d4078aed5b7cc0

**Expected Result:** Negative value cannot be withdraw. It will show validation message.
**Actual Result:** I have given negative value in withdraw option, after withdrawing that value is being credited to account.

**Bug type:** Functional
**Dokan Version**: 3.2.11
**Dokan pro Version**: 3.3.6
**WooCommerce Version:** 5.6.0
**Wordpress Version**: 5.8.1

abdulhadicse:
The seller is able to withdraw the negative amount and the amount is added to the seller's account after the withdrawal. 
Now the admin has not set the negative withdrawal limit amount.

fix #1331


includes/functions.php
```diff
@@ -4119,3 +4119,22 @@ function dokan_get_withdraw_threshold( $user_id ) {
...
+ * @param mixed $option_value
+ * @param mixed $option_name
+ *
+ * @since DOKAN_LITE_SINCE
+ *
+ * @return mixed $optins
+ */
+function dokan_withdraw_limit_value_validation( $option_value, $option_name ) {
+    if( $option_value['withdraw_limit'] < 0 ) {
+        $option_value['withdraw_limit'] = 0;
```
ManiruzzamanAkash, line 4135: This should be re-written proper way. and need to fix formatting. Do you exactly know, you could get `'withdraw_limit'` there. Cause, it's a global filter hook for settings every option's save.

This could be re-written as below - 
```php
if ( ! empty( $option_value['withdraw_limit'] ) ) {
    $option_value['withdraw_limit'] = $option_value['withdraw_limit'] < 0 ? 0 : $option_value['withdraw_limit'];
}
```

PR #1336, opened 2021-09-22
fix/withdraw-limit-amount-validation

### LINKED ISSUE #1331: withdraw option field is not validate, validation is required

abdulhadicse:
The vendor is able to withdraw the negative amount and the amount is adding the vendor account after they withdraw.

fix #1331

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -4119,3 +4119,22 @@ function dokan_get_withdraw_threshold( $user_id ) {
 
     return ( $threshold_day ) ? absint( $threshold_day ) : 0;
 }
+
+/**
+ * Dokan Withdraw Limit Value Validation
+ *
+ * @param mixed $option_value
+ * @param mixed $option_name
+ *
+ * @since DOKAN_LITE_SINCE
+ *
+ * @return mixed $option_value
+ */
+function dokan_withdraw_limit_value_validation( $option_value, $option_name ) {
+    if( $option_value['withdraw_limit'] < 0 ) {
+        $option_value['withdraw_limit'] = 0;
+    }
+    return $option_value;
+}
+
+add_filter( 'dokan_save_settings_value', 'dokan_withdraw_limit_value_validation' , 10, 2 );


includes/functions.php
```diff
@@ -4119,3 +4119,22 @@ function dokan_get_withdraw_threshold( $user_id ) {
...
     return ( $threshold_day ) ? absint( $threshold_day ) : 0;
 }
+
+/**
+ * Dokan Withdraw Limit Value Validation
+ *
+ * @param mixed $option_value
+ * @param mixed $option_name
+ *
+ * @since DOKAN_LITE_SINCE
```
xaviranik, line 4129: Kindly remove DOKAN_LITE_SINCE and manually put the version here. Currently, the DOKAN_LITE_SINCE  replacement script is not working.

includes/functions.php
```diff
@@ -4119,3 +4119,22 @@ function dokan_get_withdraw_threshold( $user_id ) {
...
+ *
+ * @return mixed $option_value
+ */
+function dokan_withdraw_limit_value_validation( $option_value, $option_name ) {
```
xaviranik, line 4133: By considering the issue, we also need to fix the issue on the admin panel where the admin sets the withdrawal amount to a negative value. Currently, the admin can set the withdrawal amount in two places.

- Vendor setup wizard
- Dokan admin panel

Let me know if there's any confusion.

xaviranik (changes requested):

xaviranik (approved):

PR #1337, opened 2021-09-22
fix: dokan product view cache plugin/system bug

RatulHasan:
fix: 1199

### DIFF
--- a/includes/PageViews.php
+++ b/includes/PageViews.php
@@ -23,12 +23,20 @@ class PageViews {
 
         echo '<script type="text/javascript">
             jQuery(document).ready( function($) {
-                var data = {
-                    action: "dokan_pageview",
-                    _ajax_nonce: "' . esc_html( $nonce ) . '",
-                    post_id: ' . get_the_ID() . ',
-                }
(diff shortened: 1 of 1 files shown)


includes/PageViews.php
```diff
@@ -23,12 +23,20 @@ public function load_scripts() {
...
+                    if (new_date != old_date){
+                        var data = {
+                            action: "dokan_pageview",
+                            _ajax_nonce: "' . esc_html( $nonce ) . '",
+                            post_id: ' . get_the_ID() . ',
+                        }
+                        $.post( "' . esc_url( admin_url( 'admin-ajax.php' ) ) . '", data );
+                        localStorage.setItem("dokan_product_viewed_' . get_the_ID() . '", new_date);
+                    }
+				}
```
alamgircsebd, line 39: What happens for existing users? which one used COOKIE

alamgircsebd (changes requested):

includes/PageViews.php
RatulHasan: I'm trying to make an upgrader file asap as you told me.

RatulHasan: This will not make any big effects on our existing users.

alamgircsebd (approved):

PR #1345, opened 2021-09-25
Feat: Added dokan stores menu in admin bar

MdAsifHossainNadim:
### Added dokan stores menu in admin bar

- Updated woocommerce shop menu title in admin bar.
- Added dokan stores menu in admin bar.
- Added "get_dokan_admin_menus()" method in AdminBar.php.
- Added wp admin menus via foreach loop in "visit_dashboard_menu()" method.

### DIFF
--- a/includes/Admin/AdminBar.php
+++ b/includes/Admin/AdminBar.php
@@ -92,6 +92,38 @@ class AdminBar {
     }
 
     /**
+     * Get admin menus data for dokan.
+     *
+     * @since 3.2.13
+     *
+     * @return array
+     */
+    public function get_dokan_admin_menus() {
+        $menus = [
+            [
+                'parent' => 'site-name',
+				'id'     => 'view-store',
+				'title'  => __( 'Visit Shop', 'dokan-lite' ),
+				'href'   => wc_get_page_permalink( 'shop' ),
+            ],
+            [
+                'parent' => 'site-name',
+                'id'     => 'view-dokan-stores',
+                'title'  => __( 'Visit Dokan Stores', 'dokan-lite' ),
+                'href'   => home_url( '/store-listing' ),
+            ],
+            [
+                'parent' => 'site-name',
+                'id'     => 'view-dashboard',
+                'title'  => __( 'Visit Vendor Dashboard', 'dokan-lite' ),
+                'href'   => dokan_get_navigation_url(),
+            ],
+        ];
+
+        return $menus;
+    }
+
+    /**
      * Show visit vendor dashboard
      *
      * @param WP_Admin_Bar $wp_admin_bar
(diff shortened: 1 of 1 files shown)


includes/Admin/AdminBar.php
```diff
@@ -91,6 +91,38 @@ public function dokan_admin_toolbar() {
...
+     * Get admin menus data for dokan.
+     *
+     * @since 3.2.13
+     *
+     * @return array
+     */
+    public function get_dokan_admin_menus() {
+        $menus = [
+            [
+                'parent' => 'site-name',
```
xaviranik, line 104: Fix alignment here.

includes/Admin/AdminBar.php
```diff
@@ -108,19 +140,22 @@ public function visit_dashboard_menu( $wp_admin_bar ) {
...
-                'parent' => 'site-name',
-                'id'     => 'view-dashboard',
-                'title'  => __( 'Visit Vendor Dashboard', 'dokan-lite' ),
-                'href'   => $dashboard,
-            ]
-        );
+        $menus = $this->get_dokan_admin_menus();
+
+        // Show wp admin menus in admin bar here, if menu URL.
+        foreach ( $menus as $menu ) :
```
xaviranik, line 146: Might consider not using conditional tags: https://developer.wordpress.org/themes/basics/conditional-tags/

includes/Admin/AdminBar.php
```diff
@@ -108,19 +140,22 @@ public function visit_dashboard_menu( $wp_admin_bar ) {
...
+            if ( '' === $menu['href'] ) :
```
xaviranik, line 147: Consider checking `isset()` here

xaviranik (changes requested):

includes/Admin/AdminBar.php
```diff
@@ -108,19 +140,22 @@ public function visit_dashboard_menu( $wp_admin_bar ) {
...
 
-        $dashboard = dokan_get_navigation_url();
-
-        if ( ! $dashboard ) {
-            return;
+        $menus = $this->get_dokan_admin_menus();
+
+        // Show wp admin menus in admin bar here, if menu URL.
+        foreach ( $menus as $menu ) {
+            if ( ! isset( $menu['href'] ) || empty( $menu['href'] ) ) {
```
ManiruzzamanAkash, line 147: Only `empty()` check is enough here. It will do the both.
And what if one `$menu['href']` is empty, but other `$menu` has.

So, best-case would be - 
```php
foreach ( $menus as $menu ) {
    if ( ! empty( $menu['href'] ) ) {
        $wp_admin_bar->add_node(
              [
                   'parent' => $menu['parent'],
                   'id'     => $menu['id'],
                   'title'  => $menu['title'],
                   'href'   => $menu['href'],
              ]
         );
   }
}
```

includes/Admin/AdminBar.php
```diff
@@ -91,6 +91,38 @@ public function dokan_admin_toolbar() {
         do_action( 'dokan_render_admin_toolbar', $wp_admin_bar );
     }
 
+    /**
+     * Get admin menus data for dokan.
+     *
+     * @since DOKAN_LITE_SINCE
```
xaviranik, line 97: DOKAN_LITE_SINCE won't work here. Kindly replace with the upcoming release version.

PR #1349, opened 2021-09-27
Fix: Vendor Contact form doesn't contain customer email

### LINKED ISSUE #1346: Vendor Contact form email doesn't contain any customer email
Describe the bug:

When a customer tries to use the Vendor Contact form widget and sends an email directly from the Single Store page, the vendor gets the email, but when the vendor replies to that email, there is no customer email on the to send the reply. 

The customer email should be there at CC or Send To filed of the email. Then the customer will get a reply.

How To Reproduce:

1. Send an email via the Vendor Contact form of any store as a customer
2. Then try to reply to that email from that vendor's email account
3. There won't be a customer email to reply to the customer

Expected behavior:

The customer email should be on the Send To or CC field as soon as the vendor tries to reply to the customer email.

Screenshots:

Here's a [screenshot](https://imgur.com/a/a9XP9T0) to demonstrate the issue

shahadat015:
When a customer contacts a vendor via the vendor contact form widget. It does send emails to the vendor. but when the vendor tries to reply to the customer he/she can't because there is no customer email in To or email cc. Email send to the vendor self instead of the customer. 

Fixes #1346

### DIFF
--- a/includes/Emails/ContactSeller.php
+++ b/includes/Emails/ContactSeller.php
@@ -14,6 +14,8 @@ use WC_Email;
  */
 class ContactSeller extends WC_Email {
 
+    private $reply_to;
+
     /**
      * Constructor.
      */
@@ -27,6 +29,7 @@ class ContactSeller extends WC_Email {
 
         // Triggers for this email
         add_action( 'dokan_trigger_contact_seller_mail', array( $this, 'trigger' ), 30, 4 );
+        add_filter( 'woocommerce_email_from_address', array( $this, 'set_reply_to' ) );
 
         // Call parent constructor
         parent::__construct();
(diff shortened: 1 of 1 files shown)


includes/Emails/ContactSeller.php
```diff
@@ -52,56 +62,68 @@ public function get_default_subject() {
...
+        $this->send( $seller_email, $this->get_subject(), $this->get_content(), $this->get_headers(), $this->get_attachments() );
+        $this->restore_locale();
+    }
+
+    /**
+     * Set customer email address for reply
+     *
+     * @since DOKAN_LITE_SINCE
+     *
+     * @return mixed
```
ManiruzzamanAkash, line 103: This will return must be a `string`, right ??
We should avoid `mixed` wherever possible.

PR #1195, opened 2021-05-07
Extend the function `dokan_get_page_url` to accept `$subpage` as an extra parameter

strategio:
Transposition of initial PR on our fork (https://github.com/OnTheGoSystems/dokan/pull/2).

- Introduced a new function `dokan_add_subpage_to_url` to avoid the
repetitive issues due to incorrect path concatenation.
- Added- the new subpage argument in the filter hook.

https://github.com/weDevsOfficial/dokan-wpml/issues/13

PR #1360, opened 2021-10-04
Extend the function `dokan_get_page_url` to accept `$subpage` as an (…

nurul-umbhiya:
…#1195)

extra parameter

- Introduced a new function `dokan_add_subpage_to_url` to avoid the
repetitive issues due to incorrect path concatenation.
- Added- the new subpage argument in the filter hook.

https://github.com/weDevsOfficial/dokan-wpml/issues/13


includes/functions.php
```diff
@@ -1040,17 +1040,36 @@ function dokan_locate_template( $template_name, $template_path = '', $default_pa
...
+    $url = get_permalink( $page_id );
+
+    if ( $subpage ) {
+        $url = dokan_add_subpage_to_url( $url, $subpage );
+    }
+
+    return apply_filters( 'dokan_get_page_url', $url, $page_id, $context, $subpage );
+}
+
+/**
```
alamgircsebd, line 1063: Please add a short description and since

alamgircsebd (changes requested):

PR #1379, opened 2021-10-18
Enha: New action hook dokan_after_product_listing_status_filter added

shahadat015:
For showing products counts for each language on the vendor dashboard inside the products screen. I have added a new action hook called `dokan_after_product_listing_status_filter` on file `dokan-lite/templates/products/listing-status-filter.php`


shahadat015:
Don't need to show count now

PR #1391, opened 2021-11-01
Fix/update refund and withdraw database table column

rajankarmakar:
There was a bug when a user missed to run the upgrader after a couple of releases if the user runs the data upgrader the database does not upgrade as expected.

fix https://github.com/weDevsOfficial/dokan-pro/pull/1558

### DIFF
--- a/includes/Upgrade/Upgrades/V_3_3_1.php
+++ b/includes/Upgrade/Upgrades/V_3_3_1.php
@@ -44,13 +44,7 @@ class V_3_3_1 extends DokanUpgrader {
         $existing_columns = $wpdb->get_results( "DESC `{$wpdb->prefix}dokan_refund`" );
 
         foreach ( (array) $existing_columns as $existing_column ) {
-            if (
-                'text' !== $existing_column->Type && // phpcs:ignore
-                (
-                    'item_totals' === $existing_column->Field ||  // phpcs:ignore
-                    'item_tax_totals' === $existing_column->Field  // phpcs:ignore
-                )
-            ) {
+            if ( in_array( $existing_column->Field, [ 'item_totals', 'item_tax_totals' ], true ) && 'text' !== $existing_column->Type ) { // phpcs:ignore
                 $wpdb->query(
                     "ALTER TABLE `{$wpdb->prefix}dokan_refund` MODIFY COLUMN {$existing_column->Field} text" // phpcs:ignore
                 );
--- a/includes/Upgrade/Upgrades/V_3_3_1.php
+++ b/includes/Upgrade/Upgrades/V_3_3_1.php
@@ -7,13 +7,13 @@ use WeDevs\Dokan\Abstracts\DokanUpgrader;
 class V_3_3_1 extends DokanUpgrader {
 
     /**
-     * Updates withdraw database table
+     * Updates withdraw database table column
      *
      * @since 3.3.7
      *
      * @return void
      */
-    public static function update_withdraw_table() {
+    public static function update_withdraw_table_column() {
         global $wpdb;
 
         include_once ABSPATH . 'wp-admin/includes/upgrade.php';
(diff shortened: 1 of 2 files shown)


includes/Upgrade/Upgrades/V_3_3_1.php
```diff
@@ -0,0 +1,56 @@
...
+namespace WeDevs\Dokan\Upgrade\Upgrades;
+
+use WeDevs\Dokan\Abstracts\DokanUpgrader;
+
+class V_3_3_1 extends DokanUpgrader {
+
+    /**
+     * Updates withdraw database table column
+     *
+     * @since DOKAN_LITE_SINCE
```
devAsadNur, line 12: Instead of `DOKAN_LITE_SINCE`, please use possible upcoming version number.

includes/Upgrade/Upgrades/V_3_3_1.php
```diff
@@ -0,0 +1,56 @@
...
+            return;
+        }
+
+        $wpdb->query(
+            "ALTER TABLE `{$wpdb->prefix}dokan_withdraw` ADD COLUMN `details` longtext AFTER `note`" // phpcs:ignore
+        );
+    }
+
+    /**
+     * Updates withdraw database table column
```
devAsadNur, line 33: Here, it should be `refund database` instead of `withdraw database`. Please fix it.

devAsadNur (changes requested):

includes/Upgrade/Upgrades/V_3_3_1.php
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @since DOKAN_LITE_SINCE
```
devAsadNur, line 35: Same as above, instead of `DOKAN_LITE_SINCE`, please consider using possible upcoming version number.

includes/Upgrade/Upgrades/V_3_3_1.php
```diff
@@ -0,0 +1,56 @@
...
+     * @return void
+     */
+    public static function update_refund_table_column() {
+        global $wpdb;
+
+        include_once ABSPATH . 'wp-admin/includes/upgrade.php';
+
+        $existing_columns = $wpdb->get_results( "DESC `{$wpdb->prefix}dokan_refund`" );
+
+        if ( ! empty( $existing_columns ) ) {
```
tanmayjay, line 46: You don't need to check if it is empty. Just cast the `$existing_columns` by `array` while running foreach loop. As the `$wpdb::get_results` method return `array|object|null`, if the value is null, the array casting will convert it into an empty array and for empty array the foreach loop will not run automatically.
Rewrite this in the following way:

```
    foreach ( (array) $existing_columns as $existing_column ) {
        if (
            'text' !== $existing_column->Type &&
            (
                'item_totals' === $existing_column->Field ||
                'item_tax_totals' === $existing_column->Field
            )
        ) {
            $wpdb->query(
                "ALTER TABLE `{$wpdb->prefix}dokan_refund` MODIFY COLUMN {$existing_column->Field} text" // phpcs:ignore
            );
        }
    }
```

tanmayjay (changes requested):

tanmayjay (approved):

devAsadNur (approved):

includes/Upgrade/Upgrades/V_3_3_1.php
```diff
@@ -0,0 +1,60 @@
...
+    public static function update_refund_table_column() {
+        global $wpdb;
+
+        include_once ABSPATH . 'wp-admin/includes/upgrade.php';
+
+        $existing_columns = $wpdb->get_results( "DESC `{$wpdb->prefix}dokan_refund`" );
+
+        foreach ( (array) $existing_columns as $existing_column ) {
+            if (
+                'text' !== $existing_column->Type && // phpcs:ignore
```
nurul-umbhiya, line 48: rearrange this condition like below:
```
if ( in_array( $existing_column->Field, [ 'item_totals', 'item_tax_totals' ], true ) && 'text' !== $existing_column->Type )

```

nurul-umbhiya (approved):

nurul-umbhiya (approved):

PR #1392, opened 2021-11-02
Fix/render vendor email setting conditionally

rajankarmakar:
There is a setting in Customizer "Hide email address". If the admin checks this setting checkbox, the vendor should not see the "Show email address in store" option on the vendor setting page.


rajankarmakar:
Unexpected PR. Accidently unwanted PR has come along with this PR. So I'm closing the PR.

Issue #1398, opened 2021-11-04
Allow hiding product categories from Dokan dashboard


davefx:
**Is your feature request related to a problem? Please describe.**
We are introducing Dokan into an existing shop. Here we have some _internal_ categories that we don't want to show to Vendors. We need a filter that allows us to hide categories from the Dokan dashboard so they are not visible to them. 

**Describe the solution you'd like**
It would be great if we have a hook that allows us to customize the product categories that are shown in the dashboard.

**Describe alternatives you've considered**
We know that we could get this by customizing the existing templates, but we consider this change to be interesting as a new feature in Dokan.

nurul-umbhiya:
I'm going to close this pr, is because there is already a hook defined: dokan_product_cat_dropdown_args
use this filter hook and add exclude param,

davefx:
The hook is defined, but it's not being called when generating the category dropdown in the modified interfaces!

nurul-umbhiya:
We are already using this hook to include specific categories from the Vendor Subscription module, this should work just fine. If this doesn't work, kindly share your code with us, will try to figure this out. thanks

davefx:
Ok. I found the actual origin of my found error: dokan_product_cat_dropdown_args was not being called from listing-filter.php. I'm now creating the correct Pull Request with the correct changes.

PR #1399, opened 2021-11-04
feat(Product): add new filter to hide categories

davefx:
New filter to allow third parties to hide product categories from Dokan dashboard

Fixes: #1398

### DIFF
--- a/templates/products/listing-filter.php
+++ b/templates/products/listing-filter.php
@@ -21,6 +21,12 @@ $product_type  = isset( $get_data['product_type'] ) ? $get_data['product_type'] 
 
         <div class="dokan-form-group">
             <?php
+
+                /**
+                 *  dokan_excluded_product_categories hook
+                 *
+                 *  @since 3.2.16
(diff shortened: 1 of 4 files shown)


nurul-umbhiya:
I'm going to close this pr, is because there is already a hook defined: dokan_product_cat_dropdown_args
use this filter hook and add exclude param,
```
add_filter( 'dokan_product_cat_dropdown_args', 'test_filter_category' ,10, 1 );
function test_filter_category( $category_args ) {
    if ( ! isset( $category_args['exclude'] ) {
        $category_args['exclude'] = [];
    }
    $category_args['exclude'] = array_merge( $category_args['exclude'], [1, 2, 3] ); // here 1,2,3 is my args
}
```

davefx:
It would be great if `dokan_product_cat_dropdown_args` were called, but the fact is that it is not!

I already tried to use it, but that filter is not being called at all when generating the dropdown filter in the product view.

Issue #1400, opened 2021-11-05
dokan_get_vendor_by_product is not able to avoid the addition of Product Seller info to the cart


davefx:
**Describe the bug**
All products in the cart will get a "Vendor: " tag, no matter if a vendor is actually found or not.

**To Reproduce**
After using the 'dokan_get_vendor_by_product' hook to try to avoid the addition of a "Vendor" tag into a cart item, Dokan keeps adding "Vendor" tag to every item in the cart.

**Expected behavior**
If the hook call "dokan_get_vendor_by_product" returns null or false, I would expect "dokan_product_seller_info" wouldn't add any Dokan-tag to the cart item. 

**Screenshots**

![image](https://user-images.githubusercontent.com/292309/140578378-beeed78a-4d76-43fd-8bdf-e4869ec9bb73.png)

**Environment (please complete the following information):**
 - Dokan Version: v3.3.0
 - Dokan Pro Version: v3.4.0
 - WooCommerce Version: 5.8.0
 - Browser: Firefox
 - Current Theme: Own

PR #1401, opened 2021-11-05
Fix[Product]: allow filter to prevent adding cart-item tag

davefx:
Allow filter dokan_get_vendor_by_product preventing the addition of the "Vendor" cart-item tag if it doesn't return a valid vendor

Fixes: #1400

### DIFF
--- a/includes/Product/functions.php
+++ b/includes/Product/functions.php
@@ -543,10 +543,6 @@ function dokan_get_vendor_by_product( $product, $id = false ) {
 
     $vendor_id = apply_filters( 'dokan_get_vendor_by_product', $vendor_id, $product );
 
-    if ( ! $vendor_id ) {
-        return false;
-    }
-
     return false === $id ? dokan()->vendor->get( $vendor_id ) : (int) $vendor_id;
 }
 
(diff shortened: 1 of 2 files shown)


includes/Product/functions.php
```diff
@@ -543,6 +543,10 @@ function dokan_get_vendor_by_product( $product, $id = false ) {
 
     $vendor_id = apply_filters( 'dokan_get_vendor_by_product', $vendor_id, $product );
 
+    if ( ! $vendor_id ) {
+        return false;
+    }
+
     return false === $id ? dokan()->vendor->get( $vendor_id ) : (int) $vendor_id;
```
nurul-umbhiya, line 550: instead of returning false, what you can do is on line 550
```
return false === $id  && ! empty( $vendor_id ) ? dokan()->vendor->get( $vendor_id ) : (int) $vendor_id;
```

nurul-umbhiya: also you can change the parameter $id  with  `$get_id` and change accordingly

nurul-umbhiya (changes requested):

davefx:
Ok. I understand what you mean.

I just moved my changes to wc-template.php.

nurul-umbhiya (approved):

PR #1417, opened 2021-11-15
fix: Withdraw details field value conflicting with old version

alamgircsebd:


nurul-umbhiya (approved):

FaraziF:
This will be valid if the vendor's payment does not have any credentials, then the withdraw table details column will be blank.

PR #1420, opened 2021-11-16
Fix[Order]: Determine the vendor a product belongs to correctly

### LINKED ISSUE #1419: Filter dokan_get_vendor_by_product ignored when determining Coupon validity
**Describe the bug**
When determining if a coupon can be applied or not, the filter dokan_get_vendor_by_product is being ignored by the ensure_vendor_coupon() function in class WeDevs\Dokan\Order\Hooks , determining the vendor-id corresponding to a product just by checking the post_author field of the product post, instead of calling the function dokan_get_vendor_by_product(). 

**To Reproduce**
1. Configure a function to customize the vendor a product belongs to using the filter hook "dokan_get_vendor_by_product". 
2. Then configure a coupon valid for that vendor. 
3. See error: the coupon won't ever be valid for products that are forcefully assigned to the vendor through the filter hook.

**Expected behavior**
The function dokan_get_vendor_by_product() should be called to determine the vendor a product belongs to, so the filter dokan_get_vendor_by_product is executed.

**Environment (please complete the following information):**
 - Dokan Version:  v3.3.1
 - Dokan Pro Version: v3.4.1
 - WooCommerce Version: v5.7.1

davefx:
Correctly determine the vendor a product belongs to, so the "dokan_get_vendor_by_product" filter hook is called.

Fixes #1419


includes/Order/Hooks.php
```diff
@@ -292,7 +292,9 @@ public function ensure_vendor_coupon( $valid, $coupon ) {
         foreach ( WC()->cart->get_cart() as $item ) {
             $product_id = $item['data']->get_id();
 
-            $available_vendors[]  = (int) get_post_field( 'post_author', $product_id );
+            $vendor = dokan_get_vendor_by_product( $product_id );
```
alamgircsebd, line 295: Pass 2nd parameter `true` then the function will be returned directly `vendor_id`
```
$vendor_id                  = dokan_get_vendor_by_product( $product_id, true );
$available_vendors[]  = $vendor_id ? $vendor_id : 0;
```

nurul-umbhiya, line 295: @davefx 
`$available_vendors[]  = dokan_get_vendor_by_product( $product_id, true );` is enough

davefx:
Suggestion accepted. Now using second parameter of dokan_get_vendor_by_product()

nurul-umbhiya (approved):

PR #1478, opened 2022-01-03
enhance/added-vendor-residence-proof-for-verify

MdAsifHossainNadim:
### Update vendor verification & added vendor residence proof.

Previous this enhance there is no "proof of residence" document to be uploaded to the seller verification module, so it is not possible for a seller to prove that the address he is registering with is a valid address. And it is mandatory for a marketplace to ensure that vendors provide the correct address for legal purposes.

Now, that's fine. Sellers can add their "proof of residence" from the address verification area. There they can find an upload proof button to upload their proof. And after submission the admin can get a url under verification address for check seller proof.

**Video:**
https://share.vidyard.com/watch/Vm4ZJtCT1VeecpqLhuKZEU?

fix: https://github.com/weDevsOfficial/dokan/issues/1404
Related with Dokan Pro: https://github.com/weDevsOfficial/dokan-pro/pull/1675

### DIFF
--- a/templates/settings/address-form.php
+++ b/templates/settings/address-form.php
@@ -141,13 +141,13 @@ $address_state   = isset( $profile_info['address']['state'] ) ? $profile_info['a
                 </select>
             <?php } ?>
             </div>
+        <?php } ?>
+
         <?php
-        }
-        
         /**
          * Add vendor address verification templates.
          * 
-         * @since 3.3.6
+         * @since DOKAN_SINCE
          * 
          * @param array $address               Vendor address info.
          * @param array $seller_address_fields Vendor required addresses.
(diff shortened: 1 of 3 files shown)


templates/settings/address-form.php
```diff
@@ -141,6 +141,19 @@
...
             <?php } ?>
             </div>
-        <?php } ?>
+        <?php
+        }
+        
+        /**
+         * Add vendor address verification templates.
+         * 
+         * @since DOKAN_PRO_SINCE
```
rajankarmakar, line 150: As this hook is defined on Dokan lite the since version should be the Dokan lite since. Kindly fix this `@since DOKAN_PRO_SINCE`

rajankarmakar (changes requested):

tanmayjay (approved):

PR #1486, opened 2022-01-11
Ensure correct order status is displayed for vendors after updating

rmilesson:
This is a very small PR to ensure that the correct order status will be visible after updating the order status.

In some cases plugin/theme authors will hook into actions like `woocommerce_order_status_x_to_x` and similar to update the order status again depending on certain store needs. In these cases Dokan would show the wrong order status.

With this PR we ensure that we return the order status assigned to the order after all hooks has run.

### DIFF
--- a/includes/Ajax.php
+++ b/includes/Ajax.php
@@ -302,6 +302,12 @@ class Ajax {
         $order = dokan()->order->get( $order_id );
         $order->update_status( $order_status );
 
+        // Get the new order status. This is needed since plugin/theme authors might
+        // change the order status behind the scenes in certain cases.
+        // For example by moving `wc-paused` to `wc-cancelled` automatically or by
+        // moving `wc-pending` to `wc-processing`.
+        $order_status = "wc-{$order->get_status()}";
+
         $statuses     = wc_get_order_statuses();
         $status_label = isset( $statuses[ $order_status ] ) ? $statuses[ $order_status ] : $order_status;
         $status_class = dokan_get_order_status_class( $order_status );


rmilesson:
Can this PR get some attention please? There should be absolutely no issues to get this merged.

nurul-umbhiya:
@rmilesson We will merge this in next release insha'Allah, kindly wat for few more days. Possible release date is 25th of Jan

includes/Ajax.php
```diff
@@ -302,6 +302,12 @@ public function change_order_status() {
         $order = dokan()->order->get( $order_id );
         $order->update_status( $order_status );
 
+        // Get the new order status. This is needed since plugin/theme authors might
```
nurul-umbhiya, line 305: @rmilesson why we need to get order status manually? there is a hook to add order status, maybe you could use that?
`apply_filters( 'wc_order_statuses', $order_statuses )`

rmilesson: This is not possible in this case since `$order_status` (fetched on line 313) can be out of date after running `$order->update_status(  $order_status ) ` on line 320.
This is why we need to re-fetch `$order_status`.

rmilesson: Consider this hook in a theme:

```
add_action(
	'woocommerce_order_status_changed',
	function ( $order_id ) {
		$order = wc_get_order( $order_id );
		
		$order->update_status( 'wc-processing' );
		$order->save();
	}
);
```

This hook would demonstrate the issue by always setting the order status to `wc-processing`. However you would show the wrong order status after updating the order status on the order dashboard.

nurul-umbhiya (approved):

PR #1491, opened 2022-01-14
fix/product status becomes out of stock on variation product

rajankarmakar:
When setting bulk regular prices from the vendor dashboard in a variable product the product stock status becomes out of stock. This issue has been fixed with this PR.

### DIFF
--- a/includes/functions-compatibility.php
+++ b/includes/functions-compatibility.php
@@ -129,7 +129,12 @@ function dokan_get_product_downloads( $product ) {
  * @param string $date_to
  */
 function dokan_save_product_price( $product_id, $regular_price, $sale_price = '', $date_from = '', $date_to = '' ) {
-    $product       = wc_get_product( absint( $product_id ) );
+    $product = wc_get_product( absint( $product_id ) );
+
+    if ( ! $product instanceof WC_Product ) {
+        return;
+    }
+
     $regular_price = wc_format_decimal( $regular_price );
     $sale_price    = '' === $sale_price ? '' : wc_format_decimal( $sale_price );
     $date_from     = wc_clean( $date_from );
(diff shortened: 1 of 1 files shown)


includes/functions-compatibility.php
```diff
@@ -129,40 +129,45 @@ function dokan_get_product_downloads( $product ) {
  * @param string $date_to
  */
 function dokan_save_product_price( $product_id, $regular_price, $sale_price = '', $date_from = '', $date_to = '' ) {
-    $product_id    = absint( $product_id );
+    $product = wc_get_product( absint( $product_id ) );
+
```
devAsadNur, line 133: Consider removing this empty line and aligning the previous line (132) as per below lines ( from 134 to 137).

includes/functions-compatibility.php
```diff
@@ -129,40 +129,45 @@ function dokan_get_product_downloads( $product ) {
...
-        update_post_meta( $product_id, '_price', $sale_price );
+        $product->set_sale_price( $sale_price );
     } else {
-        update_post_meta( $product_id, '_price', $regular_price );
+        $product->set_sale_price( $regular_price );
     }
 
-    if ( '' !== $sale_price && $date_from && strtotime( $date_from ) < strtotime( 'NOW', current_time( 'timestamp' ) ) ) {
-        update_post_meta( $product_id, '_price', $sale_price );
+    if ( '' !== $sale_price && $date_from && strtotime( $date_from ) < strtotime( 'NOW', dokan_current_datetime()->getTimestamp() ) ) {
```
devAsadNur, line 158: Consider applying `empty()` check here instead of empty string check. So that, the checking will work for 0 or any other falsy values too.

devAsadNur (approved):

PR #1512, opened 2022-01-24
enhance/geocoding/reverse-geocoding-when-dragging-marker-in-map

### LINKED ISSUE #1509: Geolocation with Google Maps not working as expected (vendor settings and store listings)
**Describe the bug**

**To Reproduce**
1.)
1. Go to the vendor settings page
2. Click on the map in a different town
3. See that the input field is not updated.

2.)
1. Go to store listing page
2. See that stores aren't being shown in the map

**Expected behavior**
1.) Selecting a town on the map should update the location in the input field
2.) There should be markers in the store-listing page's map for each store

**Screenshots**
Screenshot_20220120_181037
Screenshot_20220120_180956

**Environment (please complete the following information):**
 - Dokan Version: 3.3.6
 - Dokan Pro Version: 3.4.4
 - WooCommerce Version: 6.1.0
 - Browser Chrome, Edge, Firefox
 - Current Theme: Dokani Develop

**Additional context**
At least for the first issue, if the gmap variable was global, I could fix it myself, but it's inside of a jQuery ready function.

NAbdulla1:
fixes #1509 

part of [Dokan Pro PR](https://github.com/weDevsOfficial/dokan-pro/pull/1716)

### DIFF
--- a/templates/maps/google-maps-with-search.php
+++ b/templates/maps/google-maps-with-search.php
@@ -141,7 +141,7 @@
                         .map( component => component.long_name )
                         .join( ', ' );
                     $input_add.val( address );
-                })
+                } )
                 .catch( e => console.log( e ) );
         }
 
--- a/templates/maps/google-maps-with-search.php
+++ b/templates/maps/google-maps-with-search.php
@@ -50,6 +50,11 @@
                 updatePositionInput(event.latLng );
             } );
 
+            window.google.maps.event.addListener( marker, 'drag', dokan_debounce_delay(
+                ( event ) => reverseGeocode( event.latLng ),
+                1000
+            ) );
+
         } catch( e ) {
             console.log( 'Google API not found.' );
         }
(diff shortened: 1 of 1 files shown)


templates/maps/google-maps-with-search.php
```diff
@@ -120,5 +125,39 @@ function autoCompleteAddress(){
                 }
             });
         }
+
+        function reverseGeocode( latLng ) {
+            let data = {
```
ManiruzzamanAkash, line 130: make this `data` variable data type from `let` to `const`

templates/maps/google-maps-with-search.php
```diff
@@ -120,5 +125,39 @@ function autoCompleteAddress(){
...
+                location: {
+                    lat: latLng.lat(),
+                    lng: latLng.lng(),
+                }
+            };
+
+            geocoder.geocode( data )
+                .then( function ( { results: addresses } ) {
+                    let address = addresses[0].address_components
```
ManiruzzamanAkash, line 139: same here - 
make this `address` variables data type from `let` to `const`

ManiruzzamanAkash, line 139: Isn't that necessary to check `addresses` length > 0 or not.
as you're using `addresses[0]`, hope, it would be better to check if `addresses.length > 0` then do rest.

templates/maps/google-maps-with-search.php
```diff
@@ -120,5 +125,39 @@ function autoCompleteAddress(){
...
+                        .filter( component => ! component.types.includes( 'plus_code' ) )
+                        .map( component => component.long_name )
+                        .join( ', ' );
+                    $input_add.val( address );
+                } )
+                .catch( e => console.log( e ) );
+        }
+
+        function dokan_debounce_delay( callback, ms ) {
+            var timer = 0;
```
ManiruzzamanAkash, line 149: you've use `let` above, but `var` here? why? You should keep the code consistency.
same for bottom `context` and `args` variable.

ManiruzzamanAkash (approved):

shashwatahalder01:
Scenario: 
Vendor settings: The location input field is only updated by dragging the location marker. The location input field should also be updated by clicking on any point on the map. 

For reference, Go to vendor product edit page > Geolocation see location input field update behavior.
https://www.awesomescreenshot.com/video/7076170?key=3c0c5d9bd925269d61efee1c7c285bf6

PR #1539, opened 2022-02-25
enhance/made_dokan_vendor_profile_progress_bar_closable

NAbdulla1:
part of [this](https://github.com/weDevsOfficial/dokan-pro/pull/1726)


assets/js/dokan.js
```diff
@@ -2250,6 +2250,10 @@ jQuery(function($) {
           );
 
           $('.dokan-ajax-response').append(resp.data.progress);
+
```
nurul-umbhiya, line 2253: remove this code from here, this file will generate automatically via grunt command

PR #1541, opened 2022-02-28
fix/used-for-variation-is-shown-for-new-attribute-even-if-product-type-is-not-variable

NAbdulla1:
'User for Variation' checkbox is hidden when new attribute is loaded and product type is not variable type

we can't use the change trigger defined on line 511 because when there is only one product type then this trigger has no effect.

Fixes [this dokan pro issue](https://github.com/weDevsOfficial/dokan-pro/issues/1755)

### DIFF
--- a/assets/src/js/product-editor.js
+++ b/assets/src/js/product-editor.js
@@ -517,7 +517,9 @@
                             let labels = $( 'div.dokan-product-attribute-wrapper label.show_if_variable' );
 
                             for( let label of labels ) {
-                                if ( label.textContent.includes( "Used for variations" ) ) {
+                                let checkBox = $( label ).find('input[type="checkbox"]');
+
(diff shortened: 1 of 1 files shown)


assets/src/js/product-editor.js
```diff
@@ -512,6 +512,16 @@
                         Dokan_Editor.loadSelect2();
                         Dokan_Editor.bindProductTagDropdown();
                         Dokan_Editor.attribute.reArrangeAttribute();
+
+                        if ( "variable" !== $( 'select#product_type' ).val() ) {
+                            let labels = $( 'div.dokan-product-attribute-wrapper label.show_if_variable' );
+
+                            for( let label of labels ) {
+                                if ( label.textContent.includes( "Used for variations" ) ) {
```
tanmayjay, line 520: Isn't there any obvious way to check this? What if the text content is translated. Kindly make it sure.

tanmayjay (changes requested):

tanmayjay (approved):

PR #1550, opened 2022-03-06
enhance/store-times-hover-feature-for-elementor

MdAsifHossainNadim:
### Update elementor store times widget & add hover feature.

Previous this enhancement, In the vendor store page has no reflection of the Day & time when page build via elementor or dokan elementor blocks, only "Store is open/closed" message is shown.

Now, after this enhancement vendor can show his stores times when build store page via elementor.

Related with dokan-pro: https://github.com/weDevsOfficial/dokan-pro/pull/1794


templates/store-header-times.php
```diff
@@ -0,0 +1,51 @@
...
+                    <span class="store-close"><?php echo esc_html( $closed_status ); ?></span>
+                <?php endif; ?>
+
+                <?php
+                // Get store times.
+                $opening_times = ! empty( $store_info['opening_time'] ) ? $store_info['opening_time'] : [];
+
+                // If dokan pro doesn't exists then get single item.
+                if ( ! dokan()->is_pro_exists() ) {
+                    $opening_times = ! empty( $opening_times ) && is_array( $opening_times ) ? $opening_times[0] : [];
```
tanmayjay, line 24: As it seems, the `$opening_times` here is holding data like `[ 'time1', 'time2',.... ]`. What you intend to do is taking only one time from here but ultimately process further considering it as an array. In that case, it's better to make an array with the first element here. for example,
 `$opening_times = ! empty( $opening_times ) && is_array( $opening_times ) ? [ $opening_times[0] ] : [];`

tanmayjay: Even though you are casting it in array later, it's safer to do it this way here.

tanmayjay (changes requested):

tanmayjay (approved):

PR #1555, opened 2022-03-07
fix/stackoverflow-in-erp-pages-because-of-dokan-lite

NAbdulla1:
[**fix**] fixed this error:
image


nurul-umbhiya:
@NAbdulla1 can you please test this issue with current development version? I think this pr is not needed anymore.

NAbdulla1:
OK vai, let me check.

PR #1567, opened 2022-03-09
Fix/solved vendor store settings validation error

MdAsifHossainNadim:
### Solved vendor store settings field validation error.

Previous this fix, find an error in the console when we type in the phone number field in the seller store settings.
Now, It's fine.

fix: https://github.com/weDevsOfficial/dokan-pro/issues/1811

### DIFF
--- a/templates/settings/store-form.php
+++ b/templates/settings/store-form.php
@@ -486,7 +486,13 @@
             $('#setting_phone').on( 'keydown', function(e) {
                 let cKey     = 67,
                     vKey     = 86,
+                    cmdKey   = 91,
+                    ctrlKey  = 17,
                     ctrlDown = false;
+
+                if (e.keyCode === ctrlKey || e.keyCode === cmdKey) {
+                    ctrlDown = true;
+                }
 
                 // Allow: backspace, delete, tab, escape, enter and .
                 if ($.inArray(e.keyCode, [46, 8, 9, 27, 13, 91, 107, 109, 110, 187, 189, 190]) !== -1 ||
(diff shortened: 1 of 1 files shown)


templates/settings/store-form.php
```diff
@@ -484,17 +484,29 @@
...
                     (e.keyCode == 65 && e.ctrlKey === true) ||
                     //Allow Ctrl+v
                     (e.keyCode === vKey && ctrlDown) ||
                     // Allow: Ctrl+c
                     (e.keyCode === cKey && ctrlDown) ||
-                     // Allow: home, end, left, right
-                    (e.keyCode >= 35 && e.keyCode <= 39)) {
-                         // let it happen, don't do anything
+                    // Allow: home, end, left, right.
+                    (e.keyCode >= 35 && e.keyCode <= 39)
```
tanmayjay, line 507: Make sure the key range here is correct.

templates/settings/store-form.php
```diff
@@ -484,17 +484,29 @@
...
+
+                if (e.keyCode === ctrlKey || e.keyCode === cmdKey) {
+                    ctrlDown = true;
+                }
+
+                if (
+                    // Allow: backspace, delete, tab, escape, enter etc.
+                    $.inArray(e.keyCode, [46, 48, 57, 8, 9, 27, 13, 91, 107, 109, 110, 187, 189, 190]) !== -1 ||
+                    // Allow: Ctrl+A
                     (e.keyCode == 65 && e.ctrlKey === true) ||
```
tanmayjay, line 501: This keycode also could be stored in a variable to understand which key it is.

tanmayjay (changes requested):

shashwatahalder01:
@MdAsifHossainNadim vai, this fix doesn't work for the AZERTY keyboard layout.

PR #1571, opened 2022-03-14
CRITICAL Uncaught Error store-open-close.php

kas5986:
CRITICAL Uncaught Error: Call to a member function getTimestamp() on bool in wp-content/plugins/dokan-lite/templates/widgets/store-open-close.php:47

above error fix when there is an empty string


nurul-umbhiya:
Already fixed

PR #1611, opened 2022-05-09
Feat/reverse withdrawal

### LINKED ISSUE #1445: Withdrawal reverse
If customer opts to pay on delivery (COD) then the admin fee is missed from equation.. there needs to be a way to flag to admin that this cost needs to be claimed.
This is available on WCFM as below
https://wclovers.com/knowledgebase/wcfm-marketplace-reverse-withdrawal/

I don't want a subscription model. I want to be able to flag to a vendor that they owe an amount for the sale and it's either taken from their withdrawal pot or they can pay it off.

nurul-umbhiya:
### All Submissions:

* [x] My code is tested

1. Implemented Reverse Withdrawal Feature. Feature Description Video Link: [Dokan Reverse Withdrawal](https://www.youtube.com/watch?v=9m0ERBQjYmY)

Related Dokan Pro PR: https://github.com/weDevsOfficial/dokan-pro/pull/1868

Closes #1445

1. npm run dev:build
3. grunt
4. composer dumpautoload -o

### Changelog entry

> - **feat:** Added reverse withdrawal feature.

### FOR PR REVIEWER ONLY:


Rat01047:
Issue-1: 
Upgrader was not run automatically, hence getting an error.

Scenario: Fresh Setup > Upgrader was not run > Error recevied.

Rat01047:
Issue-2: 
Withdraw option is missing from Vendor Dashboard Menu

Scenario: Fresh Installation > When "Reverse Withdraw" is enabled > Withdraw option goes missing from Vendor Dashboard Menu

Screenshot given for reference:
image

Rat01047:
Issue-3: 
"Cash on delivery" unchecked in Settings - Reverse Withdrawal > but still Cash on delivery transaction is being counted in Vendor - Dashboard - Reverse Withdrawal Menu Option

Scenario: Uncheck Reverse Withdrawal in Dokan - Settings > Make Order from Customer > Check Vendor - Dashboard - Reverse Withdrawal Menu Option > Record of transaction is taken > This should not occur, since the option was not enabled from Admin

Screenshot given for reference: 

image

image

Rat01047:
Issue-4: 
Email Body > Showing Error in Email body

Scenario: If Email is sent before the assigned date > the Email received has error in the email body.

Screenshot given for reference:
image

Rat01047:
Issue-5: 
Even thought Grace Period is given > Action is being taken immediately

Scenario: Set Billing Date + Grace Period > Email should have invoice with correct date + Vendor Dashboard should have correct date > But it is showing that immediate action has been taken.

Screenshot given for reference:
Screenshot 2022-05-25 at 12 22 07 PM

Screenshot 2022-05-25 at 12 22 18 PM

Rat01047:
Issue-6: 
Icon not loading

Scenario: Dokan > Reverse Withdrawal > Icons not loading

Rat01047:
Issue-6:
Icon not loading

Scenario: Dokan > Reverse Withdrawal > Icons not loading

includes/ReverseWithdrawal/Manager.php
```diff
@@ -0,0 +1,782 @@
...
+        $cache_key   = 'get_transactions_' . md5( wp_json_encode( $args ) );
+        if ( is_numeric( $args['vendor_id'] ) ) {
+            $cache_group = "reverse_withdrawal_{$args['vendor_id']}";
+        } elseif ( ! $this->is_empty( $args['vendor_id'] ) && 1 === count( $args['vendor_id'] ) ) {
+            $cache_group = "reverse_withdrawal_{$args['vendor_id'][0]}";
+        }
+
+        $data = Cache::get( $cache_key, $cache_group );
+
+        if ( in_array( $args['return'], [ 'vendor_transaction' ], true ) && false === $data ) {
```
tanmayjay, line 216: Is `in_array` needed here? `===` would do.

templates/emails/plain/reverse-withdrawal-invoice.php
```diff
@@ -0,0 +1,59 @@
...
+
+printf( esc_attr__( 'Hi %s,\n\n', 'dokan-lite' ), esc_attr( $seller_info->get_shop_name() ) );
+
+printf( esc_attr__( 'Your %1$s %2$s invoice is now available for store: %3$s.\n\n', 'dokan-lite' ), $data['{month}'], $data['{year}'], $seller_info->get_shop_name() );
+
+printf(esc_attr__('Summary for %1$s : \n\n', 'dokan-lite'), $seller_info->get_shop_name());
+
+printf( esc_attr__( 'Reverse withdrawal charges for %1$s %2$s: %3$s \n\n', 'dokan-lite' ), $data['{month}'], $data['{year}'], wc_price( $due_status['balance']['payable_amount'] ) );
+
+printf( esc_attr__( 'Due Date %1$s : \n\n', 'dokan-lite' ), 'immediate' === $due_status['due_date'] ? ucfirst( $due_status['due_date'] ) : dokan_format_date( $due_status['due_date'] ) );
```
tanmayjay, line 35: `esc_html` should be used instead of all these `esc_attr`

templates/reverse-withdrawal/transaction-listing.php
```diff
@@ -0,0 +1,75 @@
...
+                    <?php
+                    // translators: 1) transaction url 2) transaction id
+                    echo sprintf( '<a href="%1$s" target="_blank">%2$s</a>', $transaction['trn_url'], $transaction['trn_id'] )
+                    ?>
+                </td>
+                <td><?php echo esc_html( $transaction['trn_date'] ); ?></td>
+                <td><?php echo esc_html( $transaction['trn_type'] ); ?></td>
+                <td><?php echo esc_html( $transaction['note'] ); ?></td>
+                <td><?php echo $transaction['debit'] === '' ? '--' : wc_price( $transaction['debit'] ); ?></td>
+                <td><?php echo $transaction['credit'] === '' ? '--' : wc_price( $transaction['credit'] ); ?></td>
```
tanmayjay, line 45: Kindly use relevant escaping function for every `echo`

templates/reverse-withdrawal/transaction-listing.php
```diff
@@ -0,0 +1,75 @@
...
+        }
+        if ( count( $items ) > 1 ) {
+            ?>
+            <tr>
+                <td></td>
+                <td></td>
+                <td></td>
+                <td></td>
+                <td></td>
+                <td><b><?php _e( 'Balance:', 'dokan-lite' ); ?></b></td>
```
tanmayjay, line 60: `_e` should not be used. `esc_html_e` is required.

tanmayjay (changes requested):

includes/Dashboard/Templates/ReverseWithdrawal.php
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /**
+     * Display notice on vendor dashboard page
+     *
+     * @since DOKAN_SINCE
+     *
+     * @return void
+     */
+    public function display_notice_on_vendor_dashboard() {
```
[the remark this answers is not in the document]
nurul-umbhiya: both  are coming from the different  hook, one will be always visible  and another one is based on condition. so both methods are necessary

includes/Install/Installer.php
```diff
@@ -375,22 +374,44 @@ public function create_vendor_balance_table() {
...
+    }
+
+    /**
+     * This method will create reverse withdrawal base product
+     *
+     * @since DOKAN_SINCE
+     *
+     * @return void
+     */
+    private function create_reverse_withdrawal_base_product() {
```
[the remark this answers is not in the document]
nurul-umbhiya: we are calling this method multiple times, one from install and one from background process, so the wrapper method is required.

includes/ReverseWithdrawal/Manager.php
nurul-umbhiya: in_array() is needed here, there would be multiple return types and the  actual code will be based on that return  type, `false === $data` is checking the cache value only

templates/reverse-withdrawal/transaction-listing.php
nurul-umbhiya: that's the condition after echo, wc_price will do the necessary escaping here

templates/emails/plain/reverse-withdrawal-invoice.php
nurul-umbhiya: esc_attr is needed here, since we are not printing any HTML tags under plain text email

tanmayjay: By definition both `esc_html` and `esc_attr` do the same thing. The only difference is they use seperate filters.
But the guidelines say here `esc_html` should be used. `esc_attr` is meant for sanitizing the attribute values.
https://developer.wordpress.org/plugins/security/securing-output/#escaping

nurul-umbhiya: good catch, I was confused with wp_strip_all_tags

templates/reverse-withdrawal/transaction-listing.php
tanmayjay: Even if `wc_price` handles escaping, just to clarify our end, if possible, we should escape as late as possible. Also, `wc_price` has some filters, there's always a possibility of explosion. And applying escaping will just be a safe play from our end.

PR #1723, opened 2022-10-06
Fix Unwanted code remove for Sweetalert library

ManiruzzamanAkash:
**[fix]:** Fix Unwanted code remove for Sweetalert library.

**Related PR's in Sweetalert2 library** which generates this unwanted behavior.
https://github.com/sweetalert2/sweetalert2/pull/2428
https://github.com/sweetalert2/sweetalert2/pull/2462

### Wordpress org issue:
https://wordpress.org/support/topic/malicious-code-22/

### What has been done

1. We have remove the unwanted code from `sweetalert2.min.css` vendor file
2. We've downgraded sweetalert2 library to `11.4.8` in `package.json`

### Scenario:

If a russian user browse his website with `.ru` or `.su` domain, then it randomly shows this alert with unwanted text.

Demo Preview for Russian Users.


package.json
```diff
@@ -18,6 +18,7 @@
     "chartjs-adapter-moment": "^1.0.0",
     "debounce": "^1.2.1",
     "papaparse": "^5.3.2",
+    "sweetalert2": "11.4.8",
```
nurul-umbhiya, line 21: this is not necessary. kindly remove it from here

ManiruzzamanAkash: Why is not necessary ?

ManiruzzamanAkash: What would be the solution for then in Dokan Admin panel sweetalert2 popups?

nurul-umbhiya: we are not using any separate library for vue, we are using the default one. this is not necessary because we don't want to include same library twice.

nurul-umbhiya: beside including any libraries in the required section will increase the bundle size which is already way to high than the recommended value.

nurul-umbhiya: I'm approving this pr for now, since this code was added long ago. If you are curious, kindly check the product advertisement module code and check how I've implemented it without adding this library to bootstrap.

PR #1902, opened 2023-06-01
[Enhancement]: Adjusted the banner upload field style on store settings form

tanjiralmamun:
### All Submissions:

* [x] My code is tested

Right now, the Store Banner Upload field alignment was not correct. It also lefts a blank space if the dimension is changed. Therefore, I adjusted the style in this PR and it now looks like as below screenshots.

**Current:** https://snipboard.io/GzRmT4.jpg
**Changes:** https://snipboard.io/JoegBH.jpg

Closes #.

1.
2.
3.

### Changelog entry

> Enter a summary of all changes on this Pull Request. This will appear in the changelog if accepted.

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:


assets/src/less/settings.less
```diff
@@ -36,8 +36,8 @@
 
     .dokan-banner {
       border: 4px dashed #d8d8d8;
-      height: 300px;
-      margin: 0 0 30px 35px;
+      min-height: 300px;
```
nurul-umbhiya, line 39: only removing the height property will solve the issue since we are overriding this value from `templates/settings/store-form.php` file, there is no need to add min-height property.
Also setting left+right to auto for margin property will center the image. Since our setting form start from letf side, this will create design inconsistency. I think this should be changed from the theme end.

nurul-umbhiya (changes requested):

PR #1939, opened 2023-07-21
PDF invoice button on sub order

aihimel:
### All Submissions:

* [x] My code is tested

### Closes

* Closes https://github.com/weDevsOfficial/dokan-pro/issues/2468

* Check https://github.com/weDevsOfficial/dokan-pro/issues/2468

### Changelog entry

**PDF invoice on Sub Order section**
Previously PDF invoice was not visible on Sub Order section under customer order view. After this update customer will be able to view invoice button on sub order section.

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:


includes/wc-functions.php
```diff
@@ -1265,3 +1265,25 @@ function dokan_modify_vendor_order_counts( $counts ) {
...
+ * PDF Invoices & Packing Slips for WooCommerce plugin integration on suborder section.
+ *
+ * @since DOKAN_SINCE
+ *
+ * @param $actions
+ * @param $order
+ *
+ * @return mixed
+ */
+function dokan_pdf_invoice_button( $actions, $order ) {
```
nurul-umbhiya, line 1280: @aihimel Can you please move this function under includes/Order/Frontend/Hooks.php file?

nurul-umbhiya (changes requested):

includes/wc-functions.php
aihimel: moved

includes/Order/Hooks.php
```diff
@@ -61,12 +61,16 @@ public function __construct() {
 
         // Init Order Cache Class
         new OrderCache();
+
+        // Suborder pdf button
+        add_filter( 'dokan_my_account_my_sub_orders_actions', [ &$this, 'suborder_pdf_invoice_button' ], 10, 2 );
```
nurul-umbhiya, line 66: bhai no need to use & here, from php5.0 objects are always called by reference.

nurul-umbhiya (changes requested):

nurul-umbhiya (approved):

PR #1958, opened 2023-08-11
Fix: Invalid filter on lite and booking module

Aunshon:
### All Submissions:

* [x] My code is tested
* [x] I've included developer documentation (optional)

Fix Invalid filter on lite and booking module hide product type filter on product page

* Closes https://github.com/weDevsOfficial/dokan-pro/issues/2469
* Dokan pro PR: https://github.com/weDevsOfficial/dokan-pro/pull/2598

1.
2.
3.

### Changelog entry

> When only dokan lite is activated then we don't need to show the product type  filter because we support only one product type in dokan lite. And when woo-commerce booking integration is enabled in dokan then we won't show product type filter but in products page. but before in products page it was removed when booking module is enabled. Now if enable booking product the filter will be workable in products page.

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/Product/Hooks.php
+++ b/includes/Product/Hooks.php
@@ -29,6 +29,9 @@ class Hooks {
         add_action( 'woocommerce_new_product', [ $this, 'update_category_data_for_new_and_update_product' ], 10, 1 );
         add_action( 'woocommerce_update_product', [ $this, 'update_category_data_for_new_and_update_product' ], 10, 1 );
 
+        // Remove product type filter if pro not exists.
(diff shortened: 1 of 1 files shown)


nurul-umbhiya:
@Aunshon Since this is only required under product listing page, can we try a if condition like: `if ( count($args['product_types]) === 1 )` then do not load the product types filter html?

nurul-umbhiya (approved):

PR #1979, opened 2023-08-22
update: store category widget list default state set to collapse

### LINKED ISSUE #1909: Large Category display issues is always in open state
Dear,

we use the newest versions of Dokan as is per this post date. Dokan 3.7.19 and Dokan Pro 3.7.22 we use woodmart theme on the site throughout which looks nice with yours now there are 2 major issues! Check out the link provided. 

Issue 1: in large shops read this one has 11k+ articles and about 30 parent and 50+ child categories the default state of the category widget is OPEN state and shows all parent and categories open. Check out he link below.

https://kupoholik.rs/store/shoptogo/section/2206/

We need the default state to be closed and only triggered on click to open the respective category which is clicked. So click event based. On mobile view this looks even more ugly and a impossible scroll. The scroll on desktop is already huge... PLS ASAP help to fix this issue as we are going live in 10 days and this issue need to be fixed by then!! This is ugly and not a good promotion for you guys who can show off this site with large number of products.  Also maybe a smart thing to integrate  this when your already doing to implement the child visibilility via this css snippet as well:

.dokan-store .store-cat-stack-dokan.cat-drop-stack .children{
    padding-left: 5%;
}

Can you provide us the code snippet to rectify this issue so that the category tree is always closed and only on click is being opened, we will insert it via code snippet. Should be easy to be done...

Issue 2: please use the same link, on mobile view on and sometimes on desktop the first and second page of the shop breaks the banners and the pictures are out of alignment it a issue on desktop and mobile.  
image
image
image
image

You can see that the margins on the right side are way off, and there is a huge overflow there. This is screentshot from a p30 pro for example but a lot of android sow this in chrome browser. However when scrolling to the next page it corrects itself this is a issue which has been from the beginning and showing mostly on android devices.  We had this issue on the ihone 12 pro max as well.  This is the correct display...

image

Cheers,

Dragan

shohag121:
### All Submissions:

* [ ] My code is tested

* Closes #1909 .

1. check issue #1909

### Changelog entry

**update: store category widget list default state set to collapse.**
Previously, if a store has a product count over 100 or more and the store has many product categories, the store category widget would display those categories and subcategories in an open state rather than collapsed state that the sidebar style gets broken. Now the list has a max height of 500px, which will be visible, and other elements will be visible by scrolling and the parent category that has a submenu will be in collapse mode.

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:


assets/src/js/dokan-frontend.js
```diff
@@ -53,12 +53,22 @@
...
+      });
+
       let selectedLi = $('.cat-drop-stack ul').find('a.selected');
       selectedLi.css({ fontWeight: 'bold' });
 
       selectedLi.parents('ul.children').each(function(i, val) {
         $(val).css({ display: 'block' });
       });
+
+      let ulArea = $( '.store-cat-stack-dokan.cat-drop-stack ul');
```
nurul-umbhiya, line 70: can we move this CSS to corresponding less file?

nurul-umbhiya (changes requested):

assets/src/js/dokan-frontend.js
shohag121: Moved as per your recommendation. Thanks for the suggestion.

nurul-umbhiya (approved):

PR #2185, opened 2024-02-29
enhance: Fixed cart coupon support for multiple vendors

devAsadNur:
### All Submissions:

* [x] My code is tested

### Related Pull Request(s)

* Full PR Link

### Closes 

* Closes: https://github.com/getdokan/dokan-pro/issues/3104

* https://github.com/getdokan/dokan-pro/issues/3104#issuecomment-1970837526

### Changelog entry

*enhance: Fixed cart coupon support for multiple vendors*

### Before Changes

Coupons couldn't be applied to the whole cart for multiple vendors. Coupons could be applied only if all cart products are from a single vendor. So, there was no scope for doing calculations or splitting amounts.

### After Changes

Added support for fixed cart coupons to multiple vendors. The total discount amount will be split into line items based on WooCommerce's default behavior.

### Feature Video (optional)

### PR Self Review Checklist:

* Bad naming: make sure you understand your code if you read it a few months from now.
* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/Order/Hooks.php
+++ b/includes/Order/Hooks.php
@@ -328,10 +328,6 @@ class Hooks {
 
         $available_vendors = array_unique( $available_vendors );
 
-        if ( $coupon->is_type( 'fixed_cart' ) && count( $available_vendors ) > 1 ) {
-            throw new Exception( __( 'This coupon is invalid for multiple vendors.', 'dokan-lite' ) );
-        }
-
         // Make sure applied coupon created by admin
         if ( apply_filters( 'dokan_ensure_admin_have_create_coupon', $valid, $coupon, $available_vendors, $available_products ) ) {
             return true;


includes/Order/Hooks.php
```diff
@@ -328,10 +328,6 @@ public function ensure_vendor_coupon( $valid, $coupon, $discount ) {
 
         $available_vendors = array_unique( $available_vendors );
 
-        if ( $coupon->is_type( 'fixed_cart' ) && count( $available_vendors ) > 1 ) {
```
nurul-umbhiya, line 331: removing the code doesn't help. The calculation will be wrong. You can't apply fixed cart discount on multiple vendor products.

devAsadNur:
The total discount amount needs to be proportionally distributed among the products based on their prices. But, the WooCommerce doesn't distribute coupon amounts like this. They split the total coupon amount based on product units added to the cart.

As per discussion with @Mohaiminulislam1989 bhai, @shams-sadek1981 bhai and @tanjiralmamun bhai, we can't add support of fixed cart coupons for multiple vendors. Instead, we can hide the fixed cart coupon creation option for admin.

So, I'm closing this PR.

PR #2344, opened 2024-08-20
Fix: Prevent Stock Quantity Update During Product Edit if Unchanged

### LINKED ISSUE #2343: Stock Not Updating Correctly During Product Edit in Vendor Dashboard
### Bug Description

When a vendor edits a product in the Dokan vendor dashboard, the product's stock does not update correctly if sales occur while the vendor is editing. Specifically, the stock reverts to the original value that was loaded when the vendor opened the edit page, ignoring any sales that happened during the edit period.

Impact: This bug causes the stock quantity to be inaccurate, which can lead to overselling or mismanagement of inventory, impacting the overall store operations.

### Step To Reproduce

1. Go to the Dokan vendor dashboard.
2. Click on "Edit" for any product with a stock quantity.
3. While the edit page is open, have some sales occur on the product (reduce the stock by a few units).
4. Without changing the stock quantity in the edit form, click "Save" to update the product.
5. Check the product's stock quantity after saving.

### Expected Behaviour

The stock quantity should be updated to reflect the sales that occurred while the vendor was editing the product. For example, if the stock was initially 100 and 5 units were sold during the editing process, the stock should update to 95 when the vendor saves the product.

### Actual Behaviour

The stock quantity reverts to the original value that was loaded when the vendor first opened the edit page (e.g., 100), ignoring any sales that happened during the editing process. This results in an incorrect stock count.

### Additional Information

- There are no error messages shown, but the incorrect behavior can be consistently reproduced following the steps above.
- I have implemented a workaround by adding a hidden field in the form that stores the original stock and comparing it before updating the stock. However, this does not prevent the stock from being updated by other functions or hooks in Dokan.

```shell
1. Dokan Version: v3.11.5
2. Dokan Pro Version: not applicable
3. WooCommerce Version: v9.1.4
4. Browser: (e.g., Chrome, Safari)
5. Current Theme: Hello Elementor
```

brunomendespereira:
### All Submissions:

* [x] My code is tested

The issue addressed in this PR involves the incorrect stock update behavior in Dokan. When a vendor edits a product, any sales occurring during the edit process are not reflected in the stock count upon saving. This leads to the stock reverting to its original value, causing discrepancies.

This PR introduces a hidden field in the edit form that stores the original stock quantity. The stock is only updated if there is a change from the original value, ensuring that sales made during the editing process are accounted for.

### Related Pull Request(s)

* None

### Closes 

* Closes #2343 (Replace with your issue number)

1. Open a product in the Dokan vendor dashboard for editing.
2. Make some sales while the edit page is open.
3. Save the product without changing the stock quantity.
4. Verify that the stock correctly reflects the sales made during the editing process.

### Changelog entry

**Fix:** Ensure accurate stock updates when vendors edit products while sales occur.

### Before Changes

The stock would revert to the original value loaded when the vendor opened the edit page, ignoring any sales made during the editing process.

### After Changes

The stock quantity is correctly updated, considering any sales made during the vendor's edit session.

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

* [x] Correct — Does the change do what it’s supposed to? ie: code 100% fulfilling the requirements?
* [x] Secure — Would a nefarious party find some way to exploit this change? ie: everything is sanitized/escaped appropriately for any SQL or XSS injection possibilities?
* [x] Readable — Will your future self be able to understand this change months down the road?
* [x] Elegant — Does the change fit aesthetically within the overall style and architecture?

## Summary by CodeRabbit

- **New Features**
	- Enhanced stock management for variable products, optimizing updates based on changes in stock amounts.
	- Added a hidden input field to capture the original stock value in the inventory management form, improving data retention for validation.

- **Bug Fixes**
	- Improved handling of low stock amounts to ensure accurate alerts, regardless of stock changes.

- **Chores**
	- Updated the version constraint for the `brain/monkey` dependency to a development version, potentially incorporating new features.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "phpunit/phpunit": "^9.6",
     "wp-phpunit/wp-phpunit": "dev-master",
     "yoast/phpunit-polyfills": "^1.0@dev",
-    "brain/monkey": "^2.0"
+    "brain/monkey": "^2.0@dev"
   },
   "config": {
     "platform": {
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "phpunit/phpunit": "^9.6",
     "wp-phpunit/wp-phpunit": "dev-master",
     "yoast/phpunit-polyfills": "^1.0@dev",
-    "brain/monkey": "^2.0@dev"
+    "brain/monkey": "^2.0"
   },
   "config": {
     "platform": {
--- a/includes/wc-functions.php
+++ b/includes/wc-functions.php
@@ -288,16 +288,15 @@ function dokan_process_product_meta( int $post_id, array $data = [] ) {
     if ( 'yes' === get_option( 'woocommerce_manage_stock' ) ) {
         $manage_stock = 'no';
         $backorders   = 'no';
-        $stock        = '';
         $stock_status = wc_clean( $data['_stock_status'] );
-
+    
         if ( 'external' === $product_type ) {
             $stock_status = 'instock';
         } elseif ( 'variable' === $product_type ) {
-
+    
             // Stock status is always determined by children so sync later
             $stock_status = '';
-
+    
             if ( ! empty( $data['_manage_stock'] ) && $data['_manage_stock'] === 'yes' ) {
                 $manage_stock = 'yes';
                 $backorders   = wc_clean( $data['_backorders'] );
@@ -306,10 +305,10 @@ function dokan_process_product_meta( int $post_id, array $data = [] ) {
             $manage_stock = $data['_manage_stock'];
             $backorders   = wc_clean( $data['_backorders'] );
         }
-
+    
         update_post_meta( $post_id, '_manage_stock', $manage_stock );
         update_post_meta( $post_id, '_backorders', $backorders );
-
+    
         if ( $stock_status ) {
             try {
                 wc_update_product_stock_status( $post_id, $stock_status );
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -24,7 +24,7 @@
     "phpunit/phpunit": "^9.6",
     "wp-phpunit/wp-phpunit": "dev-master",
     "yoast/phpunit-polyfills": "^1.0@dev",
-    "brain/monkey": "^2.0"
+    "brain/monkey": "^2.0@dev"
```
shohag121, line 27: There's no need to include the `composer.json` and `composer.lock` files in this PR. Could you please revert those changes?

includes/wc-functions.php
```diff
@@ -288,16 +288,15 @@ function ( $value ) {
     if ( 'yes' === get_option( 'woocommerce_manage_stock' ) ) {
         $manage_stock = 'no';
         $backorders   = 'no';
-        $stock        = '';
         $stock_status = wc_clean( $data['_stock_status'] );
-
+
```
shohag121, line 292: Could you please remove the extra spaces?

shohag121 (commented):
You're doing an amazing job—keep up the great work!

includes/wc-functions.php
brunomendespereira: Done

composer.json
brunomendespereira: I reverted the composer.json to its previous state, but unfortunately, I lost the composer.lock file in the process. I had to generate a new one using composer update to ensure the dependencies were correctly locked in place.

mrabbani (approved):

brunomendespereira:
Hi,

Thank you very much for the detailed feedback. I have reviewed the errors mentioned and I'm currently looking into the causes. Please let me know if there is anything else I can do to assist, or if the test failures might have been due to automation issues. If any further action is needed from my side, I’m available to help.

Thank you again for your guidance.

Best regards,
Bruno Mendes

shohag121:
@shashwatahalder01 Could you please test this PR and let us know if any adjustments are required?

PR #3106, opened 2026-02-19
fix: add min max price for variable product in rest api

juah255:
### All Submissions:

* [x] My code is tested

### Related Pull Request(s)

* Full PR Link

### Closes 

* Closes https://github.com/getdokan/plugin-internal-tasks/issues/1487

* Steps or issue link

### Changelog entry

*Title*

### Before Changes

### After Changes

### Feature Video (optional)

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

## Summary by CodeRabbit

* **New Features**
  * REST product responses for variable products now include min_price and max_price to improve price display and filtering for storefronts and integrations.
  * The product REST schema exposes min_price and max_price as read-only fields so consumers can read these values reliably without altering server data.


MdAsifHossainNadim:
## Code Review: PR #3106 — Add min/max price for variable product in REST API

### Issues Found

#### 1. [ERROR] Wrong `@since` version tag
**Location:** `includes/REST/ProductController.php` → `add_min_max_price_to_variable_product()`

```php
// Current:
* @since 4.0.0

// Fix — should use placeholder since this is new code:
* @since DOKAN_SINCE
```

Dokan is already past version 4.0.0. New code should use the `DOKAN_SINCE` placeholder, which gets replaced at release time.

#### 2. [ERROR] Schema fields present but response fields conditionally absent
**Location:** `includes/REST/ProductController.php` → `get_item_schema()` and `prepare_data_for_response()`

The schema declares `min_price` and `max_price` as properties for ALL products, but the response only includes them for `variable` products. This means for simple/grouped/external products, the fields are in the schema but missing from the response — an API contract violation.

**Fix:** Always include the fields in the response, setting them to `null` for non-variable products:

```php
$data['min_price'] = $product->is_type( 'variable' ) ? $product->get_variation_price( 'min', true ) : null;
$data['max_price'] = $product->is_type( 'variable' ) ? $product->get_variation_price( 'max', true ) : null;
```

And update the schema type to allow null:

```php
'min_price' => [
    'description' => __( 'Minimum price for variable product.', 'dokan-lite' ),
    'type'        => [ 'string', 'null' ],
    'context'     => [ 'view', 'edit' ],
    'readonly'    => true,
],
```

#### 3. [WARNING] Schema fields placed before `id`
**Location:** `includes/REST/ProductController.php` → `get_item_schema()`

The `min_price` and `max_price` schema definitions are inserted at the very top of `properties`, before `id`. Convention is that `id` comes first. New fields should be added at the end or in a logical grouping (e.g., near `price`, `regular_price`, `sale_price`).

#### 4. [WARNING] `is_a()` instead of `instanceof`
**Location:** `includes/REST/ProductController.php` → `add_min_max_price_to_variable_product()`

```php
// Current:
if ( ! is_a( $product, 'WC_Product' ) ) {

// More idiomatic PHP:
if ( ! $product instanceof WC_Product ) {
```

`instanceof` is faster (no string lookup), more type-safe (IDE support), and the standard WordPress/WooCommerce pattern. The `WC_Product` class is already imported at the top of the file.

#### 5. [WARNING] Global side effect — WC filter registered from Dokan constructor
**Location:** `includes/REST/ProductController.php` → `__construct()`

```php
add_filter( 'woocommerce_rest_prepare_product_object', [ $this, 'add_min_max_price_to_variable_product' ], 10, 2 );
```

This hooks into WooCommerce's native REST response globally, meaning `min_price` and `max_price` will appear on `/wc/v3/products` for ALL consumers (not just Dokan). While adding read-only fields shouldn't break anything, it modifies another plugin's API contract.

Consider:
- Is this intentional? If so, document it in the method docblock.
- Would it be better to use Dokan's own filter (`dokan_rest_prepare_product_object`) or a separate hook registration rather than a REST controller constructor?

#### 6. [WARNING] Missing type hints on method parameters
**Location:** `includes/REST/ProductController.php` → `add_min_max_price_to_variable_product()`

```php
// Current:
public function add_min_max_price_to_variable_product( $response, $product )

// Better (matching docblock):
public function add_min_max_price_to_variable_product( WP_REST_Response $response, $product ): WP_REST_Response
```

Note: `$product` should remain untyped since the WC filter may pass unexpected types (the `is_a()` guard handles this). But `$response` and the return type can be typed.

---

### Verdict

The implementation is functionally correct — it covers both WC and Dokan REST endpoints with proper variable product detection. However, a few issues need addressing:

1. **Must fix:** Wrong `@since 4.0.0` → `DOKAN_SINCE`, schema/response consistency (fields should be present for all products as `null` when not variable)
2. **Should fix:** Move schema fields after `id`, use `instanceof`, add type hints, document the intentional WC global filter side effect

MdAsifHossainNadim (approved):
## Looks Good ✅

All previous issues addressed. Clean implementation — fields always present (`null` for non-variable), proper schema, type hints, documented WC hook.

Minor notes (can ignore):

### 1. [SUGGESTION] Static guard for filter registration
The `static $wc_filter_registered` pattern in the constructor works fine since the callback method is stateless. An alternative would be `has_filter()` check, but current approach is acceptable.

### 2. [SUGGESTION] `WC_Product_Variable` instanceof check
Good improvement over the previous `is_type('variable')` — more precise and type-safe.

mrabbani (approved):

PR #3117, opened 2026-03-05
feat: add vendor earning field to order REST API response

kzamanbd:
## Summary
- Adds a new `earning` field to the order REST API response that returns the vendor's earning for the order
- Uses `dokan()->commission->get_earning_by_order()` to calculate the earning and formats it with `wc_format_decimal()`

## Closes
* https://github.com/getdokan/plugin-internal-tasks/issues/1558

## Test plan
- [ ] Make a GET request to `dokan/v1/orders/{id}` as a vendor
- [ ] Verify the response contains an `earning` field with the correct decimal value
- [ ] Confirm the earning matches what is shown in the vendor dashboard for that order

## Summary by CodeRabbit

* **New Features**
  * Order API responses now include a new "earning" field showing vendor earnings directly in order data.
* **Bug Fixes**
  * Earnings are now formatted consistently and safely handled when unavailable, preventing incorrect or missing values in responses.

### DIFF
--- a/includes/REST/OrderController.php
+++ b/includes/REST/OrderController.php
@@ -295,10 +295,10 @@ class OrderController extends DokanRESTController {
 
         $vendor_earning = dokan()->commission->get_earning_by_order( $object->get_id() );
 
-        if ( is_wp_error( $vendor_earning ) ) {
-            $vendor_earning = null;
+        $formatted_earning = null;
+        if ( ! is_wp_error( $vendor_earning ) && ! is_null( $vendor_earning ) ) {
+            $formatted_earning = wc_format_decimal( $vendor_earning );
         }
-        $vendor_earning = wc_format_decimal( $vendor_earning );
 
         return array(
             'id'                   => $object->get_id(),
@@ -319,7 +319,7 @@ class OrderController extends DokanRESTController {
             'shipping_tax'         => $data['shipping_tax'],
             'cart_tax'             => $data['cart_tax'],
             'total'                => $data['total'],
-            'earning'              => $vendor_earning,
+            'earning'              => $formatted_earning,
             'total_tax'            => $data['total_tax'],
             'prices_include_tax'   => $data['prices_include_tax'],
             'customer_id'          => $data['customer_id'],
--- a/includes/REST/OrderController.php
+++ b/includes/REST/OrderController.php
@@ -295,6 +295,11 @@ class OrderController extends DokanRESTController {
 
         $vendor_earning = dokan()->commission->get_earning_by_order( $object->get_id() );
 
+        if ( is_wp_error( $vendor_earning ) ) {
+            $vendor_earning = null;
+        }
+        $vendor_earning = wc_format_decimal( $vendor_earning );
+
         return array(
             'id'                   => $object->get_id(),
             'parent_id'            => $data['parent_id'],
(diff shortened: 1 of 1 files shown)


MdAsifHossainNadim:
## Code Review: PR #3117 — Add Vendor Earning Field to Order REST API

### 1. [ERROR] Missing REST schema definition for `earning` field

**Location:** `includes/REST/OrderController.php` — `get_item_schema()` (line 911+)

The `earning` field is added to the response data in `get_formatted_item_data()` but has no corresponding entry in `get_item_schema()`. This breaks the REST API contract — clients relying on schema introspection won't discover the field, and WordPress REST infrastructure won't include it in `context`-filtered responses.

**Fix:** Add schema entry in `get_item_schema()` alongside the other financial fields:

```php
'earning' => array(
    'description' => __( 'Vendor earning for the order.', 'dokan-lite' ),
    'type'        => 'string',
    'context'     => array( 'view' ),
    'readonly'    => true,
),
```

---

### 2. [WARNING] Performance: N+1 commission calculation on list endpoint

**Location:** `includes/REST/OrderController.php:295-300`

`get_earning_by_order()` is called inside `get_formatted_item_data()`, which runs for **every order** in the `get_items()` list endpoint (line 484). The commission method involves DB queries, order meta lookups, and potentially full commission recalculation (`OrderCommission::calculate()`). For a page of 20 orders, that's 20+ extra DB round-trips.

**Fix:** Consider caching or batch-fetching earnings, or at minimum document the performance impact.

---

### 3. [WARNING] `earning` field breaks logical field grouping

**Location:** `includes/REST/OrderController.php:321` (in diff)

The `earning` field is inserted between `total` and `total_tax` — splitting the WooCommerce-standard financial fields. This could confuse API consumers expecting standard WC field ordering.

**Fix:** Place `earning` after the WooCommerce-standard fields (e.g., after `order_shipment` at the end of the response array, or after `total_tax`).

---

### 4. [WARNING] Missing `@since` tag

**Location:** `includes/REST/OrderController.php:295-300`

The new earning calculation block has no `@since` annotation. The field addition to the response should be documented with a version tag, ideally on the schema definition.

---

**Verdict:** Needs fixes before merge. The main blocker is #1 — the field must be registered in `get_item_schema()` for proper REST API compliance.

MdAsifHossainNadim (approved):
Looks Good ✅

Clean, focused addition. Null handling covers all `get_earning_by_order()` return types correctly.

**Suggestions (non-blocking):**

1. **Null earning has multiple meanings** — Parent orders (with sub-orders), errors, and unprocessed orders all return `null`. Consider expanding the schema description for clarity:
   ```php
   'description' => __( 'Vendor earning for the order. Null for parent orders with sub-orders.', 'dokan-lite' ),
   ```

2. **Schema context should include `'edit'`** — Currently `['view']` only, so the field won't appear in edit-context responses. Since `readonly: true` already prevents writes, adding `'edit'` makes the field available in more contexts:
   ```php
   'context' => array( 'view', 'edit' ),
   ```

3. **Performance note** — `get_earning_by_order()` runs for every order in list responses. Most orders short-circuit at cached meta checks, but unprocessed orders fall through to full commission calculation which can be heavier. Acceptable for typical use, just worth being aware of for large paginated lists.

includes/REST/OrderController.php
```diff
@@ -336,6 +343,7 @@ protected function get_formatted_item_data( $object ) {
             'coupon_lines'         => $data['coupon_lines'],
             'refunds'              => $data['refunds'],
             'order_shipment'       => $data['order_shipment'],
+            'earning'              => $formatted_earning,
```
mrabbani, line 346: Pls add prefix `vendor_earning`.

mrabbani (approved):

PR #3121, opened 2026-03-10
feat: manage update notification dot visibility for Dokan Lite/Pro

Shamim-97:
### All Submissions:

* [ x My code satisfies feature requirements
* [x] My code is tested

### Related Pull Request(s)

* Full PR Link

### Closes 

* Closes https://github.com/getdokan/dokan-pro/issues/5467

* Steps or issue link

### Changelog entry

*Title*

### Before Changes

### After Changes

### Feature Video (optional)

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the admin dashboard's handling of version notifications to properly dismiss notices when new versions are detected and accessed.

### DIFF
--- a/src/admin/dashboard/pages/changelog/index.tsx
+++ b/src/admin/dashboard/pages/changelog/index.tsx
@@ -48,7 +48,9 @@ const ChangelogPage = () => {
             if ( pkg === 'pro' && proVersions !== null ) {
                 return;
             }
-
+            if ( dokanAdminDashboardSettings?.header_info?.has_new_version ) {
+                 dismissWhatsNewNotice( pkg );
+            }
             setLoading( true );
             try {
                 const response = await apiFetch< string >( {
@@ -74,6 +76,25 @@ const ChangelogPage = () => {
         },
         [ liteVersions, proVersions ]
     );
+    const dismissWhatsNewNotice = async (pkg: 'lite' | 'pro') => {
+        const action = pkg === 'pro' ? 'dokan-pro-whats-new-notice' : 'dokan-whats-new-notice';
+
+        try {
+            const response = await fetch(ajaxurl, {
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/x-www-form-urlencoded',
+                },
+                body: new URLSearchParams({
+                    action: action,
+                    nonce: dokanAdminDashboard?.nonce || '',
+                    dokan_promotion_dismissed: 'true',
+                }),
+            });
+        } catch (error) {
+            console.error('Request failed:', error);
+        }
+    };
 
     // Load initial data
     useEffect( () => {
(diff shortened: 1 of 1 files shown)


MdAsifHossainNadim (changes requested):
## Request for Changes

The approach is clean — visiting changelog to auto-dismiss the "What's New" notice makes sense. A few things need to be fixed before merging.

---

### Needs Fix

**1. Unused `response` variable (lint error)**

```tsx
const response = await fetch(ajaxurl, {
```
`response` is assigned but never used. Either remove the assignment (`await fetch(...)`) or check `response.ok` for proper error handling.

**2. Code formatting doesn't follow project standards**

The new code is missing spaces inside parentheses per WordPress JS coding standards:

```tsx
// Current:
const dismissWhatsNewNotice = async (pkg: 'lite' | 'pro') => {
} catch (error) {
body: new URLSearchParams({

// Should be:
const dismissWhatsNewNotice = async ( pkg: 'lite' | 'pro' ) => {
} catch ( error ) {
body: new URLSearchParams( {
```

Please run `npm run lint:js` and fix formatting.

**3. Use `apiFetch` instead of raw `fetch`**

The rest of the component uses `@wordpress/api-fetch` for API calls. The new function uses raw `fetch` to `ajaxurl`, which is inconsistent and bypasses WordPress's built-in request handling (nonce headers, middleware, error handling). Please use `apiFetch` or add a comment explaining why raw `fetch` is needed here.

---

### Suggestion

**`dismissWhatsNewNotice` is not in `fetchChangelog`'s `useCallback` dependency array** — it could reference a stale closure. Consider adding it to the deps, wrapping it in `useCallback`, or moving the dismiss logic into a separate `useEffect`.

MdAsifHossainNadim (approved):
### Looks Good ✅

**Suggestions:**

- `dismissWhatsNewNotice( pkg )` is not awaited at `index.tsx:79-81`, and `has_new_version` (set via `wp_localize_script`) never updates client-side. Toggling between lite/pro fires duplicate dismiss AJAX requests. Consider tracking dismissed state locally with a `useState< Set< string > >` and `await` the call to prevent redundant requests.

- Only `wp_ajax_dokan-whats-new-notice` is registered in `includes/Admin/Notices/WhatsNew.php:18`. No `wp_ajax_dokan-pro-whats-new-notice` handler exists in Dokan Lite. Please verify the handler exists in Dokan Pro — if not, the pro dismiss silently fails with a `400` and the notification dot persists. Either confirm & add a code comment noting the Pro dependency, or use the same action for both and differentiate via a body parameter.

dev-shahed:
Notes: needs to test within zip

PR #3151, opened 2026-04-23
Refactor: product table for admin products

rakib1904107:
### All Submissions:

* [x] My code is tested

### Related Pull Request(s)

* Full PR Link: https://github.com/getdokan/dokan-pro/pull/5570

### Closes 

* Closes # https://github.com/getdokan/plugin-internal-tasks/issues/1691

* Steps or issue link

### Changelog entry

*Title*

### Before Changes

### After Changes

### Feature Video (optional)

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/REST/ProductController.php
+++ b/includes/REST/ProductController.php
@@ -53,7 +53,7 @@ class ProductController extends DokanRESTController {
     /**
      * Post status
      */
-    protected $post_status = [ 'publish', 'pending', 'draft' ];
+    protected $post_status = [ 'publish', 'pending', 'draft', 'future' ];
 
     /**
      * Class constructor.
@@ -69,6 +69,8 @@ class ProductController extends DokanRESTController {
             add_filter( 'woocommerce_rest_prepare_product_object', [ $this, 'add_min_max_price_to_variable_product' ], 10, 2 );
             $wc_filter_registered = true;
         }
+
+        $this->post_status = dokan_get_post_status() ? array_keys( dokan_get_post_status() ) : $this->post_status;
     }
 
     /**
@@ -537,12 +539,23 @@ class ProductController extends DokanRESTController {
     public function get_product_summary( $request ) {
         $seller_id = dokan_get_current_user_id();
 
+        // Match the list's exclusion scope so tab counts can't drift from it —
+        // Pro modules add `product_pack` etc. via this filter.
+        $exclude_types = array_values(
+            array_unique(
+                array_merge(
+                    [ 'booking', 'auction' ],
+                    (array) apply_filters( 'dokan_product_listing_exclude_type', [] )
+                )
+            )
+        );
+
         $data = [
-            'post_counts'      => dokan_count_posts( 'product', $seller_id ),
+            'post_counts'      => dokan_count_posts( 'product', $seller_id, $exclude_types ),
             'products_url'     => dokan_get_navigation_url( 'products' ),
-            'instock_count'    => dokan_count_stock_posts( 'product', $seller_id, 'instock' ),
-            'outofstock_count' => dokan_count_stock_posts( 'product', $seller_id, 'outofstock' ),
-            'months'           => $this->get_product_months_data( dokan_get_current_user_id() ),
+            'instock_count'    => $this->count_products_by_stock_status( $seller_id, 'instock', $exclude_types ),
+            'outofstock_count' => $this->count_products_by_stock_status( $seller_id, 'outofstock', $exclude_types ),
+            'months'           => $this->get_product_months_data( $seller_id ),
         ];
 
         /**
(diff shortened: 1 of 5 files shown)


MdAsifHossainNadim:
## Review

Two coordinated changes for the admin "Products" dashboard list:

1. **Lite** — Bundle the per-action `/products/summary` call into the `/products` collection response via a new `X-Dokan-Product-Summary` header. Adds `future` to default post statuses, dynamically picks up Pro statuses (e.g. `reject`) via `dokan_get_post_status()`, and exposes a new `dokan_product_list_table_tabs` JS filter so Pro can inject status tabs.
2. **Pro** — Adds a new "Rejected" tab via the JS filter, and an `include_hidden_products` request flag so the dashboard list bypasses the `exclude-from-catalog` visibility filter that the subscription / RFQ modules apply by default.

The intent is good — eliminating the duplicate round-trip and making the tabs extensible.

---

### CRITICAL

#### 1. `isset()` check on `include_hidden_products` is a permanent bypass

**Pro — 3 sites**: `RequestForQuotationController.php:1179`, `subscription/Helper.php:1286`, `subscription/ProductStatusChanger.php:179`.

```php
if ( isset( $params['include_hidden_products'] ) ) {
    return $args; // skips exclude-from-catalog
}
```

Any request that includes the param at all — even `?include_hidden_products=0` or `=false` — bypasses the visibility filter. Worse, in `ProductStatusChanger::filter_products_for_api`:

```php
$args['include_hidden_products'] = $request->get_param( 'include_hidden_products' );
```

This unconditionally writes the key into `$args` even when the request did not send it (`get_param` returns `null`, but `isset()` on the array key downstream still returns `true`). The downstream `Helper::set_default_tax_query` then permanently skips visibility for any caller that goes through `filter_products_for_api`.

**Fix:**

```php
if ( dokan_string_to_bool( $request->get_param( 'include_hidden_products' ) ) ) {
    return $args;
}
```

Plus drop the unconditional `$args[...] = $request->get_param(...)` write — replace with a conditional set, or pass the value explicitly into the helper.

This needs to be a registered, sanitized parameter on the Lite endpoint schema (`get_product_collection_params`), not a free-form magic key.

---

### ERROR

#### 2. `count_products_by_stock_status()` duplicates `dokan_count_stock_posts()` and loses caching + Pro extensibility

`includes/REST/ProductController.php` (new method, ~line 612)

`dokan_count_stock_posts( 'product', $seller_id, $stock_status, $exclude_types )` already exists, already accepts the exclude-types argument, and provides:

- Object cache via `Cache::get/set` keyed per vendor
- `dokan_count_posts_{$stock_type}` filter for Pro override
- `dokan_count_stock_posts_cache_key` filter

The new in-controller `WP_Query` skips all three — every collection request now triggers two uncached `WP_Query`s (in-stock + out-of-stock) per vendor.

**Fix:** Just call the existing helper:

```php
'instock_count'    => dokan_count_stock_posts( 'product', $seller_id, 'instock', $exclude_types ),
'outofstock_count' => dokan_count_stock_posts( 'product', $seller_id, 'outofstock', $exclude_types ),
```

Then delete `count_products_by_stock_status()`.

#### 3. `format_collection_response` override leaks vendor counts logic into public endpoints

`top_rated`, `best_selling`, and `related_product` all call `$this->format_collection_response( ... )` (lines 640, 667, and the related-products method). The override now runs on those calls too.

The PR works around this by guarding with `get_product_summary_permissions_check()`, but that's a behavioral patch over a structural problem — the override is doing two unrelated jobs (pagination + dashboard summary). A reviewer/maintainer touching either path later can easily break the assumption.

**Suggested fix:** Don't override `format_collection_response`. Instead, attach the summary header in `get_items()` (or wherever the dashboard list endpoint is served) only after the parent call returns. The header logic belongs next to the endpoint that needs it, not in a shared formatter.

#### 4. Header payload can exceed practical proxy/server limits

`X-Dokan-Product-Summary` contains `wp_json_encode( $summary->get_data() )` — `post_counts`, `months[]` (up to N months of vendor product history), `subscription_remaining` from Pro, plus anything `dokan_product_listing_summary_data` filter appends.

Most reverse proxies cap individual headers around 8 KB (nginx default `large_client_header_buffers` is 8K; CloudFront 16K total response headers). For an old vendor with many months of history plus a few Pro filters, this can silently get stripped or truncated.

**Suggested fix:** Either return the summary in the response body (e.g., as a `_summary` envelope alongside `data`), or split into a couple of small numeric headers and keep the heavy bits (`months`, `subscription_remaining`) on the existing `/summary` endpoint. The free-form Pro filter `dokan_product_listing_summary_data` is the riskiest part.

#### 5. Public API behavior change: default `post_status` widened for every consumer

```php
$this->post_status = dokan_get_post_status() ? array_keys( dokan_get_post_status() ) : $this->post_status;
```

Every caller of `/dokan/v1/products` who doesn't pass an explicit `status` param now gets `publish + pending + draft + future` (and any Pro-registered statuses like `reject`). Previously just `publish + pending + draft`.

This affects 3rd-party integrations and is not documented in the changelog. At minimum it needs a `@since` note on the property and a changelog entry. Consider gating "future" behind an opt-in param for the dashboard list specifically rather than changing the controller-wide default.

---

### WARNING

#### 6. `'reject'` status: hardcoding in Lite is wrong layering

`src/dashboard/products/ProductList.tsx:35-53`:

```ts
case 'reject': return 'danger';
case 'reject': return __( 'Rejected', 'dokan-lite' );
```

`'reject'` is a Pro concept (product-rejection module). Hardcoding it in Lite leaks Pro vocabulary into Lite. Lite already falls back to `status` slug for unknown statuses in `useProducts.applySummary` (`knownLabels[key] ?? key`), and Pro registers the tab via `dokan_product_list_table_tabs`. Pro should also register the badge variant + label via a similar JS filter (e.g., `dokan_product_status_label`, `dokan_product_status_badge_variant`) — keep Lite ignorant of Pro statuses.

#### 7. Pro tab text-domain inconsistency

`dokan-pro/src/features/products/tabs/index.tsx:32`:

```ts
label: __( 'Rejected', 'dokan' ),
```

Verify the rest of dokan-pro frontend uses `'dokan'` (some files use `'dokan-lite'` after the consolidation). Worth grepping siblings to be consistent.

#### 8. Initial mount no longer fetches summary independently

After removing the `useEffect( fetchStatusCounts )`, the very first render shows zeroed `statusCounts` until `fetchProducts` resolves. With slow networks or auth retries, the tabs flash "All (0) | Published (0) | …" before populating. Minor UX regression. The placeholder `useState` initializer in `useProducts.ts:55-83` masks this for the known statuses, but Pro's `Rejected` tab (count from filter) shows 0 on first paint.

#### 9. `count_products_by_stock_status` query is missing cache flags

`'fields' => 'ids'` is good, but for a count-only query (`posts_per_page => 1`, only `found_posts` used), passing `'update_post_meta_cache' => false` and `'update_post_term_cache' => false` further trims the work. Moot if you switch to the existing `dokan_count_stock_posts()` helper per #2.

---

### SUGGESTION

#### 10. `exclude_types` derivation duplicated

`includes/REST/ProductController.php:543-552` reproduces what `dokan_count_posts()` already does internally (it defaults `$exclude_product_types` to `[ 'booking', 'auction' ]`). Either expose a small helper in `includes/functions.php` (`dokan_get_excluded_product_types()`) or rely on the helper's default and only merge Pro extras when needed.

#### 11. TS `( string & {} )` escape hatch in `ProductStatus`

```ts
export type ProductStatus = 'all' | 'publish' | … | ( string & {} );
```

Acceptable pattern for "known + arbitrary string". A short comment noting "Pro modules may register additional statuses (e.g. `reject`)" would help the next reader.

#### 12. PR template gaps

Both PRs leave **Changes proposed**, **How to test**, **Changelog entry**, **Before/After Changes** empty. Reviewers and QA will need these — please fill before requesting QA. Also tick the *Changelog entry* checklist item and add a `Changelog Updated` label once added.

---

### Test Coverage

No PHPUnit changes for the new public method `count_products_by_stock_status` or the new `format_collection_response` override; no Playwright update for the new "Rejected" tab. Given the public-API behavior change in #5 and the security-sensitive flag in #1, please add:

- PHPUnit: `format_collection_response` returns summary header only for users with `dokan_view_product_status_report` cap.
- PHPUnit: `include_hidden_products=0` does **not** bypass visibility (regression test for #1).
- Playwright: vendor with rejected products sees the Rejected tab and a non-zero count.

---

### Summary

Direction is right (one round-trip, JS-filter extensibility), but the security bypass in #1 and the duplicated/uncached count logic in #2 must be addressed before merge. #3 and #4 are architectural — worth fixing now while the surface is small. The Lite/Pro coupling around `'reject'` (#6) should also be cleaned up so Pro-only vocabulary doesn't leak into Lite.

kzamanbd (approved):

PR #3226, opened 2026-05-22
fix: Approve exact fractional balance without false rejection.

MdAsifHossainNadim:
### All Submissions:

* [ ] My code is tested

Admin approval of a withdraw request was failing with "You don't have enough balance for this request" when the requested amount exactly equalled the vendor's remaining balance and either side carried decimals. The same boundary also caused `Hooks::update_vendor_balance()` to silently skip the credit-row insert, leaving an approved withdraw with no offsetting balance entry.

Root cause: both sites compared `dokan_get_seller_balance()` (a `float`) against the request amount with strict `>` / `<`, and IEEE-754 drift from float subtraction (e.g. balance computed as `5.4999999998` after a partial approval instead of `5.5`) flipped the boundary in production data.

Fix: compare in integer cents (`* 100`, rounded) in both call sites so the drift is rounded away into the same bucket without changing the rest of the validation contract. No new dependency added (no bcmath); matches WooCommerce's own scaled-integer pattern.

Files changed:
- `includes/Withdraw/Manager.php` — `is_valid_approval_request()` pre-approval guard.
- `includes/Withdraw/Hooks.php` — `update_vendor_balance()` post-approval credit-insert guard (also drops the now-unused `NumberUtil` import).

### Related Pull Request(s)

* Full PR Link

### Closes 

* Closes getdokan/dokan-pro#5675

* Steps or issue link

1. As a vendor, accumulate a withdrawable balance of `10.5` (e.g., insert a debit row of `10.5` in `wp_dokan_vendor_balance`).
2. Submit a withdraw request for `5` and have admin approve it. Confirm the credit row is inserted and the remaining balance is `5.5`.
3. Submit a second withdraw request for `5.5` (the exact remaining balance) and have admin approve it.
4. **Expected:** approval succeeds, credit row is inserted, vendor balance lands at exactly `0`.
5. Repeat with the order reversed (`5.5` first, then `5`) to confirm there is no regression for the previously-working path.

### Changelog entry

*Title*

Fix: withdraw approval failing on exact fractional balance

Previously, approving a withdraw request whose amount equalled the vendor's remaining balance — when the balance carried decimals from prior partial approvals — could fail with "You don't have enough balance for this request" or, worse, succeed in flipping the withdraw status to approved without inserting the matching credit row in `wp_dokan_vendor_balance`. The behaviour depended on the order in which decimal-vs-whole amounts were approved.

After this change, the balance-vs-amount comparison is performed in integer cents, so floating-point drift can no longer flip the strict-comparison boundary. An exact-balance withdrawal approves cleanly and the credit row is always written when the request reaches that step.

### Before Changes

Reproduction video (from the linked issue): https://d.pr/v/E3NVQj

### After Changes

### Feature Video (optional)

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

* [ ] Correct — Does the change do what it's supposed to? ie: code 100% fulfilling the requirements?

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved precision of withdrawal balance validation during approval processing
  * Fixed notification display layering in the withdraw interface

### DIFF
--- a/includes/Withdraw/Hooks.php
+++ b/includes/Withdraw/Hooks.php
@@ -2,8 +2,6 @@
 
 namespace WeDevs\Dokan\Withdraw;
 
-use Automattic\WooCommerce\Utilities\NumberUtil;
-
 class Hooks {
 
     /**
@@ -112,7 +110,11 @@ class Hooks {
     public function update_vendor_balance( $withdraw ) {
         global $wpdb;
 
-        if ( NumberUtil::round( dokan_get_seller_balance( $withdraw->get_user_id(), false ), 2 ) < NumberUtil::round( $withdraw->get_amount(), 2 ) ) {
+        // Compare in integer cents — over-rejecting on float drift would leave an approved withdraw with no offsetting credit row.
+        $balance_cents = (int) round( (float) dokan_get_seller_balance( $withdraw->get_user_id(), false ) * 100 );
+        $amount_cents  = (int) round( (float) $withdraw->get_amount() * 100 );
+
+        if ( $balance_cents < $amount_cents ) {
             return;
         }
 
--- a/includes/Withdraw/Manager.php
+++ b/includes/Withdraw/Manager.php
@@ -41,7 +41,11 @@ class Manager {
             return new WP_Error( 'dokan_withdraw_empty', __( 'Withdraw amount required ', 'dokan-lite' ) );
         }
 
-        if ( $amount > $balance ) {
+        // Compare in integer cents so an exact-balance withdrawal is not rejected by IEEE-754 drift.
+        $balance_cents = (int) round( (float) $balance * 100 );
+        $amount_cents  = (int) round( (float) $amount * 100 );
+
+        if ( $amount_cents > $balance_cents ) {
             return new WP_Error( 'dokan_withdraw_not_enough_balance', __( 'You don\'t have enough balance for this request', 'dokan-lite' ) );
         }
 
--- a/src/admin/dashboard/pages/withdraw/index.tsx
+++ b/src/admin/dashboard/pages/withdraw/index.tsx
@@ -326,10 +326,15 @@ const WithdrawPage = () => {
             isEligible: ( item ) => item?.status === 'pending',
             callback: ( items ) => {
                 const failedItems = items.filter( ( item ) => {
-                    const requestedAmount = parseFloat( item?.amount ) || 0;
-                    const currentBalance = parseFloat( item?.user?.balance ) || 0;
-                    
-                    return currentBalance < requestedAmount;
+                    // Compare in integer cents so float drift can't reject an exact-balance request — mirrors the backend approval guard.
+                    const requestedCents = Math.round(
+                        ( parseFloat( item?.amount ) || 0 ) * 100
+                    );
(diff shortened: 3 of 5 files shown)


MdAsifHossainNadim:
## Verdict: Approve ✅

Reviewed against `.claude/skills/dokan-code-review` (PR template, REST/security checklist, label taxonomy).

### What the PR does

1. **`includes/Withdraw/Manager.php` — `is_valid_approval_request()`**
   Replaces the raw `$amount > $balance` strict-float compare with an integer-cents comparison after the existing `wc_format_decimal( …, 2 )` rounding. Defensive symmetry with the Hooks fix below; with `wc_format_decimal` already shielding the inputs, this is primarily a consistency/intent change.

2. **`includes/Withdraw/Hooks.php` — `update_vendor_balance()`** *(the actual bug)*
   The pre-fix `NumberUtil::round( $balance, 2 ) < NumberUtil::round( $amount, 2 )` operated on raw floats with no upstream shielding, so IEEE-754 drift could round the two sides into different 2-dp buckets and silently `return` before inserting the credit row — leaving the withdraw approved but with no offsetting entry in `wp_dokan_vendor_balance`. Now compared in integer cents (`(int) round( $value * 100 )`), so a `5.4999999998`-style drift collapses into the same bucket as `5.5`. Also drops the now-unused `NumberUtil` import.

3. **`src/layout/index.tsx` + `src/dashboard/withdraw/tailwind.scss` — toaster z-index**
   `react-hot-toast` ships with an inline `z-index: 9999`, which renders *behind* the WP modal backdrop (`100000`). With the vendor withdraw flow using a modal for confirmations, any approval/rejection toast was hidden. The fix opts the toaster into a `dokan-toaster` container class and raises it to `100001 !important` when the vendor withdraw wrapper is on the page.

### Verification

- Backend: walked the boundary math; integer-cents collapses drift around the 2-dp boundary correctly, no behavior change for clearly-under or clearly-over cases.
- Frontend: confirmed in the browser at `/dashboard/new/#withdraw` — the toaster container resolves to a computed `z-index: 100001`, sitting above the modal backdrop (`100000`). Visually verified by triggering a toast over an open modal: the toast renders on top.

### Carry-over suggestions (non-blocking)

- The `Manager.php` change is structurally a no-op given the upstream `wc_format_decimal( …, 2 )`. Worth keeping for symmetry, but not behavior-changing on its own.
- 2-dp precision is hardcoded; future-proofing for stores with `wc_get_price_decimals() != 2` (e.g. crypto) could thread `pow(10, wc_get_price_decimals())` through both call sites.
- PHPUnit boundary tests for `is_valid_approval_request` and `update_vendor_balance` would lock the contract against future drift-handling regressions — recommended as a follow-up.
- The admin React dashboard withdraw page (`src/admin/dashboard/pages/withdraw`) also opens a `DokanModal` for approve/cancel actions and is the original symptom site; a parallel `containerClassName` + scoped z-index rule there would close the visible-toast gap on the admin side as well.

dev-shahed:
### Issue not solved, while approving from here: ###
image

### Solved from here:  ###
image

@MdAsifHossainNadim

mrabbani (approved):

PR #3253, opened 2026-06-09
fix(reverse-withdrawal): single COD enable toggle in new settings

akzmoudud:
## Summary

Maps the two legacy Reverse Withdrawal settings ("Enable Reverse Withdrawal" + "Enable Reverse Withdrawal for this Gateway") onto the single new **"Activate Reverse Withdrawal (Cash On Delivery)"** toggle, and fixes a bug that left the feature permanently disabled.

Refs and closes: getdokan/plugin-internal-tasks#1943

## What changed

- **Single toggle in new settings.** Removed the standalone `reverse_withdrawal_payment_gateways` multicheck field (and its builder) from `SettingsSchema`. The new UI now shows one switch, mapped 1:1 to legacy `dokan_reverse_withdrawal.enabled`. COD is the only shipped gateway, so it follows the master switch.
- **Gateway list derives from the toggle.** `SettingsHelper::get_enabled_payment_gateways()` now returns the registered gateways when `is_enabled()` is true (and `[]` otherwise), instead of reading the now-unused `payment_gateways` option. The 2nd legacy field is ignored at runtime.
- **Bug fix — `is_enabled()` was always `false`.** It read the *new* flat id `reverse_withdrawal_enabled` out of the *legacy* `dokan_reverse_withdrawal` section. The bridge overlay only ever projects onto the legacy path `enabled`, so that key never existed there and the helper was stuck at `'off'` — disabling every Reverse Withdrawal hook (order, cart, cron, dashboard) even when the admin had turned it on. Now reads the legacy `enabled` key, consistent with every sibling helper.

## Why ignore the gateway field

COD is the only available gateway. Gating the feature behind a second checkbox added no value, so the master switch alone now controls it — enabling it activates COD reverse withdrawal directly.

## Verification

Exercised the full read/write round-trip via `wp eval` against both stores:

| Scenario | new id | `is_enabled()` | COD gate |
|---|---|---|---|
| New toggle ON | `on` | `true` | `true` |
| Old form: master ON, gateway unchecked | `on` | `true` | `true` |
| New toggle OFF | `off` | `false` | `false` |

All other Reverse Withdrawal fields (`billing_type`, `reverse_balance_threshold`, `monthly_billing_day`, `due_period`, `failed_actions`, `display_notice`, `send_announcement`) remain mapped as before.

### DIFF
--- a/includes/Admin/Settings/Schema/SettingsSchema.php
+++ b/includes/Admin/Settings/Schema/SettingsSchema.php
@@ -1200,52 +1200,6 @@ class SettingsSchema {
 					'value' => 'off',
 				],
             ],
-            self::reverse_withdrawal_payment_gateways_field(),
-        ];
-    }
-
-    /**
-     * Build the `reverse_withdrawal_payment_gateways` multicheck field.
-     *
-     * Options are derived from the `dokan_reverse_withdrawal_payment_gateways`
-     * filter (the same source the legacy {@see SettingsHelper::get_reverse_withrawal_payment_gateways()}
-     * uses) so Pro modules / extensions that already extend that filter continue
-     * to register gateways with one declaration. The new field stores a flat
-     * list of selected gateway ids; each slot bridges through
-     * {@see MulticheckArrayTransformer} to the legacy WP-multicheck shape under
-     * `dokan_reverse_withdrawal.payment_gateways`.
-     *
-     * @return array
-     */
-    private static function reverse_withdrawal_payment_gateways_field(): array {
-        $gateways = apply_filters(
-            'dokan_reverse_withdrawal_payment_gateways',
-            [ 'cod' => esc_html__( 'Cash on delivery', 'dokan-lite' ) ]
-        );
-
-        $options    = [];
-        $legacy_key = [];
-        foreach ( $gateways as $value => $label ) {
-            $options[]            = [
-                'value' => (string) $value,
-                'label' => (string) $label,
-            ];
-            $legacy_key[ $value ] = 'dokan_reverse_withdrawal.payment_gateways.' . $value;
-        }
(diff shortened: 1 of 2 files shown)


akzmoudud:
### Testing notes

**Behaviour change:** the old `Enable Reverse Withdrawal for this Gateway` field no longer has any impact on reverse withdrawal. The feature is now controlled solely by `Enable Reverse Withdrawal`, which is **one-to-one mapped** with the new settings toggle `Activate Reverse Withdrawal (Cash On Delivery)`.

**How to test**

Old settings ↔ new settings mapping:
- [ ] Enable `Enable Reverse Withdrawal` in old settings → new toggle `Activate Reverse Withdrawal (Cash On Delivery)` turns **ON**.
- [ ] Disable `Enable Reverse Withdrawal` in old settings → new toggle turns **OFF**.
- [ ] Enable the new toggle → old `Enable Reverse Withdrawal` turns **ON**.
- [ ] Disable the new toggle → old `Enable Reverse Withdrawal` turns **OFF**.

Gateway field has no effect:
- [ ] Enable `Enable Reverse Withdrawal` and **uncheck** `Enable Reverse Withdrawal for this Gateway` → reverse withdrawal still **works** for COD (place a COD order, confirm the reverse withdrawal charge is recorded).
- [ ] Enable `Enable Reverse Withdrawal` and **check** the gateway field → behaves the same (no difference).
- [ ] Disable `Enable Reverse Withdrawal` (regardless of the gateway field) → reverse withdrawal does **not** run.

Other reverse withdrawal fields (`Billing Type`, `Reverse Balance Threshold`, `Monthly Billing Date`, `Grace Period`, `After Grace Period` actions, `Display Notice`, `Send Announcement`) should continue to save and map correctly — no regression.

kzamanbd (approved):

PR #3288, opened 2026-06-29
fix: Enforce per-product ownership on v3 products batch endpoint.

MdAsifHossainNadim:
### All Submissions:

* [x] My code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)
* [x] My code satisfies feature requirements
* [ ] My code is tested — no automated test in this PR (see **Test scope**); manual REST steps provided
* [ ] My code passes the PHPCS tests — not run locally (dev deps absent); CI PHPCS will verify
* [x] My code has proper inline documentation

### Changes proposed in this Pull Request:

**Problem** — automated WordPress.org security scan (`dokan-lite 5.0.6`), tracked as `getdokan/dokan-pro#5818`.

The `dokan/v3/products/batch` REST route authorized callers with a **capability-only** check — `batch_items_permissions_check()` → `current_user_can( 'dokan_edit_product' )` — whereas the single-item update route (`/products/(?P<id>)`) *also* verifies **ownership** via `check_permission()` → `dokan_is_product_author()`. The scanner flagged this asymmetry as a broken-authorization / IDOR risk: an authenticated vendor could submit **another vendor's** product ID in a batch `update`/`delete`.

**Root cause** — `batch_items()` delegates to `WC_REST_Controller::batch_items()`, which processes each item by calling the controller's *per-item* permission methods (`update_item_permissions_check()` / `delete_item_permissions_check()`). The v3 controller did not override them, so the batch path never ran Dokan's `dokan_is_product_author()` ownership gate that the single-item route enforces.

**How it's resolved** — add the two per-item overrides that WooCommerce's batch loop invokes:

| Method (new) | Gate applied per item |
|---|---|
| `update_item_permissions_check()` | delegates to existing `check_permission()` → `dokan_edit_product` **+ `dokan_is_product_author()`** |
| `delete_item_permissions_check()` | `dokan_delete_product` **+ `dokan_is_product_author()`** |

The route-level `batch_items_permissions_check()` is intentionally kept as the coarse capability "door" (with a clarifying comment); per-product ownership is now enforced **per item**, matching the single-item route. The `create` array is unchanged — new products are authored by the current vendor. Note the v3 controller exposes product deletion **only** through the batch route, so `delete_item_permissions_check()` is the sole ownership gate for deletes.

[...]

PR #3298, opened 2026-06-30
test: stabilize v3 products batch-authorization test against seller-cap pollution

shohan0120:
### All Submissions:

* [x] My code is tested

`ProductControllerV3BatchAuthorizationTest` (introduced in #3288) **passes in isolation but fails in the full PHPUnit run** (`Tests: 245, … Errors: 1, Failures: 5`).

Root cause is test pollution, not the controller logic: an earlier test in the suite mutates the **shared `seller` role's capabilities** without restoring them. By the time this test executes, the `seller` role has lost `dokan_edit_product`, so:

* the single‑item route returns `dokan_rest_cannot_edit` instead of the expected `dokan_rest_cannot_view`, and
* the batch endpoint drops its `update` / `delete` sections,

which makes the assertions exercise a **missing capability** rather than the **per‑product ownership gate** they are meant to verify.

This PR re‑grants the canonical `seller` capabilities (sourced from `dokan_get_all_caps()`, exactly as the installer does) in the test's `set_up()`, so the test is resilient to suite‑wide capability pollution. **No production code is changed** — only the test's setup is hardened.

1. Run the full suite: `npm run phpunit` — `ProductControllerV3BatchAuthorizationTest` now passes (previously 5 failures + 1 error).
2. Confirm it still passes in isolation: `npm run phpunit -- --filter ProductControllerV3BatchAuthorizationTest`.

### Changelog entry

**Title:** Stabilize the v3 products batch‑authorization test

**Description:** The batch‑authorization test for the `dokan/v3/products` endpoint failed during the full PHPUnit run because a shared `seller` role capability was stripped by an earlier test. The test now re‑establishes the seller capabilities in setup, so it reliably validates the per‑product ownership gate without depending on suite execution order.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test cleanup so WordPress role changes do not carry over between tests.
  * This helps prevent inconsistent test results caused by leftover in-memory role settings.

### DIFF
--- a/tests/php/src/DokanTestCase.php
+++ b/tests/php/src/DokanTestCase.php
@@ -137,7 +137,15 @@ abstract class DokanTestCase extends WP_UnitTestCase {
     public function tear_down() {
         Monkey\tearDown();
 
-        parent::tear_down();
+        parent::tear_down(); // Rolls back the per-test DB transaction (incl. the wp_user_roles option).
+
+        // The DB option is restored by the rollback, but the cached WP_Roles singleton still holds any
+        // add_cap()/remove_cap() a test performed — those in-memory mutations would otherwise leak into
+        // every subsequent test in the run. Rebuild the singleton from the (now-canonical) option so
+        // role-capability changes cannot cross a test boundary. Mirrors the install-time reset in
+        // tests/php/bootstrap.php.
+        $GLOBALS['wp_roles'] = null; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+        wp_roles();
     }
 
 


MdAsifHossainNadim:
### Review feedback

Thanks for chasing this down — the diagnosis is correct and the test does become order-independent. The notes below are about **altitude** and **redundancy**, not about whether the result works. Example code included for each.

---

#### 1. The fix mutates the global `WP_Roles` singleton with no reset → it leaks seller caps *forward* into every later test (same bug class, opposite direction)

Each test runs inside a DB transaction that `WP_UnitTestCase::tear_down()` rolls back, **but the in-memory `WP_Roles` singleton is never rebuilt** — that is exactly why an earlier test's `remove_cap()` leaks into this one. `bootstrap.php` rebuilds it only **once at install** (`$GLOBALS['wp_roles'] = null; wp_roles();`), never per test. So this PR's `wp_roles()->add_cap( 'seller', … )` now leaks the *full* canonical cap set into all subsequent tests in the run.

The durable fix is to isolate the suite at the source — rebuild the singleton in the base `tear_down()` so role-cap changes can't survive a test boundary:

```php
// tests/php/src/DokanTestCase.php
public function tear_down() {
    Monkey\tearDown();

    parent::tear_down(); // rolls back the per-test DB transaction (incl. wp_user_roles)

    // The DB option is now restored, but the cached WP_Roles singleton still holds
    // any add_cap()/remove_cap() a test made. Rebuild it from the (canonical) option
    // so role-cap mutations cannot leak into the next test. Mirrors bootstrap.php.
    $GLOBALS['wp_roles'] = null; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
    wp_roles();
}
```

With this in place, **the entire cap-granting block in this test's `set_up()` can be deleted** — every test, not just this one, starts from the canonical role state, and the polluting test is neutralized too.

---

#### 2. The per-user `add_cap` loop is redundant with the role re-grant, and its comment is inaccurate

The comment says `SellerFactory` "snapshots the role's caps onto each user at creation (via `WP_User::set_role`)." `WP_User::set_role( 'seller' )` only stores the role assignment (`wp_capabilities = ['seller' => true]`); it does **not** copy the role's individual caps onto the user. Effective caps are recomputed from `wp_roles()` every time a `WP_User` is constructed — and every test method calls `wp_set_current_user( $this->seller_id1 )`, which builds a **fresh** `WP_User` that re-reads the just-restored role caps. So once the role is re-granted, this loop adds nothing the permission checks consume:

```php
// Redundant once the role is restored — the fresh WP_User from
// wp_set_current_user() already inherits these caps from the role.
foreach ( [ $this->seller_id1, $this->seller_id2 ] as $seller_id ) {
    $seller = new \WP_User( $seller_id );
    foreach ( $caps as $cap ) {
        $seller->add_cap( $cap );
    }
}
```

**Preferred:** adopt #1 and drop the whole block. **If** you'd rather keep a local guard in this test, at least remove the per-user loop and correct the comment:

```php
public function set_up() {
    parent::set_up();

    // An earlier test strips the shared `seller` role's dokan_* caps from the
    // in-memory WP_Roles singleton (the DB is rolled back per test, the singleton
    // is not). Re-grant the canonical caps so these assertions exercise the
    // per-product OWNERSHIP gate, not a missing capability.
    $roles = wp_roles();
    foreach ( dokan_get_all_caps() as $cap_group ) {
        foreach ( array_keys( (array) $cap_group ) as $cap ) {
            $roles->add_cap( 'seller', $cap );
        }
    }

    // ... existing route registration + product factory setup ...
}
```

> If the per-user loop was added because the role re-grant alone was observed to be insufficient, that would point at a different caching cause worth pinning down — the stated `set_role` rationale doesn't hold.

---

#### 3. Minor: per-test cap churn

`set_up()` runs before each of the 6 test methods. The role loop issues ~40 `add_cap()` → ~40 `update_option` writes, and the per-user loop ~40 caps × 2 users → ~80 `update_user_meta` writes — ~120 DB writes per test for setup alone. Both #1 (reset once in base teardown) and #2 (drop the per-user loop) eliminate the bulk of this.

---

**Summary**

- ✅ Diagnosis and end result are correct.
- 🔵 Move the real fix into `DokanTestCase::tear_down()` (rebuild `WP_Roles`) so the whole suite is isolated and this `set_up()` block can be removed.
- 🔵 Otherwise, drop the redundant per-user loop and fix the comment.

One caveat: I reasoned through the standard WP capability path rather than running the suite, so finding #2's redundancy is "very likely" rather than proven — but the #1 recommendation makes the block moot either way.

<sub>🤖 Claude Code-assisted review.</sub>

MdAsifHossainNadim (approved):
Approving — verified by running the **full PHPUnit suite** in wp-env on this branch (not just CI, which has no PHPUnit job):

- **With the fix:** `OK (245 tests, 953 assertions)`
- **With the `tear_down` reset reverted (baseline):** also `OK (245 tests, 953 assertions)`

The original failure did not reproduce in my environment even without the fix — consistent with order-dependent fixture pollution (`phpunit.xml` pins no `executionOrder`, so test discovery order varies by environment). The change is a **safe, monotonic improvement** to test isolation: it can only restore the canonical role state, never worsen it, and it introduces no regressions. 👍

**Optional, non-blocking follow-up** (fine to do separately): the reset rebuilds `WP_Roles` via `get_option( 'wp_user_roles' )` immediately after `ROLLBACK`, but WordPress flushes the object cache only in the *next* test's `set_up()` — so at this point the cache can still hold the polluting test's mutated option, and in the worst case (a polluter running directly before a role-sensitive test) the rebuild reads stale caps. Rebuilding after a cache flush — e.g. in `set_up()` after `parent::set_up()`, or `wp_cache_delete( 'alloptions', 'options' )` before the rebuild — would make isolation fully order-independent. Not required for merge.

PR #3313, opened 2026-07-09
feat: add pending badges for vendors and withdraw menu items

dev-shahed:
### All Submissions:

* [ ] My code is tested

### Related Pull Request(s)

* Full PR Link https://github.com/getdokan/dokan-pro/pull/5921

### Closes 

* Closes #

* Steps or issue link

### Changelog entry

*Title*

### Before Changes

### After Changes

### Feature Video (optional)

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

## Summary by CodeRabbit

* **New Features**
  * Admin menus now show live count badges for pending items, making it easier to spot outstanding vendor and withdrawal requests at a glance.
* **Style**
  * Improved badge readability in the Dokan admin menu, including consistent colors across hover, focus, and active states.

### DIFF
--- a/assets/src/less/global-admin.less
+++ b/assets/src/less/global-admin.less
@@ -414,3 +414,14 @@
     padding-bottom: 0.5rem;
     font-size: 13px;
 }
+
+/**
+ * Keep the Dokan admin menu pending-count badges readable in every state.
+ */
+#adminmenu #toplevel_page_dokan .wp-submenu .awaiting-mod,
+#adminmenu #toplevel_page_dokan .wp-submenu li a:hover .awaiting-mod,
+#adminmenu #toplevel_page_dokan .wp-submenu li a:focus .awaiting-mod,
+#adminmenu #toplevel_page_dokan .wp-submenu li.current a .awaiting-mod {
+    background-color: #3858e9;
+    color: #fff;
+}
--- a/includes/Admin/Dashboard/Pages/Vendors.php
+++ b/includes/Admin/Dashboard/Pages/Vendors.php
@@ -8,11 +8,10 @@ class Vendors extends AbstractPage {
     }
 
     public function menu( string $capability, string $position ): array {
-        $status_count  = dokan_get_seller_status_count();
-        $pending_count = absint( $status_count['inactive'] ?? 0 );
+        $pending_count = dokan_get_pending_vendor_count();
 
         $menu_title = $pending_count ? sprintf(
-            /* translators: %s: Inactive vendor count badge */
+            /* translators: %s: Pending vendor count badge */
             __( 'Vendors %s', 'dokan-lite' ),
             '<span class="awaiting-mod count-1"><span class="pending-count">'
             . number_format_i18n( $pending_count )
--- a/includes/Admin/Dashboard/Pages/Vendors.php
+++ b/includes/Admin/Dashboard/Pages/Vendors.php
@@ -8,9 +8,20 @@ class Vendors extends AbstractPage {
     }
 
     public function menu( string $capability, string $position ): array {
+        $status_count  = dokan_get_seller_status_count();
+        $pending_count = absint( $status_count['inactive'] ?? 0 );
+
+        $menu_title = $pending_count ? sprintf(
+            /* translators: %s: Inactive vendor count badge */
+            __( 'Vendors %s', 'dokan-lite' ),
+            '<span class="awaiting-mod count-1"><span class="pending-count">'
+            . number_format_i18n( $pending_count )
+            . '</span></span>'
+        ) : __( 'Vendors', 'dokan-lite' );
+
         return [
             'page_title' => __( 'Vendors', 'dokan-lite' ),
-            'menu_title' => __( 'Vendors', 'dokan-lite' ),
+            'menu_title' => $menu_title,
             'route'      => 'vendors',
             'capability' => $capability,
             'position'   => 100,
--- a/includes/Admin/Dashboard/Pages/Withdraw.php
+++ b/includes/Admin/Dashboard/Pages/Withdraw.php
@@ -19,9 +19,20 @@ class Withdraw extends AbstractPage {
      * @inheritDoc
      */
     public function menu( string $capability, string $position ): array {
+        $withdraw_count = dokan_get_withdraw_count();
+        $pending_count  = absint( $withdraw_count['pending'] ?? 0 );
+
+        $menu_title = $pending_count ? sprintf(
+            /* translators: %s: Pending withdraw request count badge */
+            __( 'Withdraw %s', 'dokan-lite' ),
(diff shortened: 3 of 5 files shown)


MdAsifHossainNadim (approved):
## Verdict: Approve ✅

Nothing blocks this on code. The badges are correct, the counts feeding them are cached, and the cache cannot go stale.

> **Disclosure:** the head commit (`99cce22d`) is a review fix I pushed to this branch, not @dev-shahed's work. Since I authored it, please give that commit in particular a second pair of eyes before merge — I've re-verified it below rather than trusting it, but I shouldn't be the only reviewer of my own code.

---

### What was fixed during review (`99cce22d`)

**1. The Vendors badge showed a count that could never be cleared.**

It read `dokan_get_seller_status_count()['inactive']`, which is derived as *"everything that is not approved"* — so it also counted users carrying no `dokan_enable_selling` meta at all. Every administrator qualifies, because `dokan_admin_user_register()` (`functions.php:1436`) only writes that meta for the `seller` role. The Vendors list matches on `dokan_enable_selling = 'no'` (`Vendor/Manager.php:70`, `=` compare), so it leaves those users out: the badge said `1`, the admin clicked it, the page showed nothing pending, and the badge stayed.

Replaced with a new `dokan_get_pending_vendor_count()` that delegates to the very query the listing is built from, so the two cannot drift apart again.

**2. The counts hit the database on every wp-admin page load.**

`dokan_get_seller_status_count()` runs two `WP_User_Query` objects (one joined on usermeta) with no cache. Menu registration happens on every admin request, so this ran everywhere.

The new helper caches in the shared `vendors` group. While verifying this I also found the withdraw side was worse than it looked: `Withdraws::query()` **read** from the object cache but never wrote to it (`Cache::set` was simply absent), so `dokan_get_withdraw_count()` was uncached in practice. Added the write for the status-count query only — a listing result can't be cached there because its total is read from `FOUND_ROWS()` immediately afterwards, which requires the query to have actually run.

---

### Verification (run against this branch on a live install)

| Check | Result |
|---|---|
| Plain new administrator, no `dokan_enable_selling` meta | badge **1 → 0**, list 0 ✅ |
| Same user set to `dokan_enable_selling = 'no'` | badge 1, list 1 ✅ |
| Three pending vendors (`number => 1` must not cap the total) | badge **3**, list **3** ✅ |
| Approve vendors via `make_active()`, no manual cache bust | badge **3 → 2 → 0** ✅ |
| Withdraw count query | cold 1 query, warm **0** ✅ |
| Withdraw **listing** totals across repeated runs | 46 / 46 — `FOUND_ROWS()` path untouched ✅ |

Cache invalidation was also checked structurally, since the helper uses Dokan's 2-week default TTL:

- `dokan_enable_selling` is written in exactly two places (`Vendor.php:1024` / `:1042`), both inside `make_active()` / `make_inactive()`. All three callers route through them — the REST endpoint (`StoreController:1033`), the admin profile screen (`UserProfile` → `vendor->activate()`), and registration. No direct writes anywhere in Dokan Pro.
- Every SQL write to the withdraw table emits a hook `WithdrawCache` listens to: `Manager::update_status` → `dokan_withdraw_status_updated`, `Withdraw::update` → `dokan_withdraw_updated`, plus delete and create. No raw writes in Pro.
- No Lite PHPUnit test touches the withdraw count path, so the new cache write can't flip an existing test.

PHPCS: `functions.php` reports identically before and after (19 pre-existing issues, 0 new).

---

### Please read before merging — not blocking

**The in-page Pending tab still shows the old inflated number.** `StoreController.php:559` fills the `X-Status-Pending` header from `dokan_get_seller_status_count()['inactive']`, and `vendors.tsx:188` renders it as the Pending tab. So on an affected site the menu badge now reads `0` while the tab it links to says "Pending 1" over an empty list.

The bug is pre-existing and independent of this PR — but before the fix those two numbers agreed (both wrong), and now they visibly disagree. It's a one-line change to point that header at `dokan_get_pending_vendor_count()`, and it would fix the legacy Vue admin at the same time. Happy either way, but I'd lean toward taking it here, since the badge is exactly what draws attention to that tab.

**One trade-off I chose deliberately.** `dokan_get_pending_vendor_count()` calls `dokan()->vendor->get_vendors()`, which writes `Manager::$total_users` on the shared singleton. Delegating to the list's own query is what stops the badge and the list drifting apart again, but it means a `get_vendors()` → `get_total()` pair straddling `admin_menu` would read my total instead of its own. Not reachable today (REST requests don't fire `admin_menu`) — just not free.

---

### Follow-ups (none blocking)

- **Shared badge helper.** This is now the 9th–11th copy of the same `<span class="awaiting-mod count-1">…` markup (Lite `Admin/Menu.php:44`; Pro `Admin/Admin.php:115`, `Refund/AdminMenu.php:31`, vendor-verification ×2, product-qa ×2, store-support ×2). Issue #5922 asks for badges "implemented consistently across the Dokan admin menus" — one `dokan_get_admin_menu_badge()` in Lite, with the existing call sites converted, would deliver that. Worth deciding explicitly rather than leaving implicit.
- **Dead code:** `Admin/Menu.php:39-46` builds `$withdraw_text` with badge markup and never uses it — line 51 registers plain `__( 'Withdraw' )`. So the legacy menu calls `dokan_get_withdraw_count()` on every admin load purely to discard the result.
- **Accessibility:** core's equivalent (`wp-admin/menu.php`, "Comments") wraps the number in `aria-hidden="true"` and adds a `screen-reader-text` span with an `_n()` string. Ours announces "Vendors 3" with no context.
- **`count-1` is hardcoded** where core emits `count-{$n}`.
- **Colour:** `#3858e9` diverges from core's `#d63638` used by the Comments/Plugins badges directly above it in the same menu, and it's applied unconditionally — non-default admin colour schemes (Midnight, Ocean, Sunrise) will clash.
- **Legacy menus get no badge.** `LegacySwitcher` keeps either the legacy or the new submenu entry per the `dokan_legacy_{title}_page` transient; the legacy entries in `Admin/Menu.php` are unbadged, so admins on that preference see nothing. Worth testing both ways.
- **HTML inside a translatable string:** `__( 'Vendors %s' )` lets a translator move `%s` to the front, which would break `LegacySwitcher`'s `explode( ' <', … )` slug parsing for that locale. Core has the same shape, so low risk — concatenating outside the translated string removes it entirely.

---

### For QA

The description is still the unfilled template — no test steps, changelog entry, or before/after screenshots. Since this came from a visual bug report (#5922, with a video), please add screenshots of the badge in at least the default and one dark admin colour scheme before handing it over.

**Reproducing the fixed bug:** add a second administrator through wp-admin (a fresh admin gets no `dokan_enable_selling` meta) and confirm the Vendors badge stays hidden, then set that user to a pending vendor and confirm it appears and clears on approval.

PR #3315, opened 2026-07-13
feat: extend product search to include SKUs

dev-shahed:
### All Submissions:

* [ ] My code is tested

### Related Pull Request(s)

* Full PR Link

### Closes 

* Closes #https://github.com/getdokan/dokan-pro/issues/5776

* Steps or issue link

### Changelog entry

*Title*

### Before Changes

### After Changes

### Feature Video (optional)

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved REST product search by matching terms via WooCommerce’s product search logic (including SKU matches), and applying results only within any existing listing filters.
  * Removed fallback behavior that previously used broader text search, reducing irrelevant products.
  * If the search term (e.g., SKU) doesn’t match any products, listings now correctly return zero results instead of the full catalog.


MdAsifHossainNadim:
## Review + reworked implementation

I audited this PR against issue #5776 and pushed a rework. **The good news first: this PR was pointing at the right class.** The new React vendor-dashboard Products list does call `GET /dokan/v1/products` → `WeDevs\Dokan\REST\ProductController`, so `prepare_objects_query()` is the correct place to fix. The bug also reproduces exactly as reported.

Three problems with the `posts_search` approach made me replace it rather than patch it:

**1. The `)))` anchor only matches the last search-term group.** `WP_Query::parse_search()` builds one parenthesised group *per term* and joins them with `AND`. Anchoring `preg_replace` to the trailing `)))` therefore ORs the SKU ids into the **final group only**. So `search=woo polo` still required "woo" in the title, and `search=shirt -blue` had its exclusion bypassed by SQL precedence.

**2. Variation SKUs never resolve to their parent.** A variation's `_sku` lives on a `product_variation` post, and the query ANDs `post_type = product`, so matching the variation's own id yields nothing. WP Admin *does* find the parent — which is the comparison the issue explicitly asks us to match.

**3. `SELECT post_id FROM wp_postmeta WHERE meta_key='_sku' AND meta_value LIKE '%…%'`** is an unindexed, unbounded, un-author-scoped scan, and every matching id site-wide gets inlined into an `IN()`.

### What I pushed instead

Delegate to `WC_Data_Store::load( 'product' )->search_products()` — literally the call WP Admin makes. It reads the indexed `wp_wc_product_meta_lookup.sku` column, tokenises multi-word terms correctly, resolves variation SKUs to parents, and supports numeric-ID search for free.

Placement is load-bearing: the block runs at the **end** of `prepare_objects_query()`, not next to `post_status`. The `on_sale` branch does `array_merge` into `post__in` (a *union*), so an early insert would have turned `?search=X&on_sale=true` into the entire on-sale set. Running last means the search set intersects. `post__not_in` is honoured by hand via `array_diff` because `WP_Query` ignores `post__not_in` once `post__in` is set, and an empty result falls back to `[ 0 ]` so a zero-match search can never return the whole catalogue.

### Verified locally (Lite 5.0.9 / Pro 5.0.7 / WC 10.9.4 / PHP 8.4)

Real `rest_do_request()` calls against `/dokan/v1/products`:

| case | before | after |
|---|---|---|
| product SKU `woo-vneck-tee` | 0 | 62 |
| **variation SKU** `woo-vneck-tee-red` | 0 | 62 (parent) |
| numeric id `62` | 0 | 62 |
| zero match | 0 | 0 (not the catalogue) |
| `search` + `exclude=62` | — | 0 |
| `search=woo` + `include=62` | — | 62 |
| `search` + `on_sale=true` | — | intersects, does not union |

Title search is **byte-identical** before and after across `V-Neck`, `Hoodie`, `shirt`, `Belt`, `woo` on both `/dokan/v1/products` and `/dokan/v2/products`. Vendor isolation confirmed: vendor 6 finds their own `dummy-vneck-tee`; vendor 1 searching the same SKU gets 0 results — `author` stays a separate top-level `AND`, so `post__in` can only narrow. PHPCS clean.

`/dokan/v2/products` inherits the fix (it does not override `$post_type`); the Pro `/dokan/v1/products/{id}/variations` endpoint is deliberately excluded by the `'product' === $this->post_type` guard and behaves exactly as before.

### Worth flagging for QA (pre-existing, not caused by this PR)

- **Status-tab counts are not search-aware.** `useProducts.ts` fetches `/products/summary` with no query args, so All/Published/Draft/Pending ignore the search term. Already true for title search today, but it becomes noticeable now that SKU search returns results.
- The **legacy** dashboard's `dokan_product_search_by_sku()` still uses `str_replace( ')))' … )` and carries the same multi-word and variation defects. Left alone deliberately to keep this PR focused — worth a separate ticket.
- Comma-separated multi-SKU search (`SKU-A,SKU-B`) works in legacy but not here; `search_products()` does not support it. Needs a product decision before designing for it.

MdAsifHossainNadim (approved):

PR #3277, opened 2026-06-22
Enhance: include auction products in the vendor product list counts

akzmoudud:
### All Submissions:

* [x] My code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)
* [x] My code satisfies feature requirements
* [x] My code is tested
* [x] My code passes the PHPCS tests
* [x] My code has proper inline documentation

### Changes proposed in this Pull Request:

The vendor product list summary (`get_product_summary`) hardcoded `[ 'booking', 'auction' ]` as excluded product types when building the status-tab counts. With the Auction module now integrated into the new product editor and surfaced in the product list, that hardcode left auction products **out of the tab counts** — so the list could show auctions while "All (N)" / status counts didn't include them.

This **removes `auction` from the hardcoded list** so its exclusion now comes from the `dokan_product_listing_exclude_type` filter — the same filter the list query (`prepare_objects_query`) uses — keeping list contents and tab counts in sync for auction. Pro modules that own a separate dashboard (booking, subscription, …) register their own types through that filter; auction simply stops opting itself out.

**`booking` is intentionally kept** in the hardcoded default (not a drift bug): booking products have always been a separate dashboard concept and must never inflate the regular product counts. The Booking module also adds `booking` through the filter, so the hardcode is redundant while that module is active and only matters in the edge case where booking products exist but the module (and thus its filter handler) is inactive.

```diff
-        // Match the list's exclusion scope so tab counts can't drift from it —
-        // Pro modules add `product_pack` etc. via this filter.
-        $exclude_types = array_values(
-            array_unique(
-                array_merge(
-                    [ 'booking', 'auction' ],
-                    (array) apply_filters( 'dokan_product_listing_exclude_type', [] )
-                )
-            )
-        );
+        // Exclusion scope for the status-tab counts. Module-owned types
+        // (auction, product_pack, …) opt out via the filter — the same filter
+        // the list query uses — so list and counts stay in sync. `booking` is
+        // kept as a deliberate defensive default (see inline comment).
+        $exclude_types = array_values(
+            array_unique(
+                array_merge(
+                    [ 'booking' ],
+                    (array) apply_filters( 'dokan_product_listing_exclude_type', [] )
+                )
+            )
+        );
```

### Related Pull Request(s)

* Dokan Pro (removes auction from the exclusion filter so it shows in the list): https://github.com/getdokan/dokan-pro/pull/5794

### Closes

* Refs getdokan/plugin-internal-tasks#1883

[...]

PR #3319, opened 2026-07-14
Enhance: request-driven include_types opt-in for the vendor product list

akzmoudud:
### All Submissions:

* [x] My code is tested

### Closes

* Closes getdokan/plugin-internal-tasks#2184

Auction products must appear in the (new) vendor dashboard product list, but nowhere else — after #3277, auctions leaked into the v2 product search behind the manual order "Add products to order" picker.

This PR makes the visibility **request-driven** instead of changing any existing filter signature:

* The Pro auction module excludes `auction` unconditionally via `dokan_product_listing_exclude_type` again (exact develop/legacy behavior), so every consumer of the shared filter — v2 search, legacy template, future consumers — is auction-free by default.
* A request may opt excluded types back in by sending `include_types`; the v1 products list (`prepare_objects_query`) and summary (`get_product_summary`) subtract it from the exclusion result via a new `get_requested_include_types()` helper. The values only ever shrink the exclusion list (never fed to the query), so unknown input is a harmless no-op. dokan-lite stays auction-agnostic.
* `useProducts` passes its query args through a new `dokan_product_list_query_args` JS filter, applied to **both** the list and summary fetches so the status-tab counts always match the list. The Pro auction module hooks it to attach `include_types: [ 'auction' ]`.
* The v2 product search rebuilds its own exclusion list and does not honor the opt-in, so the manual-order picker stays hard-excluded — even against hand-crafted requests.
* The legacy product page never runs this app, so legacy listings/counts stay auction-free with no editor-mode check.

### Related Pull Request(s)

* Base/parent (lite): #3277
* Companion (pro, must ship together): https://github.com/getdokan/dokan-pro/pull/5939 — restores unconditional exclusion, gates `remove_auctions` on the opt-in, and registers the JS filter callback.

1. Build assets in both plugins (`npm run build`), new product editor mode, vendor with auction products (some finished/closed).
2. Vendor dashboard → Products: auctions appear in the list (including finished ones) and in the status-tab counts.
3. Vendor dashboard → Orders → create manual order → "Add products to order": searching never returns auction products.
4. Legacy editor mode: legacy product listing and counts contain no auctions (unchanged legacy behavior).
5. `GET /dokan/v2/products?search=...&include_types[]=auction` (crafted): still returns no auctions.

### Changelog entry

*Enhance: show auction products in the vendor dashboard product list*

Previously auction products were hidden from the vendor product list (or, after #3277, leaked into the manual-order product search). Now the product list and its counts show auctions via an explicit `include_types` request opt-in, while every other product listing/search keeps excluding them.


akzmoudud:
Review follow-up: fixed the Major finding — `dokan_count_stock_posts` cache key now hashes the exclusion list (mirroring `dokan_count_posts`), so opt-in and plain summary calls no longer share a cache entry. Verified live: without-opt-in → 2 instock, with-opt-in → 9 instock, without again → 2 (no poisoning in either direction). The Minor booking/include_types ordering point is intentionally deferred for now.

akzmoudud:
closes in https://github.com/getdokan/dokan/pull/3327

PR #3324, opened 2026-07-17
fix(commission): keep admin and vendor shares summing to the order total

kzamanbd:
### All Submissions:

* [x] My code is tested

The admin's commission and the vendor's earning are both money, and both get rounded before they are paid out — but they were rounded **independently**, so the two shares stopped adding up to what the customer actually paid.

On an order of `6.45` at 30% commission (`6.45 × 30% = 1.935` — exactly half a cent):

| | Before | After |
|---|---|---|
| Admin commission | 1.93 | **1.94** |
| Vendor earning | 4.51 | 4.51 |
| **Sum** | **6.44** ❌ | **6.45** ✅ |

The missing cent then propagated. With a `0.49` gateway fee borne by the admin, Dokan recorded the admin's earning as `1.44` while the admin actually kept `1.45`.

`Calculator` now rounds **only** the admin's share to the currency's precision and gives the vendor the **exact remainder** — in both `calculate()` (sales) and `calculate_for_refund()` (refunds, where the proration is what generates the long decimal tails).

This also settles a disagreement between the two ways admin earning is derived, which is why the same order could report different figures in different places:

* `Commission.php:391` — `order_total - net_amount` (residual; used by reports)
* `OrderCommission.php:509` — `get_admin_net_commission() + get_total_admin_fees()` (recomputed; used by order details)

Since `net_amount` is now `order_total - sum(rounded admin shares)`, both derivations land on the same figure by construction.

#### Notes for reviewers

* **`Calculator::calculate()` now returns currency-precision values** where it previously returned raw floats. `test_calculator`'s expectation moved `67.792 → 67.79` to match.
* **`calculate_for_refund()`'s `$vendor_earning` parameter is now unused** — the vendor's share is derived from `$refund_amount` instead. Kept in the signature for backward compatibility. This is safe because `$vendor_earning + $admin_commission == $item_total` holds by construction in `calculate()`.
* **Two `OrderCommissionTest` refund expectations were corrected**, not merely bumped. They encoded the bug: recording a `5.63` refund to the vendor while dropping the vendor's balance by only `5.625`. They now use the rounded figures so the record and the balance move by the same amount.
* **Existing orders are not backfilled.** Commissions already stored under the old split keep their figures until `RegenerateOrderCommission` runs.
* Rounding is per line item, so on a multi-item order the admin can gain up to half a cent versus an exact split. Nothing is ever lost — the invariant holds per item and therefore per order.

### Related Pull Request(s)

* N/A

### Closes

* Closes getdokan/dokan-pro#5937 — cross-repo reference, so it will not auto-close from this PR.

**Manual**

1. Set a global commission of **percentage, 30%**.
2. Place an order with a single product totalling **6.45**.
3. Check the admin commission and vendor earning on the order.
   * Expected: admin **1.94**, vendor **4.51** — summing to exactly `6.45`.
   * Before this PR: admin `1.93`, vendor `4.51` — summing to `6.44`.
4. Using a gateway where the admin bears the processing fee (e.g. Stripe, fee `0.49`), confirm the admin's recorded earning is **1.45**, not `1.44`.
5. Partially refund an order and confirm the vendor's balance drops by exactly the amount shown on the refund record.

**Automated**

```bash
npx wp-env run tests-cli --env-cwd=wp-content/plugins/dokan-lite \
  php -d memory_limit=2G vendor/bin/phpunit --group commission
```

```
OK (24 tests, 188 assertions)
```

Full suite: `250 tests, 965 assertions`, with one failure — `OrderStatusChangeTest::test_blocked_status_transitions` — which reproduces identically on a clean `develop` and is unrelated to commission. (The suite needs `-d memory_limit=2G`; the default 512M OOMs in a REST test.)

`CommissionCalculatorTest` gains a data-driven test asserting the `admin + vendor == total` invariant, covering the issue's exact numbers, a half-cent that rounds up, a clean split, and a 100% commission edge case — plus a refund-sum test.

### Changelog entry

*Fix: Admin commission and vendor earning no longer lose a cent to rounding*

Admin commission and vendor earning were each rounded on their own, so whenever a commission landed on a half-cent the two shares no longer added up to the order total — an order of 6.45 at 30% paid out 1.93 + 4.51 = 6.44. The remaining cent was credited to neither party, and admin earnings, gateway fee deductions, and reports all drifted from the amounts actually transacted. The admin's share is now rounded to the store's currency precision and the vendor receives the exact remainder, so the two always reconcile against the order total.

### Before Changes

`6.45` at 30% with a `0.49` admin-borne gateway fee: admin commission shown as `1.93`, vendor earning `4.51` (sum `6.44`, a cent short of the `6.45` paid), admin earning recorded as `1.44` against the `1.45` actually received.

### After Changes

Same order: admin commission `1.94`, vendor earning `4.51` (sum `6.45`), admin earning `1.45`.

_No screenshots: this is a backend arithmetic fix with no UI surface of its own. The figures above are covered by the automated tests listed under "How to test"._

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved commission rounding to match the store’s configured currency precision.
  * Ensured admin and vendor commission shares consistently add up to the sale or refund total.
  * Corrected refund commission calculations, including orders with coupons and shared discounts.
* **Tests**
  * Added coverage for currency rounding edge cases and refund share accuracy.

### DIFF
--- a/includes/Commission/Calculator.php
+++ b/includes/Commission/Calculator.php
@@ -2,6 +2,7 @@
 
 namespace WeDevs\Dokan\Commission;
 
+use Automattic\WooCommerce\Utilities\NumberUtil;
 use WeDevs\Dokan\Commission\AbstractCommissionCalculator;
 use WeDevs\Dokan\Commission\Model\Commission;
 use WeDevs\Dokan\Commission\Model\Setting;
@@ -113,8 +114,10 @@ class Calculator extends AbstractCommissionCalculator {
     public function calculate(): Commission {
         $raw_admin_commission = $this->calculate_raw_admin_commission();
 
+        // Round only the admin's share and let the vendor take the exact remainder.
+        // Rounding both shares leaves a fraction of a cent credited to neither party.
         $net_amount       = $this->get_total();
-        $admin_commission = min( $raw_admin_commission, $net_amount );
+        $admin_commission = min( NumberUtil::round( $raw_admin_commission, wc_get_price_decimals() ), $net_amount );
 
         $vendor_earning = $net_amount - $admin_commission;
 
@@ -197,8 +200,10 @@ class Calculator extends AbstractCommissionCalculator {
 
         $refund_amount = abs( $refund_amount );
 
-		$vendor_earning_for_refund   = ( $vendor_earning / $item_total ) * $refund_amount;
-		$admin_commission_for_refund = ( $admin_commission / $item_total ) * $refund_amount;
+		// Split the refund the same way calculate() splits a sale, so the amounts
+		// refunded always add back up to the amount taken off the order.
+		$admin_commission_for_refund = NumberUtil::round( ( $admin_commission / $item_total ) * $refund_amount, wc_get_price_decimals() );
+		$vendor_earning_for_refund   = $refund_amount - $admin_commission_for_refund;
 
         $commission->set_vendor_net_earning( $vendor_earning_for_refund );
         $commission->set_admin_net_commission( $admin_commission_for_refund );
--- a/tests/php/src/Commission/CommissionCalculatorTest.php
+++ b/tests/php/src/Commission/CommissionCalculatorTest.php
@@ -36,7 +36,8 @@ class CommissionCalculatorTest extends DokanTestCase {
             ->set_settings( $settings )
             ->calculate();
 
-        $this->assertEquals( 67.792, $comm->get_admin_commission() );
+        // 67.792 before the commission split was rounded to the currency's precision.
+        $this->assertEquals( 67.79, $comm->get_admin_commission() );
     }
 
     public function test_calculate_for_refund() {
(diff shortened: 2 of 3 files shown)


MdAsifHossainNadim (approved):
## Verdict: Approve ✅

Reviewed the diff and independently verified the fix against the codebase (call-site tracing, storage-path audit, and executing the boundary arithmetic). Summary of what was checked:

### Correctness — verified

- **The invariant claim holds at every call site.** `calculate_for_refund()` now derives the vendor share as `refund − rounded admin share`, which is only safe when `vendor + admin == item_total` on entry. Both callers satisfy this: `OrderLineItemCommission::adjust_refunds()` passes values fresh from `calculate()`, and `OrderCommission::calculate_for_refund()` explicitly recalculates with refund-adjustment **off** (`OrderCommission.php:133–137`) before deriving refund shares. dokan-pro has **no** callers of `calculate_for_refund`, so the BC-kept unused `$vendor_earning` parameter is contained.
- **`min()` ordering is right.** `min( round(raw), net_amount )` caps *after* rounding, so a rounded-up commission can never exceed the item total and the vendor share can never go negative. Negative (admin-subsidy) commissions behave correctly through both paths.
- **The vendor's float-noise remainder is absorbed at every storage/display sink.** `dokan_orders.net_amount` is `decimal(19,4)` written via wpdb `%f`, vendor balance takes the same route, display goes through `wc_price()`, and the withdraw path already uses `NumberUtil::round`. Confirmed `0.15 − 0.08 = 0.06999…` serializes as `0.070000`.
- **Boundary arithmetic in the corrected test expectations checks out** (executed, not eyeballed): `round(-0.625, 2) = -0.63`, `round(-0.175, 2) = -0.18` (a genuine float-representation boundary — it holds), `round(6.45 × 0.30, 2) = 1.94`. The reworked `OrderCommissionTest` refund blocks are internally consistent (`25.87 = 31.5 − 5.63`; displayed `net_commission` 1.13 unchanged). The old expectations did encode the bug — refund record 5.63 vs. balance movement 5.625.
- **Conventions:** `NumberUtil` + `wc_get_price_decimals()` matches existing usage in `Withdraw.php:509` and `Vendor.php:908`.

### About the red CI

The failing **PHPCS Inspections** job is *not* a phpcs failure — the changed files yield only pre-existing `json_encode` warnings on untouched lines. The job's **PHPUnit step** fails with 2 failures in `OrderStatusChangeTest` (`pending → cancelled` staying `pending`, asserted at the WooCommerce parent-order level before any commission code runs). Every other open PR's run of this workflow from the same day is red with the same job — this is a develop-wide environmental issue, not introduced here, and should be fixed separately.

### Optional (non-blocking)

1. `calculate_for_refund()`'s docblock still describes `$vendor_earning` as "Original vendor earning" — worth a note that it's retained for BC and the vendor share is now derived from `$refund_amount` (with a `@since DOKAN_SINCE` change entry).
2. Make `half_cent_split_provider()` `static` to match `get_order_commission_data()` and be PHPUnit 10-ready.

Tight scope, honest test corrections rather than expectation-bumping, and the PR description matches what the code actually does. 👍

PR #3326, opened 2026-07-17
test: Playwright e2e coverage for the React vendor setup wizard

MdAsifHossainNadim:
## What

Playwright coverage for the React vendor setup wizard (Lite PR: https://github.com/getdokan/dokan/pull/3325).

### Related Pull Request(s)

* https://github.com/getdokan/dokan/pull/3325
* https://github.com/getdokan/dokan-pro/pull/5942

## Changes

```
tests/pw/tests/e2e/vendor-setup-wizard/
├── vendorSetupWizard.spec.ts       # happy path, edge cases, negative cases
└── vendorSetupWizardPage.ts        # page object + test data

tests/pw/utils/dbData.ts            # + appearanceSettings.vendor_setup_wizard = 'latest'
tests/pw/README.md                  # the React UI gate is three options now, not two
tests/pw/test-cases.md              # the wizard's cases
```

The React wizard is gated behind `dokan_appearance.vendor_setup_wizard`, so the env seed sets it and the spec opts in for its own run (restoring the previous value afterwards) — without that the legacy wizard renders and every assertion fails at mount.

## Coverage

| Case | Tags |
|---|---|
| Vendor sees the welcome screen | `@lite @exploratory @vendor` |
| Vendor completes the wizard end to end (store → payment → ready), asserting `dokan_profile_settings` after each save, the rail position, and that the step swap keeps one page load | `@lite @vendor` |
| Vendor skips from the welcome screen | `@lite @vendor` |
| Store step shows previously saved values on revisit | `@lite @vendor` |
| A half-filled bank form is accepted (payment can be finished from the dashboard) | `@lite @vendor` |
| A malformed payment email is rejected inline and keeps the vendor on the step | `@lite @vendor @negative` |
| A guest cannot reach the wizard from its URL | `@lite @guest @negative` |

## How to test

```bash
cd tests/pw
npx playwright test tests/e2e/vendor-setup-wizard --grep @lite
```

A fresh `docker:setup` seeds the appearance flags; on a partially-seeded env, run the snippet in `tests/pw/README.md` § *Required Dokan options*.


MdAsifHossainNadim:
## Review verdict: ✅ Approve — pending one green run

Reviewed at `6f1e5a0ab`. The specs are correct against the implementation as it stands and typecheck clean under the pw tsconfig; what's missing is an execution, not a change.

### Why this needed updating

The branch was written against an earlier revision of the wizard and would have failed wholesale for one reason: **the React wizard is gated behind `dokan_appearance['vendor_setup_wizard']`**, whose stored default is `legacy`. The env seed never set it, so the suite would have rendered the legacy wizard and failed at mount on every assertion.

That's fixed in two places, deliberately belt-and-braces:
- `dbData.appearanceSettings` (applied by `_env.setup.ts`) so a fresh `docker:setup` seeds it, and the README's *Required Dokan options* snippet now lists three flags instead of two;
- the spec opts in during `beforeAll` and restores the previous value in `afterAll`, so a partially-seeded env still runs.

### What else moved with the implementation

| Change | Why |
|---|---|
| Start Journey → `getByRole('button')` | it moves the SPA now; Skip stays a link because it leaves the wizard |
| Partial bank moved out of the negative block | the payment step is optional by design — a half-filled bank form saves, so the case now asserts the saved meta instead of an error |
| Malformed payment email added as the negative case | shape checks survive the relaxed required-field rules |
| Rail position + a page marker | pins the two SPA contracts: the rail tracks the step, and the whole walk stays one page load |
| `test-cases.md` | synced with the above |

The rail assertion is intentionally `Step N of \d+` rather than a fixed total, so it holds whether or not Pro's verification step is present.

### Coverage

Welcome screen · full walk (store → payment → ready) asserting `dokan_profile_settings` after each save · skip from welcome · store values prefilled on revisit · half-filled bank accepted · malformed email rejected inline · guest can't reach the wizard.

### The one gap

These have **not been executed** against an environment running the branch. Most of what they assert was driven by hand in a browser during review — Start Journey being a button, the `(Required)` badges, the rail text, a half-filled bank saving, the step swap keeping one page load — but two paths are reasoned rather than run: the malformed-email inline alert (the 400 and its field error are covered by a unit test; only the rendering is unproven) and the rail on a Pro-active env.

```bash
cd tests/pw && npx playwright test tests/e2e/vendor-setup-wizard --grep @lite
```

The red check is the same infrastructure failure as the parent PR: `npm i` can't reach the private `getdokan/dokan-ui` over SSH, which also fails on `develop`.

<sub>Review performed with Claude Code.</sub>

PR #3328, opened 2026-07-20
fix(intelligence): resolve stale AI model ids instead of fatalling on /ai/generate

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

`Gemini::get_default_model_id()` returned `gemini-2.0-flash`, an id that is **no longer among the registered Gemini models** (`gemini-2.5-flash`, `gemini-2.5-pro`, `gemini-2.5-flash-lite-preview-06-17`). `Provider::get_model()` returns `null` for an unknown id, and `handle_request()` passed that null straight into `method_exists()` — the exact `TypeError` in the report.

The reporter's second observation — *"some dropdowns, which were previously selected, appear empty after a few days"* — is **the same root cause, not a separate symptom**. That same unresolvable id is the `default` of the admin Model `<select>`, so it is not among the option keys and the select renders blank; saving the tab then persists `gemini-2.0-flash` into the `dokan_ai` option, which is what later reaches `/ai/generate`.

What changed:

1. **`Providers/Gemini.php`** — default now points at a registered model.
2. **`Services/Provider.php`** — new `resolve_model( $type, $model_id )` degrades a retired saved id to a working model of the right type instead of returning null, plus `get_default_model_id_by_type()`. Both are additive; the `get_default_model_id(): string` abstract signature is untouched so Pro's `BriaAi` keeps compiling.
3. **`REST/AIRequestController.php`** — returns `WP_Error` for every misconfiguration instead of dereferencing a null provider/model. Also fixes **a second, independent fatal one line earlier**: `$manager->get_provider()` returns null for an unregistered id, so a site still holding the 4.1 default `dokan_ai_engine = 'chatgpt'` hit `Call to a member function get_model() on null`.
4. **`Intelligence/Admin/Settings.php`** — type-aware default, plus a read-time remap of retired ids so the select self-heals.
5. **`src/admin/pages/Settings.vue`** — the hydration loop assigned the whole saved section *inside* the per-field loop, discarding defaults already applied to earlier fields. Hoisted the copy out of the loop.

Two deliberate calls worth reviewer attention:

- **Gated on `$manager->is_configured( $type )` rather than `empty( $provider_id )`.** `active_engine()` falls back to `openai`, so an emptiness check would have been dead code and an unconfigured site would have attempted a keyless upstream call.
- **`AIProviderInterface` was NOT widened.** `resolve_model()` / `get_default_model_id_by_type()` are not on the public interface, so every call site guards with `instanceof Provider` (or `method_exists` in Pro). Adding them to the interface would be a BC break for third-party providers.

Also enforces `dokan_ai_image_gen_availability` at the API — image generation is billed to the marketplace owner and the admin toggle previously held only in the UI.

### Related Pull Request(s)

* dokan-pro companion (image-model default): filed alongside — must ship in the **same release train**, though it is `method_exists`-guarded so merge order does not matter.

### Closes

* Closes https://github.com/getdokan/dokan-pro/issues/4910

Put the site into the reported configuration and call the endpoint:

```bash
wp eval '
$o = get_option( "dokan_ai", [] );
$o["dokan_ai_engine"]         = "gemini";
$o["dokan_ai_gemini_model"]   = "gemini-2.0-flash";  // retired id
$o["dokan_ai_gemini_api_key"] = "dummy";
update_option( "dokan_ai", $o );
wp_set_current_user( 1 );
$r = new WP_REST_Request( "POST", "/dokan/v1/ai/generate" );
$r->set_param( "prompt", "a short product title" );
$r->set_param( "payload", [ "field" => "title", "type" => "text" ] );
$res = rest_do_request( $r );
echo $res->get_status() . " " . wp_json_encode( $res->get_data() ) . "\n";'
```

**Before:** `PHP Fatal error: Uncaught TypeError: method_exists(): Argument #1 ... null given ... AIRequestController.php:97` with `method_exists(NULL, 'process_text')` — byte-identical to the trace in the issue.
**After:** no fatal; a clean JSON error response.

Verified matrix (all previously fatal or misleading, now clean — Lite 5.0.9 / Pro 5.0.7 / WC 10.9.4 / PHP 8.4):

| scenario | after |
|---|---|
| stale `gemini-2.0-flash` | 403 upstream error (model resolved, request attempted) |
| legacy engine `chatgpt` | 400 `dokan_ai_not_configured` (was a fatal on line 93) |
| unknown engine | 400 `dokan_ai_not_configured` |
| no API key | 400 `dokan_ai_not_configured` |
| `type=image`, toggle off | 400 "AI image generation is not enabled for this marketplace." |
| `payload.field`/`type` as arrays | 403, no `TypeError` |

Blank-dropdown half:

```
raw saved value      : gemini-2.0-flash
value seen by the UI : gemini-2.5-flash
select options       : gemini-2.5-flash, gemini-2.5-pro, gemini-2.5-flash-lite-preview-06-17
field default        : gemini-2.5-flash
default IS an option : YES   (before: NO -> blank dropdown)
other section intact : {"foo":"bar"}
```

**Browser check still needed from a reviewer** (my local session could not reach `/wp-admin`): Dokan → Settings → **AI Assist**, set Engine = Gemini, confirm the Model select pre-selects "Gemini 2.5 Flash" instead of rendering blank. `src/` changed, so run `npm run build` first — `assets/js/` is gitignored and no build artefacts are committed.

PHPCS clean on all four PHP files. `src/admin/pages/Settings.vue` has a pre-existing eslint parse error at 4:52, identical on `develop` — unrelated to this change; the script block passes `node --check`.

### Changelog entry

**Fix — AI Assist: a retired model id no longer crashes generation or blanks the settings dropdown**

Previously, when a provider retired a model that was already saved in settings (Gemini's default `gemini-2.0-flash`), the Model dropdown rendered blank and any vendor AI generation request crashed with a PHP fatal error. Dokan now falls back to a working model of the correct type, remaps retired ids when settings load, and returns a clear error message for every misconfiguration instead of failing fatally.

## Summary by CodeRabbit

- **New Features**
  - AI model selection now automatically falls back to a supported default when a saved/retired model isn’t available.
  - AI requests now resolve text vs. image compatible models more reliably.
- **Bug Fixes**
  - Updated the default Gemini model to **Gemini 2.5 Flash**.
  - Fixed AI settings loading to copy saved section values once and only apply field defaults when a value is missing.

### DIFF
--- a/includes/Intelligence/Admin/Settings.php
+++ b/includes/Intelligence/Admin/Settings.php
@@ -7,12 +7,53 @@ use WeDevs\Dokan\Intelligence\Manager;
 use WeDevs\Dokan\Intelligence\Services\ChatgptResponseService;
 use WeDevs\Dokan\Intelligence\Services\GeminiResponseService;
 use WeDevs\Dokan\Intelligence\Services\Model;
+use WeDevs\Dokan\Intelligence\Services\Provider;
 
 class Settings implements Hookable {
     public function register_hooks(): void {
         add_filter( 'dokan_settings_sections', [ $this, 'render_appearance_section' ] );
         add_filter( 'dokan_settings_fields', [ $this, 'render_ai_settings' ] );
         add_filter( 'dokan_rest_api_class_map', [ $this, 'rest_api_class_map' ] );
+        add_filter( 'dokan_get_settings_values', [ $this, 'remap_retired_model_values' ], 10, 2 );
+    }
+
+    /**
+     * Replace saved model ids that a provider has since retired.
+     *
+     * @since DOKAN_SINCE
+     *
+     * @param array  $values     Section setting values.
+     * @param string $section_id Section id.
+     *
+     * @return array
+     */
+    public function remap_retired_model_values( $values, $section_id ): array {
+        if ( 'dokan_ai' !== $section_id || ! is_array( $values ) ) {
+            return (array) $values;
+        }
+
+        $manager = dokan_get_container()->get( Manager::class );
+
+        foreach ( [ Model::SUPPORTS_TEXT, Model::SUPPORTS_IMAGE ] as $type ) {
+            $prefix = $manager->get_type_prefix( $type );
+
+            foreach ( $manager->get_engines( $type ) as $provider_id => $provider ) {
+                $key = 'dokan_ai_' . $prefix . $provider_id . '_model';
+
+                if ( ! $provider instanceof Provider || empty( $values[ $key ] ) || ! is_scalar( $values[ $key ] ) ) {
+                    continue;
+                }
+
+                $models = $provider->get_models_by_type( $type );
+
+                // A retired model id is not among the select options, so it would render blank.
+                if ( ! isset( $models[ $values[ $key ] ] ) ) {
+                    $values[ $key ] = $provider->get_default_model_id_by_type( $type );
+                }
+            }
+        }
+
+        return $values;
     }
 
     /**
@@ -90,7 +131,7 @@ class Settings implements Hookable {
                 'type'    => 'select',
                 'options' => array_map( fn( $model ) => $model->get_title(), $provider->get_models_by_type( Model::SUPPORTS_TEXT ) ),
                 'desc'    => __( 'More advanced models provide higher quality output but may cost more per generation.', 'dokan-lite' ),
-                'default' => $provider->get_default_model_id(),
+                'default' => $provider instanceof Provider ? $provider->get_default_model_id_by_type( Model::SUPPORTS_TEXT ) : '',
                 'is_lite' => true,
                 'show_if' => [
                     'dokan_ai_engine' => [
--- a/includes/Intelligence/Admin/Settings.php
+++ b/includes/Intelligence/Admin/Settings.php
@@ -25,11 +25,11 @@ class Settings implements Hookable {
      * @param array  $values     Section setting values.
      * @param string $section_id Section id.
      *
-     * @return array
+     * @return mixed Untouched when the section is not the AI section.
      */
-    public function remap_retired_model_values( $values, $section_id ): array {
+    public function remap_retired_model_values( $values, $section_id ) {
         if ( 'dokan_ai' !== $section_id || ! is_array( $values ) ) {
-            return (array) $values;
+            return $values;
         }
 
         $manager = dokan_get_container()->get( Manager::class );
--- a/includes/Intelligence/REST/AIRequestController.php
+++ b/includes/Intelligence/REST/AIRequestController.php
@@ -101,13 +101,12 @@ class AIRequestController extends DokanBaseVendorController {
             $prefix      = $manager->get_type_prefix( $type );
             $provider_id = $manager->active_engine( $type );
             $model_id    = dokan_get_option( 'dokan_ai_' . $prefix . $provider_id . '_model', 'dokan_ai', '' );
+            $model_id    = is_scalar( $model_id ) ? (string) $model_id : '';
             $provider    = $manager->get_provider( $provider_id );
 
             if ( ! $provider instanceof AIProviderInterface ) {
                 return $this->configuration_error( __( 'The configured AI provider is no longer available. Kindly reach out to Marketplace Owner.', 'dokan-lite' ) );
             }
-
-            $model_id = is_scalar( $model_id ) ? (string) $model_id : '';
 
             // A model id saved before the provider retired that model must not break generation.
             $model = $provider instanceof Provider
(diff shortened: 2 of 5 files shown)


MdAsifHossainNadim:
## CI note — the 2 PHPUnit failures are pre-existing on `develop`

The "Run PHPCS inspection" workflow is red, but the **PHPCS step itself passes** — the failing step inside it is `Running the phpunit tests`:

```
1) WeDevs\Dokan\Test\Orders\OrderStatusChangeTest::test_allowed_status_transitions
   with data set #4 ('pending', 'cancelled')
2) WeDevs\Dokan\Test\Orders\OrderStatusChangeTest::test_pending_to_any_status
   with data set #3 ('pending', 'cancelled')
   Failed asserting that two strings are equal.  -'cancelled'  +'pending'
Tests: 245, Assertions: 951, Failures: 2.
```

These are order-status transition tests and are unrelated to this PR. Confirmed pre-existing: the **identical two failures, same data sets, same assertion** appear on getdokan/dokan#3329, which changes only a CSS shim in `includes/ThemeSupport/Astra.php` and cannot possibly affect order status logic — and on #3315 as well.

Worth a separate ticket: `pending → cancelled` is not applying in the test environment.

kzamanbd (approved):
Approved.

Root-cause analysis holds up. Verified against the code:

- `gemini-2.5-flash` is registered (`Models/GeminiTwoDotFiveFlash.php:14`), so the default resolves again.
- The second fatal is real: `Manager::get_provider()` returns `null` for an unregistered id (`Manager.php:100-108`), and the old code called `->get_model()` on it — a site still holding `dokan_ai_engine = 'chatgpt'` hit that, not the model path.
- Gating on `is_configured( $type )` rather than `empty( $provider_id )` is the right call: `active_engine()` falls back to `openai`, so an emptiness check would have been dead code.
- Image toggle default `'off'` matches the two existing lite call sites (`REST/ProductControllerV3.php:431`, `REST/VendorDashboardController.php:385`) — consistent, no BC break.
- Pro stays compatible: `BriaAi extends Provider` so the `instanceof Provider` guards hold, pro image models set `generation_type` before `request()` so the new image-key check and the model's own key read agree, and pro's client reads `err.message` rather than the error code, so the new `dokan_ai_not_configured` code surfaces fine.
- CI: "Run PHPCS inspection" is red on the PHPUnit step, `OrderStatusChangeTest` (`pending` -> `cancelled`). Confirmed the identical two failures and data sets on #3329, which changes only a CSS shim. Pre-existing, not from this PR. The PHPCS step itself flags nothing in the changed files. The separate ticket suggested in the comment is warranted.

Pushed two follow-ups onto the branch in c6cad33c:

1. `resolve_model()` read `$models[ $model_id ]` straight off `get_models_by_type()`. That array comes from the `dokan_intelligence_provider_{id}_models` filter, and only `Model::enlist()` keys it by model id — a third party appending `$models[] = $model` produced an entry `get_model()` finds by `get_id()` and `resolve_model()` could not, so a live model id read as retired and was silently swapped for the default. Now re-keyed through `get_models_map()`, with `has_model_id()` so the settings remap stops repeating the assumption.
2. Resolution returned early, skipping the `dokan_intelligence_{provider}_provider_{model_id}_model` filter that `get_model()` applies, so third-party model overrides stopped reaching `/ai/generate`. Added `dokan_intelligence_{provider}_provider_resolved_model` as the resolution-path counterpart.

Plus `tests/php/src/Intelligence/ProviderTest.php` (11 tests): saved-id hit, retired id -> default, retired default -> first model, no model of that type -> null, foreign array key, `has_model_id()`, filter override, and the read-time remap. The one worth keeping an eye on is `test_every_text_provider_default_model_id_is_registered()` — it asserts across every registered text provider that its default model id is still registered, which is exactly the invariant that broke here. Retiring a model without moving the default now fails in CI instead of fatalling in production.

Two things still outstanding, neither blocking the merge:

- The blank-dropdown half is not fully fixed in Lite alone. Pro's image Model field still does `'default' => $provider->get_default_model_id()` (`dokan-pro/includes/Intelligence/Settings.php:95`), which is a text id for both registered image providers — `gpt-3.5-turbo` for OpenAI and now `gemini-2.5-flash` for Gemini. Neither is among the image options, so that select still renders blank on a site that never saved it; the read-time remap only heals a value that was already saved. Please link the companion PR and add `Dependency With Pro` so they ship in the same train.
- Screenshot still owed for the visual half, and `npm run build` is needed before checking it. Worth a QA pass on the settings page generally, not just AI Assist: the hydration fix means a field missing from the saved section now gets its declared default into the payload, so saving any tab persists defaults that were previously absent. I only found one place in Lite where a declared default disagrees with the fallback the reader uses (`store_map`, `'on'` in the field vs `'off'` in `Widgets/StoreLocation.php:49`) — out of scope here, filing separately — but the Pro sections are unaudited.

Note the tests are unexecuted locally (no Docker on my side), so CI is the first real run of them.

PR #3336, opened 2026-07-21
fix(dashboard): stop building vendor dashboard layout config on background requests

akzmoudud:
### All Submissions:

* [x] My code is tested

Closes: https://github.com/getdokan/dokan-pro/issues/5580

`FullWidthVendorLayout::register_vendor_dashboard_assets()` was hooked to `init`, which runs on **every** request — WP-Cron, Action Scheduler pings (every ~30s on any WooCommerce site), admin-ajax, REST and admin screens. On each of those it built the entire vendor dashboard layout config: vendor lookup, header nav, and a full `dokan_get_dashboard_nav()` pass (~50 navigation URL builds), then attached it as inline data to a script that is never enqueued outside the seller dashboard — so all of that work was discarded.

Measured on a plugin-heavy dev site: ~55 wasted `dokan_get_navigation_url()` calls per background request, ~150,000+ per day on an idle site. With WPML active each of those multiplies further (every URL translation re-enters the dashboard nav builder).

**The fix (9-line diff):**

1. Hook the registration to `wp_enqueue_scripts` (priority 5, before the enqueue callback at 10) instead of `init` — it only fires on front-end page renders, so background requests never reach the code at all.
2. Bail unless `is_user_logged_in() && dokan_is_seller_dashboard()` — the same condition the existing enqueue callback already uses. This check is unreliable on `init` (query not parsed yet) but valid on `wp_enqueue_scripts`, which is what makes the hook move necessary rather than just adding a guard.

Verified: zero calls from cron/AJAX/REST/admin/non-dashboard pages after the change; the dashboard page itself registers, localizes and enqueues exactly as before. Translated (WPML) dashboard pages keep working since dokan-wpml maps the page ID via the existing `dokan_get_current_page_id` filter.

### Related Pull Request(s)

* Companion (independent, same investigation) Pro PR: https://github.com/getdokan/dokan-pro/pull/5959

1. Enable the new vendor dashboard layout (`dokan_appearance.vendor_layout_style = latest`).
2. Add `error_log()` inside `dokan_get_navigation_url()` (or use Query Monitor's hook panel) and hit `wp-cron.php`, `admin-ajax.php` or any `/wp-json/` route: **before** this PR each hit logs dozens of navigation URL builds, **after** it logs none.
3. Open the vendor dashboard as a vendor: the React app, sidebar nav and header must load exactly as before (`vendorDashboardLayoutConfig` present in page source).

### Changelog entry

*Stop building vendor dashboard assets and layout config on background requests*

Previously the full-width vendor dashboard registered its assets and built its entire layout configuration on every request to the site, including WP-Cron, Action Scheduler, AJAX and REST API calls, where the result was always discarded. Now the work only happens on an actual logged-in seller dashboard page view, removing constant background overhead on every WooCommerce site running Dokan.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vendor dashboard asset loading to run only on the seller dashboard.
  * Prevented unnecessary dashboard scripts, styles, and configuration from loading on other pages.
  * Updated asset loading behavior for the latest vendor layout to improve reliability.

### DIFF
--- a/includes/Shortcodes/FullWidthVendorLayout.php
+++ b/includes/Shortcodes/FullWidthVendorLayout.php
@@ -30,7 +30,10 @@ class FullWidthVendorLayout implements Hookable {
         // Register vendor dashboard assets if the vendor layout is not legacy.
         $vendor_layout = dokan_get_option( 'vendor_layout_style', 'dokan_appearance', 'legacy' );
         if ( 'latest' === $vendor_layout ) {
-            add_action( 'init', [ $this, 'register_vendor_dashboard_assets' ], 99 );
+            // On wp_enqueue_scripts (not init): it only fires on front-end page
+            // renders — never for cron, Action Scheduler, AJAX, REST or admin —
+            // and the seller-dashboard check inside needs the parsed query.
+            add_action( 'wp_enqueue_scripts', [ $this, 'register_vendor_dashboard_assets' ], 5 );
             add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_vendor_dashboard_assets' ] );
             add_filter( 'template_include', [ $this, 'rewrite_vendor_dashboard_template' ] );
         }
@@ -94,6 +97,10 @@ class FullWidthVendorLayout implements Hookable {
      * @return void
      */
     public function register_vendor_dashboard_assets() {
+        if ( ! is_user_logged_in() || ! dokan_is_seller_dashboard() ) {
+            return;
+        }
+
         $admin_dashboard_file = DOKAN_DIR . '/assets/js/vendor-dashboard/layout/index.asset.php';
         if ( file_exists( $admin_dashboard_file ) ) {
             $dashboard_script = require $admin_dashboard_file;


kzamanbd (approved):
Approved.

## Verified

- **Hook move is required, not cosmetic.** `dokan_is_seller_dashboard()` resolves through `$wp_query->queried_object_id` (`includes/functions.php:206`), which is empty on `init` — so adding the guard without moving the hook would have disabled the dashboard entirely. Moving to `wp_enqueue_scripts` is what makes the guard valid.
- **Ordering is sound.** Priority 5 runs ahead of `enqueue_vendor_dashboard_assets` at 10, and the `dokan-react-components` style dependency still resolves because WordPress defers dependency resolution to print time rather than enqueue time.
- **Template path unaffected.** `template_include` fires before `wp_head()`, so the fullwidth template still swaps in.
- **No downstream breakage.** Grepped both plugins on current `develop`: the `dokan-vendor-dashboard` handle has no external enqueue, localize or `wp_script_is()` consumer. The only `dokan_vendor_dashboard_layout_config` consumer is Pro's subscription module (`modules/subscription/module.php:187`), which is hook-agnostic and works fine at the new timing.
- **Unadvertised bonus:** `dokan_get_dashboard_nav()` now runs after the query is parsed, so nav active-state is computed with real context instead of blind at `init:99`.

## CI

PHPCS green. PHPUnit (PHP 8.3) green across the full suite. PHPUnit (PHP 7.4) is red on infrastructure only — it failed in `Start WordPress environment` with Debian bullseye apt 404s (`libc6`, `dpkg-dev`) while building the wp-env container, so no test ever executed. Not attributable to this change.

The e2e suite has not been triggered on `0aec1d1c9`. Keeping `Needs: Testing` on for that reason — the dashboard render path this touches deserves a real run before merge.

## Non-blocking follow-ups

Neither gates this PR; both are worth a look.

1. **Duplicate guard.** There are now two callbacks on `wp_enqueue_scripts` carrying byte-identical conditions (lines 100 and 223). Consider folding registration into the enqueue callback so one guard owns the decision and the priority coupling disappears:

   ```php
   public function enqueue_vendor_dashboard_assets() {
       if ( ! is_user_logged_in() || ! dokan_is_seller_dashboard() ) {
           return;
       }

       $this->register_vendor_dashboard_assets();

       wp_enqueue_script( $this->script_key );
       wp_enqueue_style( $this->script_key );
   }
   ```

   If you prefer keeping both hooks, extracting the condition into a private `should_load()` would at least stop the two copies drifting apart.

2. **Test coverage.** A hook-timing regression like the original is invisible until someone profiles again. Asserting that `wp_script_is( $this->script_key, 'registered' )` stays false off-dashboard would pin this down cheaply.

## On the CodeRabbit finding

The 🔴 Critical on lines 196-197 is a false positive — please don't apply the suggested patch. `Manager::get()` is `return new Vendor( $vendor );` (`includes/Vendor/Manager.php:194`) and never returns a falsy value, so `$vendor->get_avatar()` cannot fatal. The `$vendor ? ... : $user_name` ternary on line 189 is dead defensive code.

There is a real bug nearby that it missed, pre-existing and outside this diff:

```php
'avatar' => $vendor->get_avatar() ?? VendorUtil::get_vendor_default_avatar_url(),
```

`get_avatar()` returns `wp_get_attachment_url()` (`Vendor.php:506`), which is `string|false`. `??` only catches `null`, so a vendor whose avatar attachment was deleted ships `"avatar": false` into the React layout config. Worth a separate issue with `?:` as the fix.

PR #3337, opened 2026-07-22
fix(security): enforce order ownership in shared bulk status helper (IDOR, completes L1)

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

The Patchstack report (getdokan/dokan-pro#5960, report `aa94c7b5`) named **two sinks** feeding the shared helper `dokan_apply_bulk_order_status_change()`:

1. the REST bulk-actions endpoint `POST /wp-json/dokan/v2/orders/bulk-actions` → `OrderControllerV2::process_orders_bulk_action()` (guarded earlier by #3291), and
2. the legacy vendor-dashboard bulk form → `includes/Order/Frontend/Hooks.php::bulk_order_status_change()`.

#3291 filtered order ids only in the REST caller, so the dashboard-form path still forwarded raw `$_POST['bulk_orders']` into the unguarded helper — a vendor could still change **another vendor's** order status.

This PR centralizes the ownership rule as a single reusable check, **`OrderUtil::current_user_can_manage_order()`** (`includes/Utilities/OrderUtil.php`): a vendor may act only on their own orders (`dokan_is_seller_has_order()`); admins/shop managers (`manage_woocommerce`) are exempt. It is called from **both** entry points:

* `dokan_apply_bulk_order_status_change()` — the shared sink, which covers the legacy dashboard form **and** the REST path; and
* `OrderControllerV2::process_orders_bulk_action()` — the REST caller, kept as defense-in-depth.

Both sinks named in the report are now guarded by one rule, and the duplicated inline closures from #3291 / the earlier commit are removed.

### Test scope

Two PHPUnit tests, **one per entry point named in the report**, so both callers of the shared guard are covered:

| Test | Path exercised | Asserts |
|---|---|---|
| `tests/php/src/Orders/BulkOrderStatusOwnershipTest.php` | Shared helper `dokan_apply_bulk_order_status_change()` — the sink the **legacy vendor-dashboard bulk form** funnels into | A vendor **cannot** change another vendor's order status; the vendor's **own** order still changes; an **admin** can change any order. (2 tests) |
| `tests/php/src/REST/OrderBulkActionsOwnershipTest.php` | The real REST route `POST /dokan/v2/orders/bulk-actions` → `process_orders_bulk_action()` (dispatched through the REST server as a vendor) | Response is 200; the vendor's **own** order becomes `completed`; **another vendor's** order is left unchanged. (1 test) |

* **Total: 3 tests, 6 assertions** — run with `npm run phpunit -- --filter 'BulkOrderStatusOwnershipTest|OrderBulkActionsOwnershipTest'`.
* **Coverage rationale:** the two tests hit the guard through both of its callers (dashboard form via the helper, and the REST endpoint), which are exactly the two sinks in the Patchstack report. The admin/shop-manager exemption is asserted at the helper level. The `dokan/v3` route needs no separate test — `OrderControllerV3` extends `OrderControllerV2` without overriding this method, so it inherits the same guarded code.
* **Not in scope here:** the download-grant/revoke and catalog-mode IDORs found in the same sweep are fixed in their own PRs (#3338, #3339, #3340).
* **Regression:** running the full `--group orders` suite shows only the two **pre-existing** `OrderStatusChangeTest` failures that are already red on `develop` (unrelated sub-order status-sync assertions); this PR adds none.

### Related Pull Request(s)

* Completes getdokan/dokan#3291 (audit finding L1); part of the getdokan/plugin-internal-tasks#1994 follow-up sweep.

### Closes

* Closes getdokan/plugin-internal-tasks#2148

1. As Vendor A, submit the vendor-dashboard bulk order-status form (or `POST /dokan/v2/orders/bulk-actions`) with `bulk_orders[]` / `order_ids[]` containing an order owned by Vendor B and `status=completed`.
2. Confirm Vendor B's order status is **unchanged**.
3. Confirm Vendor A can still bulk-update their own orders, and an administrator can update any order.
4. Automated: `npm run phpunit -- --filter 'BulkOrderStatusOwnershipTest|OrderBulkActionsOwnershipTest'` → passes (3 tests, 6 assertions).

### Changelog entry

**Fix — Cross-vendor order status change via the vendor-dashboard bulk form (IDOR)**

The vendor-dashboard bulk order-status form could change the status of orders belonging to other vendors, because the shared `dokan_apply_bulk_order_status_change()` helper trusted caller-supplied order ids. Order ownership is now enforced through a single `OrderUtil::current_user_can_manage_order()` guard used by both the REST bulk-actions endpoint and the dashboard form (admins/shop managers exempt), closing the second sink named in the Patchstack report and completing the #3291 fix.

### Before Changes

A vendor could POST another vendor's order ids to the dashboard bulk form and change those orders' statuses.

### After Changes

Foreign order ids are dropped at the shared guard; only the caller's own orders (or an admin's) are updated.

### DIFF
--- a/includes/Order/functions.php
+++ b/includes/Order/functions.php
@@ -958,16 +958,8 @@ function dokan_apply_bulk_order_status_change( $postdata ) {
     $status = sanitize_text_field( wp_unslash( $postdata['status'] ) );
     $orders = array_map( 'absint', $postdata['bulk_orders'] );
 
-    // Guard ownership at the shared sink so every caller (REST bulk-actions + the legacy dashboard bulk form) is covered: a vendor may only touch their own orders, admins/shop managers excepted.
-    if ( ! current_user_can( 'manage_woocommerce' ) ) {
-        $vendor_id = dokan_get_current_user_id();
-        $orders    = array_filter(
-            $orders,
-            function ( $order_id ) use ( $vendor_id ) {
-                return dokan_is_seller_has_order( $vendor_id, $order_id );
-            }
-        );
-    }
+    // Guard ownership at the shared sink so every caller (REST bulk-actions + the legacy dashboard bulk form) is covered.
+    $orders = array_filter( $orders, [ \WeDevs\Dokan\Utilities\OrderUtil::class, 'current_user_can_manage_order' ] );
 
     // -1 means bluk action option value
     $excluded_status = [ '-1', 'cancelled', 'refunded' ];
--- a/includes/Order/functions.php
+++ b/includes/Order/functions.php
@@ -958,6 +958,17 @@ function dokan_apply_bulk_order_status_change( $postdata ) {
     $status = sanitize_text_field( wp_unslash( $postdata['status'] ) );
     $orders = array_map( 'absint', $postdata['bulk_orders'] );
 
+    // Guard ownership at the shared sink so every caller (REST bulk-actions + the legacy dashboard bulk form) is covered: a vendor may only touch their own orders, admins/shop managers excepted.
+    if ( ! current_user_can( 'manage_woocommerce' ) ) {
+        $vendor_id = dokan_get_current_user_id();
+        $orders    = array_filter(
+            $orders,
+            function ( $order_id ) use ( $vendor_id ) {
+                return dokan_is_seller_has_order( $vendor_id, $order_id );
+            }
+        );
+    }
+
     // -1 means bluk action option value
     $excluded_status = [ '-1', 'cancelled', 'refunded' ];
 
--- a/includes/REST/OrderControllerV2.php
+++ b/includes/REST/OrderControllerV2.php
@@ -5,6 +5,7 @@ namespace WeDevs\Dokan\REST;
 use WC_Customer_Download;
 use WC_Data_Store;
 use WC_Product;
+use WeDevs\Dokan\Utilities\OrderUtil;
 use WP_Error;
 use WP_REST_Server;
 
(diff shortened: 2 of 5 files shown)


MdAsifHossainNadim:
## Verdict: ✅ Approve

### The exact case

A cross-vendor IDOR in the bulk order-status flow. The Patchstack report named **two sinks** feeding the shared helper `dokan_apply_bulk_order_status_change()`:

1. **REST**: `POST /dokan/v2/orders/bulk-actions` → `OrderControllerV2::process_orders_bulk_action()` — guarded earlier by #3291.
2. **Legacy vendor-dashboard bulk form** → `Order/Frontend/Hooks.php::bulk_order_status_change()` — **still open**. It verified the nonce, the `dokan_manage_order` cap, and the `order_status_change` setting, but never checked *whose* orders the submitted ids belonged to. Vendor A could POST `bulk_orders[]` containing Vendor B's order ids with `status=completed` and change Vendor B's order statuses.

### How it's handled

The ownership rule is centralized as `OrderUtil::current_user_can_manage_order()` — a vendor may act only on orders that `dokan_is_seller_has_order()` confirms are theirs (a prepared `LIMIT 1` lookup against `dokan_orders`); admins/shop managers (`manage_woocommerce`) are exempt. The guard is applied:

- **at the shared sink** `dokan_apply_bulk_order_status_change()` — so every caller (dashboard form + REST) is covered by one rule, and
- **in the REST caller** as defense-in-depth, replacing the duplicated inline closure from #3291.

Foreign order ids are silently dropped; the caller's own orders still update.

### Verified during review

- Grep across **Lite and Pro** confirms exactly two production callers of the helper — no third entry point left unguarded.
- Vendor staff keep working: `dokan_get_current_user_id()` resolves staff to their parent vendor id.
- Tests cover both entry points (helper directly + real REST dispatch) with positive, negative, and admin-exemption assertions — 3 tests / 6 assertions, green.
- The red **PHPCS inspection** job is its *phpunit step* failing on the two `OrderStatusChangeTest` cases (`pending → cancelled`) that are already failing on `develop`; they call `$order->update_status()` directly and never touch this code path. Not caused by this PR.

### Non-blocking nits (fine as follow-up)

- Add `@since DOKAN_SINCE` to the two new test class docblocks (CodeRabbit flagged the same).
- Add a one-line `@since DOKAN_SINCE` behavior note on `dokan_apply_bulk_order_status_change()`'s docblock — it's public API since 3.7.10 and now requires a logged-in user context (no-user/CLI callers will no-op).

Approving with **Dev Review Done** + **Needs: Testing** for QA to verify the dashboard bulk form and REST scenarios from the test instructions.

PR #3338, opened 2026-07-22
fix(security): enforce product ownership when granting order downloads via REST (IDOR, L7)

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

`grant_order_downloads()` — the callback for `POST /wp-json/dokan/v2/orders/<id>/downloads` (and `dokan/v3`, since `OrderControllerV3 extends OrderControllerV2` without overriding it) — is authorized only by `get_single_order_permissions_check()`, which verifies the caller owns the **order** in `<id>`. It then loops over the caller-supplied `ids[]` (product ids) and calls `wc_downloadable_file_permission()` for each with only an `if ( ! $product ) continue;` guard — **no product-ownership check**. A vendor placing an order they control can grant that order's customer working download permissions for **another vendor's** downloadable files.

This is the REST twin of **L4** (getdokan/plugin-internal-tasks#2002 / getdokan/dokan#3293), which only fixed the AJAX handler `grant_access_to_download()`.

The fix skips any product the caller does not own — `dokan_is_product_author()` — with `manage_woocommerce` exempt so admins/shop managers are unaffected, mirroring #3293.

### Related Pull Request(s)

* Mirrors getdokan/dokan#3293 (audit finding L4) for the REST path

### Closes

* Closes getdokan/plugin-internal-tasks#2149

1. As Vendor A (owning order `O`), `POST /wp-json/dokan/v2/orders/O/downloads` with `{"ids":[<Vendor B's downloadable product id>]}`.
2. Confirm **no** download permission is created for Vendor B's product.
3. Confirm granting Vendor A's **own** downloadable product on `O` still works.
4. Automated: `npm run phpunit -- --filter OrderDownloadsGrantOwnershipTest` → passes (1 test, 3 assertions).

### Changelog entry

**Fix — Cross-vendor download-permission granting via the REST orders/downloads endpoint (IDOR)**

Previously, a vendor could grant download access to another vendor's downloadable files through `POST dokan/v2|v3/orders/<id>/downloads`, because the endpoint checked only order ownership, not product ownership. The endpoint now grants only the caller's own products (admins/shop managers exempt), mirroring the earlier AJAX fix (#3293).

### Before Changes

`POST dokan/v2/orders/<own-order>/downloads` with another vendor's product ids granted that customer access to the victim's files.

### After Changes

Foreign product ids are skipped; only the caller's own products are granted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed bulk “grant downloads” behavior so vendors can’t grant download permissions for products owned by other vendors.
  * Store administrators and managers can still grant downloads across vendors.
  * Added eligibility checks to ensure only valid, permitted product downloads receive granted permissions.
* **Tests**
  * Added REST authorization coverage to verify cross-vendor protections on the order downloads grant endpoint, including admin/shop manager override behavior.

### DIFF
--- a/includes/REST/OrderControllerV2.php
+++ b/includes/REST/OrderControllerV2.php
@@ -281,7 +281,9 @@ class OrderControllerV2 extends OrderController {
 
         foreach ( $product_ids as $product_id ) {
             $product = dokan()->product->get( $product_id );
-            if ( ! $product ) {
+
+            // Only grant downloads for the vendor's own products, never another vendor's files (admins/shop managers exempt).
+            if ( ! $product || ( ! current_user_can( 'manage_woocommerce' ) && ! dokan_is_product_author( $product_id ) ) ) {
                 continue;
             }
 
--- a/tests/php/src/REST/OrderDownloadsGrantOwnershipTest.php
+++ b/tests/php/src/REST/OrderDownloadsGrantOwnershipTest.php
@@ -1,0 +1,123 @@
+<?php
+
+namespace WeDevs\Dokan\Test\REST;
+
+use Automattic\WooCommerce\Internal\ProductDownloads\ApprovedDirectories\Register as DownloadApprovedDirectories;
+use WeDevs\Dokan\REST\OrderControllerV2;
+use WeDevs\Dokan\Test\DokanTestCase;
+use WC_Data_Store;
+use WC_Product_Download;
+use WP_REST_Request;
+
+/**
+ * Authorization test for the REST download-grant endpoint.
+ *
+ * Covers the cross-vendor IDOR (audit L7 / plugin-internal-tasks#2149): the REST
+ * twin of L4 (getdokan/dokan#3293). A vendor granting downloads on their own order
+ * must not be able to grant access to another vendor's downloadable files.
+ *
+ * @group dokan-order-downloads
+ * @group dokan-authorization
+ * @group security
+ *
+ * @covers \WeDevs\Dokan\REST\OrderControllerV2::grant_order_downloads
+ */
+class OrderDownloadsGrantOwnershipTest extends DokanTestCase {
+
+    /**
+     * REST namespace for this controller.
+     *
+     * @var string
+     */
+    protected $namespace = 'dokan/v2';
+
+    /**
+     * Downloadable product owned by the caller (vendor A).
+     *
+     * @var int
+     */
+    protected int $own_product;
+
+    /**
+     * Downloadable product owned by another vendor (vendor B).
+     *
+     * @var int
+     */
+    protected int $foreign_product;
+
+    /**
+     * Order owned by vendor A.
+     *
+     * @var int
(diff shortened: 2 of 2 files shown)


MdAsifHossainNadim:
## ✅ Verdict: Approve

### What this fixes
`grant_order_downloads()` (`POST /dokan/v2|v3/orders/<id>/downloads`) was authorized **only** by order ownership (`get_single_order_permissions_check`), then looped over the caller-supplied `ids[]` and granted download permissions with just an `if ( ! $product )` guard — **no product-ownership check**. A vendor placing an order they control could grant that order's customer working download access to **another vendor's** downloadable files (cross-vendor IDOR / CWE-862 — the REST twin of L4 / #3293, which only patched the AJAX handler).

The fix adds a per-product ownership guard inside the loop:

```php
if ( ! $product || ( ! current_user_can( 'manage_woocommerce' ) && ! dokan_is_product_author( $product_id ) ) ) {
    continue;
}
```

### Why the fix is approved
- **Correct & complete** — closes exactly the gap described in the issue: foreign products are skipped, the caller's own products are still granted.
- **Right altitude / consistent** — mirrors the shipped AJAX fix (`Ajax.php`, #3293) and the sibling `process_orders_bulk_action()` pattern in this same controller. The `manage_woocommerce` exemption is *required* on the REST path (unlike AJAX) because `get_single_order_permissions_check` deliberately lets admins/shop-managers through — without the exemption they'd lose the ability to grant downloads for products they don't personally author. `dokan_is_product_author()` is the correct primitive; its `dokan_is_product_author` filter is the seam Pro uses to make staff/multi-user stores author-aware.
- **No regression** — the legitimate "grant your own product" flow is unchanged, and the order-ownership permission layer is untouched.

### Verification
- **Unit (full REST dispatch):** `phpunit --filter OrderDownloadsGrantOwnershipTest` → **OK (2 tests, 6 assertions)**, covering both branches — vendor **blocked** from a foreign product, admin/shop-manager **exempt**. Reverting the guard makes the vendor case **fail** (`Failed asserting that an array does not contain <foreign id>`), confirming the test is a real regression guard rather than a tautology.
- **Full REST-stack on live data** (`rest_do_request`, the pipeline WP core uses to serve a request): vendor on own order → `200`, own granted / **foreign blocked**; admin → `200`, foreign granted (exemption); vendor on *another vendor's* order → **`403`** (order-ownership layer intact).
- **PHPCS:** 0 errors on the changed files.

Test coverage + a test-isolation fix were added in `ea32b0662` (admin-exemption case; restore the shared `DownloadApprovedDirectories` mode in `tear_down()`; `@since` placeholder) — addressing the CodeRabbit notes.

> _Out of scope (already tracked):_ the sibling `revoke_order_downloads()` has the same family of gap — it deletes a caller-supplied `permission_id` without binding it to the order — tracked separately as **L8 / plugin-internal-tasks#2150 → #3339**.

PR #3339, opened 2026-07-22
fix(security): verify download permission belongs to order before revoking (IDOR, L8)

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

`revoke_order_downloads()` — the callback for `DELETE /wp-json/dokan/v2/orders/<id>/downloads` (and `dokan/v3`) — deleted a download permission straight from the caller-supplied `permission_id` (`WC_Data_Store::load('customer-download')->delete_by_id()`) with **no check that the permission belongs to the route order**. The route's permission callback only verifies order ownership, so a vendor owning any single order could enumerate small sequential `permission_id`s and delete **another vendor's / customer's** download permission (cross-vendor denial of download access).

The sibling `update_order_download()` in the same controller already guards this (`$download->get_order_id() !== $order_id`); this PR applies the same check to the revoke path.

### Related Pull Request(s)

* Related to getdokan/dokan#3338 (L7) — same endpoint group

### Closes

* Closes getdokan/plugin-internal-tasks#2150

1. As Vendor A (owning order `A`), create a download permission on Vendor B's order `B`.
2. `DELETE /wp-json/dokan/v2/orders/A/downloads` with `{"permission_id":<B's permission id>}`.
3. Confirm the response is **400** and Vendor B's permission still exists.
4. Confirm Vendor A can still revoke a permission that belongs to their own order `A`.
5. Automated: `npm run phpunit -- --filter OrderDownloadsRevokeOwnershipTest` → passes (2 tests, 4 assertions).

### Changelog entry

**Fix — Cross-vendor download-permission deletion via the REST orders/downloads endpoint (IDOR)**

Previously, `DELETE dokan/v2|v3/orders/<id>/downloads` deleted a download permission by id without confirming it belonged to that order, letting a vendor delete other vendors' download permissions. The endpoint now rejects a permission that does not belong to the (ownership-checked) route order.

### Before Changes

A vendor could delete any download permission by passing its `permission_id` to a `DELETE` on one of their own orders.

### After Changes

A permission whose `order_id` does not match the route order is rejected with a 400.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved download revocation validation by confirming the permission belongs to the specified order.
  - Invalid, unknown, or mismatched download permissions now return a clear error instead of being removed incorrectly.
  - Vendors can continue to revoke valid downloads associated with their own orders.

- **Tests**
  - Added coverage for authorized revocations, unauthorized cross-order attempts, and unknown permissions.

### DIFF
--- a/includes/REST/OrderControllerV2.php
+++ b/includes/REST/OrderControllerV2.php
@@ -390,8 +390,18 @@ class OrderControllerV2 extends OrderController {
     public function revoke_order_downloads( $requests ) {
         $download_id   = $requests->get_param( 'download_id' );
         $product_id    = $requests->get_param( 'product_id' );
-        $order_id      = $requests->get_param( 'id' );
+        $order_id      = absint( $requests->get_param( 'id' ) );
         $permission_id = $requests->get_param( 'permission_id' );
+
+        // Only revoke a permission that belongs to the ownership-checked order, so a foreign permission_id can't be deleted.
+        $download = new WC_Customer_Download( $permission_id );
+        if ( ! $download->get_id() || $download->get_order_id() !== $order_id ) {
+            return new WP_Error(
+                'dokan_rest_download_permission_invalid_order',
+                esc_html__( 'Download permission does not belong to this order.', 'dokan-lite' ),
+                [ 'status' => 400 ]
+            );
+        }
 
         try {
             $data_store = WC_Data_Store::load( 'customer-download' );
--- a/includes/REST/OrderControllerV2.php
+++ b/includes/REST/OrderControllerV2.php
@@ -391,11 +391,17 @@ class OrderControllerV2 extends OrderController {
         $download_id   = $requests->get_param( 'download_id' );
         $product_id    = $requests->get_param( 'product_id' );
         $order_id      = absint( $requests->get_param( 'id' ) );
-        $permission_id = $requests->get_param( 'permission_id' );
+        $permission_id = absint( $requests->get_param( 'permission_id' ) );
 
-        // Only revoke a permission that belongs to the ownership-checked order, so a foreign permission_id can't be deleted.
-        $download = new WC_Customer_Download( $permission_id );
-        if ( ! $download->get_id() || $download->get_order_id() !== $order_id ) {
+        // Only this order's own permission may be revoked; a foreign or unknown id is rejected (WC throws on an unknown id, caught here).
+        try {
+            $download         = new WC_Customer_Download( $permission_id );
+            $belongs_to_order = $download->get_id() && $download->get_order_id() === $order_id;
+        } catch ( \Exception $e ) {
+            $belongs_to_order = false;
+        }
+
+        if ( ! $belongs_to_order ) {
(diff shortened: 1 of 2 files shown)


MdAsifHossainNadim:
## ✅ Verdict: Approve

### What this fixes
`revoke_order_downloads()` (`DELETE /dokan/v2|v3/orders/<id>/downloads`) deleted a download permission straight from the caller-supplied `permission_id` with **no check that it belonged to the route order**. Since the route's permission callback only verifies order ownership, a vendor owning any single order could enumerate small sequential `permission_id`s and delete **another vendor's/customer's** download permission (cross-vendor denial of download access). The fix binds the permission to the ownership-checked order, mirroring the sibling `update_order_download()`.

### Hardening added during review (`6b2fc4d19`)
The original guard constructed `new WC_Customer_Download( $permission_id )` **unguarded**. `WC_Customer_Download` **throws `Invalid download.`** for any nonexistent *non-zero* id (id `0` is fine), so a bogus `permission_id` became an uncaught exception → **HTTP 500**, where the pre-fix code was a harmless no-op. So the fix as-authored regressed bad-input handling from 200 to a 500. Changes:
- Wrap the construction in `try/catch` → a foreign **or unknown** id now returns a clean **400** (never a 500).
- `absint()` the `permission_id`.
- Add `test_revoke_rejects_unknown_permission` — the original two tests only used *real* permission ids, so the throw path was never exercised.
- Test hygiene: restore the shared `DownloadApprovedDirectories` mode in `tear_down()`; add `@since DOKAN_SINCE`; `revoke_request()` now delegates to the base `delete_request()` helper.

### Why it's approved
- **Correct & complete** — the permission↔order binding is exactly the missing check, and it belongs in this data-layer callback (vendor/order ownership is already enforced by `get_single_order_permissions_check`). Collapsing not-found + wrong-order into one `400` also avoids a permission-existence enumeration oracle.
- **Robust** — no longer 500s on bad input.
- **No regression** — revoking a permission that belongs to the caller's own order still succeeds.

### Verification
- **Unit:** `phpunit --filter OrderDownloadsRevokeOwnershipTest` → **OK (3 tests, 5 assertions)**. Mutation-tested twice: removing the order-binding guard **fails** the foreign case (`200 ≠ 400`); reverting to the unguarded construction **errors** the unknown-id case (`Exception: Invalid download.`) — both are genuine regression guards, not tautologies.
- **Full REST stack** (`rest_do_request`) **and real over-the-wire HTTP** (authenticated admin `DELETE` on the live site): foreign → `400` (survives), unknown → `400`, own → `200` (revoked).

### Noted follow-up (out of scope here)
The sibling **`update_order_download()`** (`PUT /orders/<id>/downloads/<permission_id>`) constructs `WC_Customer_Download` the **same unguarded way**, so it 500-fatals on an unknown non-zero id (its `404` branch is only reachable for id `0`). Same latent robustness bug, different method/route — worth a small follow-up, ideally a shared `get_order_download_or_error( $permission_id, $order_id )` helper unifying both.

PR #3346, opened 2026-07-24
fix(settings): stop reading Dokan settings before init

kzamanbd:
### All Submissions:

* [x] My code is tested

Every request on this branch logs the WordPress 6.7 doing-it-wrong notice twice:

```
PHP Notice:  Function _load_textdomain_just_in_time was called incorrectly.
Translation loading for the dokan-lite domain was triggered too early. ...
```

A `doing_it_wrong_run` backtrace pins it to one chain:

```
FullWidthVendorLayout::register_hooks()   @ plugins_loaded  (dokan-class.php:278)
 → dokan_get_option()                     functions.php:1077
 → LegacySettingsRepository::all()
 → LegacySettingsBridge::hydrate_legacy_from_new() → build_map() → harvest_from_schema()
 → SettingsSchema::get_schema() → esc_html__( '…', 'dokan-lite' )   ← before init
```

Reading *any* Dokan setting walks the legacy settings bridge, and the bridge harvests its legacy-key mapping out of `SettingsSchema::get_schema()` — a ~2,800-line schema full of `esc_html__()` calls. Both call sites below run on `plugins_loaded`, but `load_plugin_textdomain()` is registered on `init` (`dokan-class.php:263`), so the translation calls land too early.

`BridgeBootstrap` already defers its own harvest to `init` priority 999 for exactly this reason; these two call sites had not been given the same treatment.

**1. `FullWidthVendorLayout` — the trigger.**

`register_hooks()` read `vendor_layout_style` on `plugins_loaded` to decide whether to register three hooks at all. The three hooks are now registered unconditionally and the setting is read inside each callback, behind a new `is_latest_layout()` helper. All three fire on `init` or later, so the read is safe there, and the repository memoizes the section snapshot so splitting one read into three costs nothing.

Externally observable behaviour is unchanged — a legacy-layout marketplace still gets no dashboard assets and no template override, the checks just moved one frame later.

**2. `LegacySettingsRepository::__construct()` — a second defect on the same path.**

The constructor looped over `known_sections()` to bind per-section `{add,update,delete}_option_{$section}` cache-flush listeners. Enumerating those section names requires the bridge mapping, so simply constructing the repository built the entire translated schema — on every front-end request that read a single option.

It was also wrong on its own terms: the map was harvested before Pro registers its `dokan_get_admin_settings_schema` filters on `init`, so Pro-only sections never got listeners and their snapshots could serve stale values after a write.

Invalidation now binds WordPress' generic `added_option` / `updated_option` / `deleted_option` hooks, all of which pass the changed option name as their first argument, and resolves membership against the snapshots actually held. No mapping is consulted, so no schema is built, so binding is safe at any point in the request. Writes to `dokan_admin_settings` flush every snapshot rather than one, since it is the overlay source for all sections. Snapshots exist only for sections that were read, so the listener is a no-op for every unrelated option.

Fix 2 alone does **not** silence the notice — with the constructor lazy, the schema build moves down one frame into `all()`. Both changes are required.

**Why not keep per-section hooks and defer them to `init`?**

That was tried on this branch in `cf703e771` and reverted in `5ecb822b7`. Deferring the binding leaves a repository first constructed *after* `init` has fired with no listeners at all, so it never invalidates and serves stale values for the rest of the request. Binding eagerly instead (guarding on `did_action( 'init' )`) reopens a worse hazard: a schema callback that itself calls `dokan_get_option()` re-enters the container *before* the shared instance is registered, so each nested call constructs another repository **and another bridge** — `LegacySettingsBridge::$building_map` is a per-instance latch and cannot see the outer build. Observed during development as:

```
PHP Fatal error:  Maximum call stack size of 8339456 bytes
(zend.max_allowed_stack_size - zend.reserved_stack_size) reached during compilation.
```

The generic hooks sidestep both because they need no mapping. The detour is not visible in this PR's diff; it is called out here only because the intermediate commit is in the branch history.

**Cost of the two strategies**, measured in the PHPUnit environment on Lite with 8 mapped sections (Pro raises both the section count and the schema size, which only widens the gap):

| | Generic (this PR) | Per-section (base) |
|---|---|---|
| `add_action` calls | 3 | 27 (`8×3 + 3`) |
| Schema build at bind time | — | 0.51 ms, 76 KB |
| Cost per option write | 0.048 µs | 0 |
| Memoized rebuild | — | 0.002 ms |

Break-even is roughly 10,600 option writes in a single request (`0.51 ms ÷ 0.048 µs`); a heavy request does a couple hundred, so the generic listeners cost about 10 µs against per-section's ~510 µs. Two caveats in the other direction: per-section's 0.51 ms is not always *extra*, because `hydrate_legacy_from_new()` builds the same map and any request that reads a Dokan setting pays it anyway; and per-section's zero per-write cost is real, since `do_action()` early-returns when a hook has no listeners. Neither changes the verdict at these magnitudes, and correctness decides it regardless — per-section cannot bind without building the schema.

### Related Pull Request(s)

* Depends on #3141 (`refactor/simplify-settings-to-flat-array`) — base branch.
* ⚠️ Overlaps #3336 (`fix(dashboard): stop building vendor dashboard layout config on background requests`), which edits the same `register_hooks()` lines but targets `develop`. That PR moves `register_vendor_dashboard_assets` from `init` to `wp_enqueue_scripts` priority 5 and keeps the pre-`init` `dokan_get_option()` read, so it does not address this notice. Whichever lands second will need a deliberate merge resolution — the two changes are compatible in intent (both want the work to happen later), they just touch the same block.

### Closes

* N/A — reported from `debug.log` on the branch.

1. Set `WP_DEBUG` and `WP_DEBUG_LOG` to `true`.
2. Check out this branch with Dokan Pro active and load any front-end page.
3. **Before:** `wp-content/debug.log` gains two `_load_textdomain_just_in_time` notices for the `dokan-lite` domain per request.
4. **After:** no `dokan-lite` notices. (Any remaining `_load_textdomain_just_in_time` traces come from `dokan-pro/includes/Module.php:169` on the `dokan` domain — a separate issue in that repo, normally hidden because Pro suppresses its own `doing_it_wrong_trigger_error` output.)
5. Confirm the vendor dashboard still behaves: with `vendor_layout_style = latest` the React layout and its assets load; with `legacy` the theme template is used and no dashboard assets are enqueued.
6. Save a setting from **Dokan → Settings** and confirm the new value is served immediately on the next read, both for a section the request had already read and after a write to `dokan_admin_settings`.

### Test coverage

`tests/php/src/Shortcodes/FullWidthVendorLayoutTest.php` — new file, three cases covering fix 1:

* `test_register_hooks_does_not_read_settings`
* `test_template_is_untouched_for_legacy_layout`
* `test_layout_setting_is_read_when_a_callback_runs`

Fix 2's invalidation behaviour is covered by the existing `LegacySettingsRepositoryTest` (20 cases), notably `test_foreign_legacy_write_flushes_only_that_section` and `test_new_flat_option_write_flushes_every_section`.

`--group admin-settings` at `memory_limit=3G`: **`OK (177 tests, 1279 assertions)`**. PHPCS clean on all changed files.

Two follow-ups, deliberately not in this PR:

* `test_constructor_does_not_build_the_settings_schema` and `test_write_to_unmapped_section_flushes_its_snapshot` were removed in `48fe86885` because they could not hold under the intermediate per-section design. They are valid again against the final code and should be restored.
* Pre-existing and unrelated: the `admin-settings` group peaks at 1.32 GB and fatals at the default 512 MB `memory_limit` (reproduced identically on the base branch, dying at the same test), and `npm run phpunit` passes `--env-cwd=wp-content/plugins/dokan` where the directory is `dokan-lite`, so it exits 127 with `chdir to cwd ... no such file or directory`.

### Changelog entry

*Fix: translations no longer load before `init`*

Reading a Dokan setting builds the admin settings schema through the legacy settings bridge, and that schema is translated. Two call sites did this before the `init` action, where the plugin's text domain is loaded — the vendor layout shortcode read its setting during hook registration on `plugins_loaded`, and the legacy settings repository built the whole schema in its constructor just to work out which options to watch. WordPress 6.7 and later flags this with a `_load_textdomain_just_in_time` notice, which was written to `debug.log` twice on every request, and the redundant schema build cost time on front-end requests that only needed a single option. Both call sites now defer the read until `init` or later, and cache invalidation no longer needs the schema at all.

### Before Changes

`debug.log`, two entries per request:

```
[24-Jul-2026 09:12:25 UTC] PHP Notice:  Function _load_textdomain_just_in_time was called incorrectly.
Translation loading for the dokan-lite domain was triggered too early. This is usually an indicator for
some code in the plugin or theme running too early. Translations should be loaded at the init action or
later. (This message was added in version 6.7.0.) in wp-includes/functions.php on line 6170
```

### After Changes

No `dokan-lite` entries — verified across repeated bootstraps on a Pro-active install.

### DIFF
--- a/includes/Admin/Settings/Repository/LegacySettingsRepository.php
+++ b/includes/Admin/Settings/Repository/LegacySettingsRepository.php
@@ -30,23 +30,31 @@ final class LegacySettingsRepository implements LegacySettingsRepositoryInterfac
     ) {
         $this->new_repo = $new_repo ?? new SettingsRepository();
         $this->bridge   = $bridge ?? new LegacySettingsBridge();
-        add_action( 'init', [ $this, 'init' ] );
-    }
-    public function init(): void {
 
-        foreach ( $this->known_sections() as $section ) {
-            add_action( "update_option_{$section}", [ $this, 'on_section_changed' ] );
-            add_action( "add_option_{$section}", [ $this, 'on_section_changed' ] );
-            add_action( "delete_option_{$section}", [ $this, 'on_section_changed' ] );
-        }
-
-        // The new flat option participates in every overlay — its writes invalidate
-        // every snapshot. Use named callbacks so the same listener isn't bound twice
-        // if the repository is instantiated more than once in a request.
-        $new_option = SettingsRepository::OPTION_KEY;
-        add_action( "update_option_{$new_option}", [ $this, 'flush_all_snapshots' ] );
-        add_action( "add_option_{$new_option}", [ $this, 'flush_all_snapshots' ] );
-        add_action( "delete_option_{$new_option}", [ $this, 'flush_all_snapshots' ] );
+        // Section invalidation uses WordPress' generic option hooks, all of
+        // which pass the changed option name as their first argument.
+        //
+        // Binding the per-section `{add,update,delete}_option_{$section}`
+        // variants instead would mean enumerating the bridge mapping here, and
+        // that builds the whole admin settings schema during construction.
+        // `dokan_get_option()` resolves this repository as early as
+        // `plugins_loaded`, so that would (a) run the schema's `esc_html__()`
+        // calls before `init` (WP 6.7+ `_load_textdomain_just_in_time` notice)
+        // and (b) recurse without bound: a schema callback that itself calls
+        // `dokan_get_option()` re-enters the container before the shared
+        // instance is registered, so every nested call builds *another*
+        // repository and bridge — the bridge's own re-entry latch is
+        // per-instance and cannot see the outer build.
+        //
+        // Deferring the binding to `init` is not a way out either: a repository
+        // first constructed after `init` has already fired would never bind at
+        // all and would serve stale snapshots for the rest of the request.
+        //
+        // Snapshots only exist for sections that were read, so `flush_cache()`
+        // is a no-op for every unrelated option.
+        add_action( 'added_option', [ $this, 'on_section_changed' ] );
+        add_action( 'updated_option', [ $this, 'on_section_changed' ] );
+        add_action( 'deleted_option', [ $this, 'on_section_changed' ] );
     }
 
     public function all( string $section ): array {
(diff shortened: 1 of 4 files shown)


kzamanbd:
Pro companion: getdokan/dokan-pro#5983 — same root cause, fixes the four Pro modules that read settings in their constructors. Merge this Lite PR first.

PR #3352, opened 2026-07-28
fix(security): require plugin capabilities, not manage_woocommerce, to install plugins (L11)

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

Four code paths that install wordpress.org plugins — and in three cases **activate** them too — were gated on `manage_woocommerce`. A **Shop Manager** natively holds `manage_woocommerce` but *not* `install_plugins`/`activate_plugins`, so that role could install plugins it cannot install through wp-admin, and install **and activate** them via the onboarding route. Installing and activating arbitrary code is a full-trust capability, so this is a privilege escalation to code execution.

The reporter named the two REST routes. A sweep of every plugin-install/activate primitive in the repo found **two more reachable instances of the same substitution**, so all four are fixed here:

| # | Location | Change | Note |
|---|---|---|---|
| 1 | `AdminExtensionsController` (`POST /dokan/v1/admin/extensions/install`) | `check_permission()` override requiring `install_plugins` | reported. This route only installs, never activates, so `activate_plugins` is deliberately **not** required |
| 2 | `AdminOnboardingController::install_required_plugins()` (`POST /dokan/v1/admin/onboarding`) | bails without `install_plugins` + `activate_plugins` | reported |
| 3 | `SetupWizard::install_plugin()` | guard changed from `manage_woocommerce` to `user_can_install_plugin()` | **new** — the shared sink behind #2 *and* the wizard's "Recommended" step |
| 4 | `SetupWizardNoWC::install_woocommerce()` | guarded via the inherited predicate | **new** — fixed slug (WooCommerce), so lower impact than #1–#3 |

`SetupWizard::user_can_install_plugin()` already returned `install_plugins`, but was only used to decide whether to *display* the Recommended step — never to authorize an install. It now also requires `activate_plugins` (the queued `WC_Install::background_installer` activates what it installs) and became `public` so the onboarding controller can reuse it rather than spelling the capability pair out a third time.

Two deliberate calls worth flagging for review:

* **The onboarding route itself stays on `manage_woocommerce`.** A Shop Manager legitimately completes onboarding, so only the install step is skipped — settings still save and the request still succeeds. The guard sits at the step rather than the route for that reason, and also stops `dokan_before_install_plugin` / `dokan_after_install_plugin` / `dokan_admin_onboarding_plugins_installed` from firing for a user who may not install.
* **`dokan_install_wp_org_plugin()` (`includes/functions.php`) is left without a capability check.** It is a documented general utility, and WP core's own `Plugin_Upgrader` likewise leaves authorization to the request layer; adding `current_user_can()` there would break any non-request caller (WP-CLI, cron). Its only production caller is #4, which this PR guards.

### Related Pull Request(s)

* Part of the getdokan/dokan-pro#5982 batch, alongside #3351 (L10) and the L12 fix
* Pro sibling finding filed as getdokan/plugin-internal-tasks#2175

### Closes

* Closes getdokan/plugin-internal-tasks#2173

As a **Shop Manager** (holds `manage_woocommerce`, lacks `install_plugins`), with a `wp_rest` nonce from any Dokan admin screen:

1. `POST /wp-json/dokan/v1/admin/extensions/install` with `{"slug":"classic-widgets"}`. **Before:** `200 {"success":true}` and the plugin is really on disk. **After:** `403 rest_forbidden`, nothing installed.
2. `POST /wp-json/dokan/v1/admin/onboarding` with `{"onboarding":true,"marketplace_goal":{…},"plugins":[{"id":"classic-editor","info":{"repo-slug":"classic-editor"}}]}`. **Before:** install queued (`woocommerce_setup_background_installing_classic-editor` set) and activated on shutdown. **After:** no install queued — but the response is still `200` and the onboarding settings are still saved.
3. As an **administrator**, confirm both still work: the extensions install succeeds and onboarding still queues its plugins.
4. Confirm the setup wizard's "Recommended" step still appears for an administrator.
5. Automated: `tests/php/src/REST/AdminPluginInstallCapabilityTest.php`.

> Note: PHPUnit could not be executed locally (Docker unavailable on this machine), so CI is the first run of the new test. Behaviour was verified end-to-end against a live install instead: pre-fix, a Shop Manager really did download `classic-widgets` to `wp-content/plugins/` through step 1 and queue the `classic-editor` install through step 2; post-fix both are blocked while the administrator paths and the onboarding save are unchanged. The test suppresses `shutdown` in `tear_down()` so the deferred installer can never reach wordpress.org during a test run.

### Changelog entry

**Fix — Shop Managers could install and activate arbitrary plugins through Dokan admin routes**

Previously, the Dokan extensions-install and onboarding REST routes, the setup wizard's plugin installer, and the no-WooCommerce wizard step all authorized plugin installation with the `manage_woocommerce` capability. A Shop Manager holds that capability without holding `install_plugins`/`activate_plugins`, so the role could install — and via onboarding install and activate — arbitrary WordPress.org plugins. These paths now require the WordPress plugin capabilities; onboarding itself remains available to Shop Managers, minus the plugin-install step.

### Before Changes

Shop Manager → `POST dokan/v1/admin/extensions/install {"slug":"classic-widgets"}` → `200`, plugin installed. Onboarding with a `plugins` array → installed **and activated**.

### After Changes

Shop Manager → `403 rest_forbidden`, nothing installed. Onboarding → `200`, settings saved, no plugin install queued. Administrators unaffected.

### Note for the reviewer

The second commit (`fix(cs)`) is unrelated to the vulnerability: `SetupWizard.php` had a pre-existing unsanitized `$_POST['reset_sub_category']` read, and Lite's PHPCS job scans whole changed files, so the branch could not go green without it. It is split into its own commit for that reason.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved commission settings so category reset preferences are saved consistently.
  - Prevented unauthorized users from installing plugins or extensions through setup, onboarding, and administration tools.
  - Added safeguards to stop installation actions when the required permissions are missing.

- **Security**
  - Plugin installation now requires both installation and activation permissions.
  - Unauthorized installation attempts are rejected with an appropriate permission error.

### DIFF
--- a/includes/Admin/SetupWizard.php
+++ b/includes/Admin/SetupWizard.php
@@ -937,9 +937,14 @@ class SetupWizard {
     }
 
     /**
-     * Should we show the WooCommerce Conversion Tracking install option?
+     * Whether the current user may install plugins through Dokan.
      *
-     * True only if the user can install plugins.
+     * Authorizes the wizard and onboarding install sinks, and gates the visibility of
+     * the 'Recommended' step. Both capabilities are required because the queued
+     * installer activates whatever it downloads.
+     *
+     * @since DOKAN_SINCE Made public for the onboarding controller, and added the
+     *                    `activate_plugins` requirement.
      *
      * @return boolean
      */
--- a/includes/Admin/SetupWizard.php
+++ b/includes/Admin/SetupWizard.php
@@ -943,8 +943,9 @@ class SetupWizard {
      *
      * @return boolean
      */
-    protected function user_can_install_plugin() {
-        return current_user_can( 'install_plugins' );
+    public function user_can_install_plugin() {
+        // The background installer activates what it installs, so both caps are required.
+        return current_user_can( 'install_plugins' ) && current_user_can( 'activate_plugins' );
     }
 
     protected function display_recommended_item( $item_info ) {
@@ -1009,7 +1010,8 @@ class SetupWizard {
      * @param array  $plugin_info Plugin info array containing name and repo-slug, and optionally file if different from [repo-slug].php.
      */
     public function install_plugin( $plugin_id, $plugin_info ) {
-        if ( ! current_user_can( 'manage_woocommerce' ) ) {
+        // Installing and activating arbitrary code is a full-trust action, so require the caps WordPress requires for it — not `manage_woocommerce`, which Shop Managers hold.
+        if ( ! $this->user_can_install_plugin() ) {
             return;
         }
 
--- a/includes/Admin/SetupWizard.php
+++ b/includes/Admin/SetupWizard.php
@@ -652,7 +652,7 @@ class SetupWizard {
         $options['admin_percentage']                          = $dokan_commission_percentage;
         $options['additional_fee']                            = isset( $_POST['dokan_commission_flat'] ) ? sanitize_text_field( wp_unslash( $_POST['dokan_commission_flat'] ) ) : 0;
         $options['commission_category_based_values']          = isset( $_POST['dokan_commission_category_based'] ) ? wc_clean( json_decode( sanitize_text_field( wp_unslash( $_POST['dokan_commission_category_based'] ) ), true ) ) : [];
-        $options['reset_sub_category_when_edit_all_category'] = isset( $_POST['reset_sub_category'] ) && false === dokan_string_to_bool( $_POST['reset_sub_category'] ) ? 'off' : 'on';
+        $options['reset_sub_category_when_edit_all_category'] = isset( $_POST['reset_sub_category'] ) && false === dokan_string_to_bool( sanitize_text_field( wp_unslash( $_POST['reset_sub_category'] ) ) ) ? 'off' : 'on';
 
         update_option( 'dokan_selling', $options );
 
--- a/includes/Admin/SetupWizardNoWC.php
+++ b/includes/Admin/SetupWizardNoWC.php
@@ -121,6 +121,15 @@ class SetupWizardNoWC extends DokanSetupWizard {
     public function install_woocommerce() {
         check_admin_referer( 'dokan-setup' );
 
+        // The wizard itself only requires `manage_woocommerce`, but installing and activating WooCommerce needs the plugin caps.
+        if ( ! $this->user_can_install_plugin() ) {
+            wp_die(
+                esc_html__( 'You do not have permission to install plugins.', 'dokan-lite' ),
+                esc_html__( 'Error installing WooCommerce plugin', 'dokan-lite' ),
+                [ 'response' => 403 ]
+            );
+        }
+
         require_once DOKAN_INC_DIR . '/functions.php';
 
         // Using output buffer to prevent outputting `trigger_error` in `plugins_api` function
--- a/includes/REST/AdminExtensionsController.php
+++ b/includes/REST/AdminExtensionsController.php
@@ -51,6 +51,20 @@ class AdminExtensionsController extends DokanBaseAdminController {
     }
 
     /**
+     * Check whether the current user may install plugins.
+     *
+     * Overrides the admin base check: this route installs code from wordpress.org, so
+     * `manage_woocommerce` is not enough — a Shop Manager holds it without `install_plugins`.
+     *
+     * @since DOKAN_SINCE
+     *
+     * @return bool
+     */
+    public function check_permission() {
+        return current_user_can( 'install_plugins' );
+    }
+
+    /**
      * Install a plugin from WordPress.org.
      *
      * @since SUSPENDED
--- a/includes/REST/AdminExtensionsController.php
+++ b/includes/REST/AdminExtensionsController.php
@@ -44,23 +44,28 @@ class AdminExtensionsController extends DokanBaseAdminController {
                             'sanitize_callback' => 'sanitize_text_field',
                         ],
                     ],
-                    'permission_callback' => [ $this, 'check_permission' ],
+                    'permission_callback' => [ $this, 'check_install_permission' ],
                 ],
             ]
         );
     }
 
     /**
-     * Check whether the current user may install plugins.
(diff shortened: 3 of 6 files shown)


MdAsifHossainNadim:
@coderabbitai Both test findings were valid and are fixed in 94c31c5bc; the Vue one I'm leaving out of scope.

**1. install-without-activate coverage — valid, fixed.** Correct that the `shop_manager` fixture lacks *both* caps, so the suite would have passed even if onboarding checked only `install_plugins`. Added `test_onboarding_does_not_install_for_user_who_cannot_activate()`, using a role with `install_plugins` granted and `activate_plugins` withheld. Verified live:

```
caps: install_plugins=y activate_plugins=n manage_woocommerce=y
onboarding -> 200 | install queued: no — blocked
```

**2. Queue-flag leak — valid, fixed.** `delete_option( 'woocommerce_setup_background_installing_classic-editor' )` is back in `tear_down()`. It had been dropped on the assumption that the per-test DB transaction covered it; you're right that it isn't worth relying on, since a leak makes the shop-manager assertion order-dependent through `install_plugin()`'s "already queued" early return.

`remove_all_actions( 'shutdown' )` stays deliberately: `install_plugin()` defers the real `WC_Install::background_installer` download to `shutdown`, and dropping it is what keeps a test run from hitting wordpress.org. WP restores `$wp_filter` per test anyway, so the blast radius is one test.

**3. `reset_sub_category` / `AdminCommission.vue` — out of scope, and I don't think it reproduces.** This PR's only change to that line is adding `wp_unslash()` + `sanitize_text_field()`, required because Lite's PHPCS job scans whole changed files (split into its own `fix(cs)` commit). On the substance: `AdminCommission.vue:82` uses `commission.reset_sub_category ?? true`, and `??` only substitutes on `null`/`undefined` — a stored `false` is preserved and round-trips through `dokan_string_to_bool()` to `'off'`. If you can point at a concrete reproduction I'll file it separately rather than fold a wizard UX change into a privilege-escalation fix.

akzmoudud:
## Code review — #3352

Reviewed the diff against the source and the linked audit (plugin-internal-tasks#2173), and did an independent completeness sweep rather than trusting the "found 4 sinks" claim. Short version: **the fix is correct, complete, and CI-verified — no correctness or security defects found.** The only thing standing between it and merge is a pre-existing, unrelated red check (details below).

### Verified, not taken on faith

| Check | Result |
|---|---|
| `SetupWizardNoWC`'s new `$this->user_can_install_plugin()` resolves (no fatal) | ✅ `SetupWizardNoWC extends DokanSetupWizard`, and `use …\SetupWizard as DokanSetupWizard` — real inheritance |
| `AdminExtensionsController::check_permission()` override doesn't over-tighten sibling routes | ✅ the controller registers **only** the install route |
| Method visibility bump to `public` is justified | ✅ `AdminOnboardingController` calls it cross-class, so `public` is required |
| **Completeness** — every plugin install/activate sink is now gated | ✅ swept all three primitives (`dokan_install_wp_org_plugin`, `WC_Install::background_installer`, `Plugin_Upgrader->install`); their only reachable callers are the four paths fixed here — no missed sink |
| Extensions route uses `install_plugins` only (it never activates) | ✅ confirmed no `activate_plugin()` on that path, so requiring `activate_plugins` there would be wrong |
| `@since DOKAN_SINCE` placeholder | ✅ canonical — `bin/version-replace.js:18` replaces it |
| New test suite runs green | ✅ CI executed 272 tests; the only failures are unrelated (below), so `AdminPluginInstallCapabilityTest` passed |

Test coverage is genuinely good: shop-manager blocked, admin allowed, onboarding installs nothing for a shop manager **but still saves and returns 200**, and the `install_plugins`-without-`activate_plugins` role is correctly rejected (the queued installer activates). Suppressing `shutdown` in `tear_down()` so the deferred installer can't reach wordpress.org is a nice touch.

### The red check is pre-existing and unrelated to this PR

The failing **"Run PHPCS inspection"** check is a bundled build + phpcs + phpunit job. PHPCS itself is clean on the five changed files. What fails is PHPUnit:

```
FAILURES! Tests: 272, Assertions: 1045, Failures: 2.
OrderStatusChangeTest::test_allowed_status_transitions  ('pending','cancelled')
OrderStatusChangeTest::test_pending_to_any_status       ('pending','cancelled')
```

`OrderStatusChangeTest` (`@group order-status`) has zero overlap with this PR's diff, and the same **"PHPCS Inspections" workflow is already red on `develop`**. So the branch inherited a baseline failure — it isn't introduced here. `mergeStateStatus` is `UNSTABLE` as a result.

Suggested: rebase on latest `develop` and re-run CI; the check should be green (or reveal that the develop-side test still needs its own fix) before merge.

### Minor / optional (non-blocking)

1. **`check_permission()` replaces the base gate instead of composing it.** It returns `current_user_can( 'install_plugins' )`, dropping the inherited `manage_woocommerce`. Practically fine since `install_plugins` is admin-only by default, but `return parent::check_permission() && current_user_can( 'install_plugins' )` reads as stricter least-privilege and avoids admitting a principal that somehow holds `install_plugins` without `manage_woocommerce`.
2. **`user_can_install_plugin()` as an instance method forces `new SetupWizard()` in the onboarding controller**, whose constructor registers `admin_menu`/`init`/`admin_init` hooks and instantiates `RecommendedPlugins`. The instantiation is pre-existing, but since an authz decision now depends on it, a `static` predicate (or a small standalone capability helper) would be lighter and side-effect-free.
3. **Description nit:** "passes PHPCS" is accurate for phpcs, but the check *named* PHPCS is red from the bundled unrelated phpunit failure — a one-line note would save the next reviewer the double-take.

### My take

Approvable on the merits — a tight, correct, completeness-verified security fix with strong negative-path tests. Not mergeable this second only because the required check is red from a pre-existing unrelated test; rebase + green CI and it's good to go. The three items above are optional.

Issue #2174, opened 2024-02-19
Issue with Stripe Express when the client doesn't provide 3D authentication


daniel-toader:
### Bug Description

The order does not go into processing when a customer proceeds as follows:

1.They create an order, enter the card details, but do not authenticate the payment through 3D Secure. Simply put, the customer does not complete the authentication (different from a failed authentication). After around 10 minutes the payment defaults to failed 3D authentication.

2.They return to the order (the order number remains unchanged), and this time they complete the 3D Secure authentication.
In this case, the order remains in a pending state. This step took place before the first payment intent defaulted to failed 3D authentication.

3.Manually changing the order status to processing completes the cycle, including the transfer to the vendor.

This combination of steps seems unlikely to happen; in our case, it's the first time in almost a year since we started using Stripe Express.

PR #3353, opened 2026-07-28
fix(security): reconcile reverse-withdrawal payments against the vendor's real debt (L12)

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

The reverse-withdrawal payment amount is client-supplied and was validated only as **greater than zero**. It then flowed verbatim through the cart, into the order line item, and on order completion into the vendor's reverse-withdrawal ledger as a `vendor_payment` — with no server-side comparison against what the vendor actually owes.

Paired with the built-in **Cash-on-Delivery** reverse-withdrawal gateway, where no money is actually collected, a vendor could mint arbitrary credit: wipe real commission debt for free and flip their due status off, which also suppresses the enforcement actions tied to being overdue. Requires the admin to have enabled Reverse Withdrawal (opt-in).

The regular withdraw flow already does this correctly and was the model for the fix: both `Withdraw\Hooks::ajax_handle_withdraw_request()` and `REST\WithdrawController::create_item()` run through `Withdraw\Manager::is_valid_approval_request()`, which rejects amounts exceeding `dokan_get_seller_balance()`. Reverse withdrawal had no equivalent.

**The fix goes at the shared sink.** The reporter named the REST route, but the same unvalidated amount reaches the same sink from the legacy AJAX handler (`wp_ajax_dokan_reverse_withdrawal_payment_to_cart`, via `$_POST['price']`). Guarding inside `Helper::add_payment_to_cart()` covers both callers at once, so no path can be reached without it — the same lesson as L1/L6, where a guard placed at one caller left the shared sink open.

Behaviour now:

| Amount | Result |
|---|---|
| more than the outstanding balance | rejected, `amount-exceeds-due-balance` |
| exactly the outstanding balance | accepted |
| part of the outstanding balance | accepted |
| anything, when nothing is owed | rejected, `no-due-balance` |
| zero or negative | rejected, `invalid-amount` (unchanged) |
| the balance is already covered by a payment order awaiting completion | rejected, `payment-already-awaiting`, naming the order with pay / cancel links |
| part of the balance is awaiting completion | only the remainder is payable |
| an order completes for more than is owed by then | credit clamped to the outstanding balance, order note added, never negative |

Two deliberate details:

* **The cap is the total outstanding `balance`, not `payable_amount`.** Under `by_month` billing `payable_amount` is only the currently-due slice, so capping there would block a vendor from paying a not-yet-due balance early. Commented in place, since the rest of this file uses `payable_amount` for due-status checks and the choice would otherwise look like a mistake.
* Comparison is done on scaled integers using `wc_get_price_decimals()`, so paying the exact outstanding balance is not rejected by floating-point drift — the same trick `is_valid_approval_request()` uses.

**In-flight orders (BUG-001, found in review).** The cap alone was not enough: the ledger is only credited when the payment order *completes*, so a vendor could place several individually valid payment orders while none had completed — the balance still read full each time — and over-credit once they all landed. Two layers close it, and either alone leaves a hole:

* `Helper::get_awaiting_payment_orders()` sums the vendor's payment orders that have not reached the ledger yet (`pending` / `processing` / `on-hold` / `failed`, skipping any already credited or settled) and deducts that from the due before the cap. Stacking is refused with `payment-already-awaiting`, and the message names the order(s) with a pay link and — for an abandoned order — a cancel link, so the vendor is never stuck behind an order only an admin can clear.
* `Order::insert_payment()` clamps the credit to the outstanding balance at completion, which is what actually closes the add-to-cart→checkout race and covers orders placed before this fix existed. A payment clamped to nothing is marked on the order (`_dokan_reverse_withdrawal_payment_settled`) so it is neither re-credited on a later completion nor held back as awaiting.

**Extensibility.** `dokan_reverse_withdrawal_payable_amount` filters the ceiling, applied at *both* ends — cart cap and completion clamp, told apart by `$context` — so whatever ceiling the cart accepts is what the ledger honours. Filtering only one end would take the vendor's money and withhold the credit.

**Also fixed on the way:** `Manager::all()` returned a bare row instead of a list for `vendor_transaction` whenever `per_page` was exactly 1 — a shape change that also hit `get_store_transactions()` for any REST client asking for one per page. It now returns a list at every page size (`balance` keeps its single-row collapse deliberately).

Also in this commit: the two new errors carry `[ 'status' => 400 ]`, and `invalid-amount` gained the same so the endpoint stops reporting a client error as `500`; and the price in the new message is stripped of markup because the AJAX caller renders the message as plain text through `dokan_sweetalert()` (`wc_price()` output would otherwise be shown as literal `<span>` markup to the vendor).

### Related Pull Request(s)

* Part of the getdokan/dokan-pro#5982 batch, alongside #3351 (L10) and #3352 (L11) — same embargoed report, no code overlap, independently mergeable in any order

### Closes

* Closes getdokan/plugin-internal-tasks#2174

1. Admin → enable **Reverse Withdrawal** (default Cash-on-Delivery gateway).
2. Give a vendor a real debt (e.g. a commission transaction of 100).
3. As that vendor: `POST /wp-json/dokan/v1/reverse-withdrawal/add-to-cart` with `{"amount":"999999"}`. **Before:** `{"status":true}` — check out with COD, have the admin complete the order, and `reverse-withdrawal/vendor-due-status` reports a large negative balance with due status off. **After:** `400 amount-exceeds-due-balance`, nothing added to the cart.
4. Confirm `101` is rejected, `100` (exact) is accepted, and `40` (partial) is accepted.
5. Repeat step 3 through the vendor dashboard's **Pay Now** button (the AJAX path) — it is rejected too, with a plain-text message.
6. As a vendor with **no** debt, any amount is rejected with `no-due-balance`.
7. **In-flight orders:** with a 100 debt, **Pay Now** 100 → COD checkout, leave the order uncompleted. **Pay Now** again → rejected `payment-already-awaiting`, naming the order. Complete the first order as admin → balance 0. **Before this fix** the second payment was accepted and completing both left the balance at −100.
8. **Completion clamp:** create two payment orders for the full debt while none has completed (two tabs), complete both → the second is credited nothing, gets an order note, and the balance is 0 rather than −100.
9. Automated: `tests/php/src/ReverseWithdrawal/PaymentAmountReconciliationTest.php` — 20 cases: the cap boundaries, the stacked-payment and remaining-balance cases, the literal two-order repro, the completion clamp, the settled-order marker (both directions), the credited-but-unfinished order, the filter at both ends, the cancel link, and the `Manager::all()` list shape.

> Verified locally under wp-env with HPOS on: `PaymentAmountReconciliationTest` OK (20 tests, 58 assertions); full suite 287 tests, 1 failure — the pre-existing `OrderStatusChangeTest ('processing','cancelled')` red that also fails on `develop`. PHPCS clean on all changed files. Also verified end-to-end on a live install (both the legacy dashboard and the new React reverse-withdrawal page post to the same AJAX sink, so both are covered).

### Changelog entry

**Fix — Vendors could clear reverse-withdrawal debt with an arbitrary payment amount**

Previously, the reverse-withdrawal payment amount was taken from the request and only checked to be greater than zero, so a vendor could submit any figure and have it credited to their ledger — clearing real commission debt for free when the Cash-on-Delivery reverse-withdrawal gateway is in use. Payments are now reconciled server-side against the vendor's actual outstanding balance: amounts above it are rejected, payment is refused when nothing is owed, payment orders still awaiting completion are deducted so several cannot be stacked, and the ledger clamps the credit at completion so it can never go negative. Exact and partial payments, including early payment of a not-yet-due balance, work as before. Vendor staff can no longer reach the REST add-to-cart route (they were already blocked on the dashboard path). Adds the `dokan_reverse_withdrawal_payable_amount` filter.

### Before Changes

`POST dokan/v1/reverse-withdrawal/add-to-cart {"amount":"999999"}` → `{"status":true}` → after COD checkout and order completion, `vendor-due-status` → `{"status":false,"balance":{"balance":-999799,…}}`.

### After Changes

`POST dokan/v1/reverse-withdrawal/add-to-cart {"amount":"999999"}` → `400 amount-exceeds-due-balance` ("Payment can not be greater than your due balance of $100.00").

### Behaviour change worth calling out

Vendor **staff** can no longer reach `POST /dokan/v1/reverse-withdrawal/add-to-cart`. The route previously used the shared read permission check, and `dokan_is_user_seller( dokan_get_current_user_id() )` resolves staff to their parent vendor, so staff passed — while the resulting order is created under the staff member's own customer id, which credits the wrong ledger. The route now gates on `dokandar`, matching the AJAX caller (`ReverseWithdrawal\Ajax`), which has always blocked staff. The vendor dashboard UI is unaffected, since it never offered this to staff.


MdAsifHossainNadim:
@coderabbitai review

MdAsifHossainNadim:
@coderabbitai full review

MdAsifHossainNadim:
@coderabbitai full review

MdAsifHossainNadim:
## Verdict: Approve ✅

Re-reviewed at `5ed040744`. The test blocker is cleared and the security fix holds up. One caveat below that should be recorded rather than discovered later.

### The fix

The reverse-withdrawal payment amount was client-supplied and validated only as `> 0`, then flowed verbatim into the vendor's ledger as a `vendor_payment` on order completion. With the Cash-on-Delivery gateway (no money collected), a vendor could mint arbitrary credit and wipe real commission debt.

The guard now lives in `Helper::add_payment_to_cart()` — the shared sink — so both callers (`Ajax.php:47` and `ReverseWithdrawalController.php:354`) are covered at once, and the amount is reconciled against the vendor's actual outstanding balance.

### Verified

- **Tests green.** `OK (6 tests, 12 assertions)` run locally on this commit. Worth noting the assertion count went 10 → 12 with the fixture fix: the two acceptance tests previously died before reaching `cart_payment_amount()`, so the checks on the value that actually lands in the ledger were dead code. They now execute.
- **The fixture fix is subtly right.** `create_reverse_withdrawal_base_product()` bails on `! current_user_can( 'manage_options' )` (`InstallerHelper.php:61-63`), so it has to run as admin before switching to the vendor. The commit does exactly that.
- **Sign convention correct** — `get_store_balance()` returns `debit - credit` (`Manager.php:381`), so positive means the vendor owes. `$due <= 0` is the right "nothing owed" test.
- **Balance is server-derived** — `get_vendor_balance()` defaults to `dokan_get_current_user_id()` (`Helper.php:422-424`), never a request parameter.
- **The obvious bypass is already closed.** I went looking for quantity multiplication: the base product is `sold_individually` (`InstallerHelper.php:100`, re-asserted at `Helper.php:203-206`), *and* the ledger credit is read from the `_dokan_reverse_withdrawal_balance` line-item meta (`Order.php:136` → `Helper.php:181`) rather than quantity × price. A tampered quantity still couldn't inflate the credit.
- `empty_cart()` before adding prevents multi-item accumulation; `is_payment_inserted()` prevents one order crediting twice.

CI's red is stale — the last run was on `26131aabd` and died building the Docker image (`composer global require phpunit` hit an HTTP/2 504), so the suite never executed there. Nothing has run on `5ed040744` yet; the local run above is the verification.

### One caveat to record: L12 is mitigated, not eliminated

The cap reads the ledger, but the ledger is only credited at **order completion** (`Order.php:35,38` → `insert_payment()` at `:88`). Nothing subtracts reverse-withdrawal orders that are already placed and unpaid — there's no pending-order guard and no checkout-time revalidation.

Reproduced: with a 100 debt, `add_payment_to_cart( 100 )` succeeds **twice in a row**, because the second call still sees the full balance outstanding.

So the attack is rate-limited rather than closed. Instead of one silent request minting 999,999, a vendor must place many separate orders and wait for the admin to complete each — conspicuous, and visible in the order list. That is a large, real improvement and exactly what #2174 asked for, but the changelog currently reads as though the class is shut. Suggest a line saying so, plus a follow-up ticket to subtract the reverse-withdrawal amounts on the vendor's unpaid orders from the balance before comparing.

### Follow-ups (non-blocking)

1. **The scaled-integer idiom is inconsistent across the codebase.** This PR correctly uses `10 ** wc_get_price_decimals()`, but `Withdraw/Hooks.php:114-115` and `Withdraw/Manager.php:45-46` hardcode `* 100`. Those are wrong for 3-decimal currencies (KWD, BHD, TND, OMR) — they round the balance to 2 dp before comparing, so a vendor can be blocked from withdrawing their last fraction. Worth a shared `dokan_money_to_minor_units()` helper and using it in all four places. (WooCommerce's `wc_add_number_precision()` is close but not a drop-in — it rounds to 4 extra decimals rather than to a whole integer, returns a float, and its precision is filterable via `woocommerce_internal_rounding_precision`, which isn't something a security comparison should depend on.)
2. **Dead validation on the route.** `amount` is declared `'type' => 'string'` with `'minimum' => 1` (`ReverseWithdrawalController.php:131-138`); `minimum` is a numeric JSON-Schema keyword and is a no-op against a string type. Harmless now the sink guards, but it reads like protection.
3. **The float-drift rationale is untested** — no case uses a fractional `due`, so the scaled comparison the comment justifies is never exercised.
4. **QA note on a deliberate behaviour change:** a zero-balance vendor submitting any amount now gets `no-due-balance` rather than building pre-paid credit. It's in the changelog, but any UI that lets a zero-balance vendor open the pay form will now surface an error.

dev-shahed:
# BUG-001 — Reverse withdrawal: vendor can still over-credit their ledger via sequential in-flight orders

PR #3353 reconciles the reverse-withdrawal payment amount against the vendor's real
debt, but only at **cart-add time**. The ledger credit is written later, at order
completion, straight from the order line-item meta and **without re-checking the
balance**. A vendor can therefore create several payment orders that are each valid
when created — because none has completed yet, so the balance still reads full — and
over-credit their ledger once all of them complete.

The reported attack (a single 999999 payment) *is* fixed. The underlying primitive —
minting credit for free through the Cash-on-Delivery gateway — is not.

## Steps to reproduce (manual)

1. Admin → enable **Reverse Withdrawal**, with the Cash-on-Delivery gateway enabled.
2. Seed a vendor with a real debt of **100** (one completed order producing 100 admin commission).
3. As that vendor: dashboard → Reverse Withdrawal → **Pay Now** `100` → check out with COD.
   Order 1 is created. **Leave it uncompleted.**
4. Check `GET /dokan/v1/reverse-withdrawal/vendor-due-status` → balance still reads **100**
   (nothing has completed, so nothing has been credited).
5. **Pay Now `100` again.** This passes the new guard → check out. Order 2 is created.
6. Admin → WooCommerce → Orders → mark **both** orders completed.
7. `GET /dokan/v1/reverse-withdrawal/vendor-due-status`.

### Expected
Balance never goes below `0`. A vendor who owed 100 has 100 credited, at most.

### Actual
Balance is **−100**. The ledger accepted 200 of credit against 100 of real debt, and
the vendor's due status flips off — which also suppresses the enforcement actions tied
to being overdue.

Scales linearly: N orders created before any completes → N × the real debt in free credit.

## Automated reproduction

`tests/ReverseWithdrawalDoubleSpendReproTest.php` (in this folder) — drop into
`tests/php/src/ReverseWithdrawal/` and run:

```bash
npx wp-env run tests-cli --env-cwd=wp-content/plugins/dokan-lite \
  vendor/bin/phpunit -c phpunit.xml --filter test_two_in_flight_payments_cannot_over_credit
```

Result on the PR branch:

```
1) WeDevs\Dokan\Test\ReverseWithdrawal\ReverseWithdrawalDoubleSpendReproTest::test_two_in_flight_payments_cannot_over_credit
A vendor owing 100 must never end below zero: the ledger took more credit than the real debt.
Failed asserting that -100.0 is equal to 0 or is greater than 0.

FAILURES!
Tests: 1, Assertions: 4, Failures: 1.
```

Both `add_payment_to_cart()` calls returned `true` — i.e. both payments passed the new guard.

## Root cause

| Location | Behaviour |
|---|---|
| `includes/ReverseWithdrawal/Helper.php:733-762` | New guard. Compares the amount against `get_vendor_balance()` — correct, but only reflects **completed** payments. Blind to in-flight orders. |
| `includes/ReverseWithdrawal/Order.php:105-120` | `insert_payment()` reads the amount from `_dokan_reverse_withdrawal_balance` line-item meta and credits it verbatim. **No balance check.** |
| `includes/ReverseWithdrawal/Manager.php:601` | `is_payment_inserted()` dedupes per `trn_id` = order id, so two orders are two independent credits. |
| `includes/ReverseWithdrawal/Cart.php:82` | `prevent_purchasing_multiple_products()` restricts within **one cart** only. Nothing prevents a second payment order while the first is unpaid. |

## Mitigating factors

COD reverse-withdrawal orders do **not** auto-complete — `Hooks.php:216` explicitly skips
them, and the credit requires `completed` or `woocommerce_payment_complete`. Each order
therefore needs an admin to complete it. That lowers severity, but the orders look like
ordinary vendor payments, so an admin completing them is the expected workflow, not a
red flag.

## Suggested fix (for the dev — not applied)

Reconcile at the completion sink as well, so every caller is covered:

```php
// includes/ReverseWithdrawal/Order.php, in insert_payment(), before building $args
$vendor_id = $order->get_customer_id();
$balance   = $manager->get_store_balance( [ 'vendor_id' => $vendor_id ] );

if ( ! is_wp_error( $balance ) ) {
    $amount = min( $amount, max( 0, (float) $balance ) );
}

if ( $amount <= 0 ) {
    return;
}
```

## Notes

- The PR's own `PaymentAmountReconciliationTest` **passes** (6 tests / 12 assertions) —
  it just had never been executed anywhere before this review. See BUG-002.
- Coverage gaps in that test file (REST route, AJAX path, sub-unit boundary) were filled
  and all pass — see `tests/` in this folder.

MdAsifHossainNadim:
## BUG-001 fixed — payments now reconcile against in-flight orders too

Thanks @dev-shahed, this reproduces exactly as described. Confirmed live before touching anything: seeded a real debt of 300, placed three COD payment orders of 300 each while none had completed (every one passed the cap, because the balance still read full), then completed all three — ledger took 900 of credit, balance **−600**, due status off.

Worth adding to the report: **no tampering is needed at all.** `#reverse_pay_balance` is `readonly` and prefilled with the full payable amount, so clicking **Pay Now** repeatedly *is* the exploit. And the new React page (`src/dashboard/reverse-withdraw`, via `useAddToCart.ts`) posts the same legacy `dokan_reverse_withdrawal_payment_to_cart` action, so both dashboards were equally affected.

### The fix — both ends, since either alone leaves a hole

**1. `Helper::add_payment_to_cart()` — stop the second order being created.**
New `Helper::get_awaiting_payment_total()` sums the vendor's reverse-withdrawal orders that have not reached the ledger yet (`pending` / `processing` / `on-hold`, skipping any already inserted via `is_payment_inserted()`), and that total is deducted from the due before the cap is applied. Stacking is refused with `payment-already-awaiting`.

**2. `Order::insert_payment()` — the completion sink clamps.**
Your suggested clamp, plus an order note when it fires. This is not belt-and-braces: the guard runs at add-to-cart, so two concurrent checkouts can both pass it before either order exists. Verified that race — two orders created simultaneously, both completed, only the first credits and the balance lands at **0, never negative**. It also covers orders placed before this fix existed.

Deliberately *not* using `min()` silently: when the clamp trims a payment the order gets a note recording what was paid vs what was credited, so the mismatch is visible to the admin rather than swallowed.

### Behaviour

| Situation | Result |
|---|---|
| 300 owed, nothing in flight | 300 accepted (unchanged) |
| 300 owed, a 300 payment awaiting completion | rejected, `payment-already-awaiting` |
| 300 owed, a 100 payment awaiting completion | up to 200 accepted, 200.01 rejected |
| Order completes for more than is owed | credit clamped to the balance, order note added |
| Over the balance / zero / negative | unchanged (`amount-exceeds-due-balance`, `invalid-amount`) |

Partial payments and early payment of a not-yet-due balance are unaffected — only the amount already committed to an unpaid order is held back.

### Verification

- **`PaymentAmountReconciliationTest`: `OK (10 tests, 24 assertions)`** — Docker was available this time, so the file's original 7 tests have now actually run, along with 3 new ones covering the stacked payment, the remaining-balance boundary, and the clamp at completion.
- **Full suite on this branch: `Tests: 277, Assertions: 1058, Failures: 1`** — the single failure is the `OrderStatusChangeTest::test_blocked_status_transitions` red that is already on `develop`. For comparison, a clean run of `develop` itself gives `Tests: 281, Errors: 1, Failures: 3`, so nothing here regresses.
- **Live, end-to-end on both dashboards** — legacy page and the new React page both now refuse the second payment with the plain-text message; the legitimate flow (pay once, admin completes) still lands the ledger at exactly 0 with the due status cleared.
- `phpcs` clean on all three files.

Two harness details behind the extra test commits, in case they bite elsewhere: the `ReverseWithdrawal\Order` hooks only bind when the feature is enabled and the container boots before any test runs, so a completed order never reached `insert_payment()` and the clamp assertion was measuring nothing (`c844b0035`); and registering those hooks inside a test leaks them for the rest of the process, which knocked over two unrelated suites until `tear_down` removed them again (`a909494e7`).

### One nuance, not fixed here

For **vendor staff**, the guard resolves the balance through `dokan_get_current_user_id()` (which maps to the parent vendor) while the order is placed under the staff member's own customer id, so the awaiting-total lookup can miss those orders. The ledger stays safe either way because the completion clamp works off `$order->get_customer_id()` — the same identity the credit targets. The underlying "staff payment credits the staff's ledger, not the vendor's" behaviour predates this PR; happy to file it separately if you agree it is worth its own issue.

Pushed as `4b6de6478` + `c844b0035` + `a909494e7`, on top of `327a6412f` which had been sitting unpushed.

MdAsifHossainNadim:
## R1 fixed — `aa2ab8227`, with R2/R4 in `4d540ebfc`

You were right on every detail, including the mechanism. Confirmed both breaks in `Manager::all()`: rows come back through `$wpdb->get_results( …, ARRAY_A )`, and there is a literal `if ( 1 === $args['per_page'] ) { $data = $data[0]; }` that collapses the result to a single row. My bulk rewrite therefore excluded nothing at all, which is strictly worse than the `is_payment_inserted()` loop it replaced.

The production consequence you traced is the part I had not thought through — the base product is `virtual` but not `downloadable`, so `needs_processing()` is true and card gateways park the order in `processing` with the ledger already credited by `woocommerce_payment_complete`. That is exactly the state the exclusion exists for, so the vendors most affected would have been the ones who paid successfully.

Fixed as you specified — array key, `count( $candidates ) + 1` so the page size can never be 1, and an `is_wp_error()` guard that holds every candidate back rather than guessing if the lookup fails.

### Verified against your probes

```
PROBE-A  processing order, is_payment_inserted=true  → awaiting count=0, total=0     (was 1 / 100)
PROBE-A2 balance=40 of genuinely new debt            → pay 40 ⇒ ACCEPTED             (was REJECTED)
PROBE-B  2 candidates, 1 settled                     → count=1, total=25             (was 2 / 85)
PROBE-C  single unsettled candidate (per_page → 1)   → count=1, total=25
```

### The test gap mattered more than the fix

Your diagnosis was precise: no existing test created an order that was both in an awaiting status *and* already credited. Three tests now cover it — the settled order alone (which is the single-candidate case that trips the page-size collapse), the settled order alongside one still waiting, and paying genuinely new debt afterwards.

I mutated `Helper.php` back to the broken version to confirm they actually catch it: **3 failures against the old code, all 15 green against the fix.** A test that cannot fail would have been no better than the last round.

### R2 / R4

Both done in `4d540ebfc`. `add_payment_to_cart()` and `get_awaiting_payment_total()` now share `sum_payment_amounts()` — the orders are still fetched as objects in the caller because the rejection message names them. Settling an order uses `save_meta_data()` instead of a full `save()` inside the status-change hook.

### R3

Consciously deferred. `wc_get_orders( [ 'limit' => -1 ] )` still hydrates the vendor's open orders, and the honest fix is an order-level marker meta so the query can filter server-side — but that only covers orders created after deploy, leaving anything already in flight at upgrade time invisible to layer 1 exactly when it matters. Better placed behind a release boundary than rushed into this PR. The N ledger queries are gone, which was the sharp edge.

### PR description

Updated to call out that vendor staff lose access to the REST `add-to-cart` route — a deliberate behaviour change, matching the `dokandar` gate the AJAX caller already had.

### Evidence

- `PaymentAmountReconciliationTest` → **`OK (15 tests, 36 assertions)`**
- Full suite → **`282 tests, 1070 assertions, 1 failure`** — the `OrderStatusChangeTest::test_blocked_status_transitions ('processing','cancelled')` red already on `develop`. No new failures.
- PHPCS on all four changed files → **0 errors**, 3 warnings on the untouched `hide_order_item_meta_*` lines.

kzamanbd (commented):
Verified locally on `4d540ebfc` (wp-env, HPOS on):

- `composer phpcs` clean on all four changed files — the only output is three pre-existing `UnusedFunctionParameter` warnings on `hide_order_item_meta_key/value`.
- `PaymentAmountReconciliationTest` → **15/15 pass, 36 assertions**.

The shape of the fix is right: guarding the shared sink covers both callers, and the completion-time clamp is the correct backstop for orders that predate the cap or race past it. The scaled-integer comparison and the `balance` vs `payable_amount` choice both hold up.

One blocking defect, reproduced with a throwaway test — see the inline comment on `get_awaiting_payment_orders()`. The rest are follow-ups.

includes/ReverseWithdrawal/Helper.php
```diff
@@ -715,6 +715,148 @@ public static function get_formatted_billing_type( $billing_type = '' ) {
...
+            ]
+        );
+
+        $candidates = [];
+
+        foreach ( $orders as $order ) {
+            if ( self::has_reverse_withdrawal_payment_in_order( $order ) ) {
+                $candidates[ $order->get_id() ] = $order;
+            }
+        }
```
kzamanbd, line 792: **Blocking — a clamped-to-zero payment order locks the vendor out permanently.**

This loop excludes candidates that have a ledger row, but not candidates carrying `Order::PAYMENT_SETTLED_META`. A payment clamped to zero writes no ledger row by design, so it is invisible to the `$settled` lookup below — and the order it belongs to stays in an awaiting status.

Path:

1. Vendor places a payment order for the full debt. The balance drops to zero before it is paid (an admin adjustment, a manual credit).
2. A card gateway fires `woocommerce_payment_complete` → `insert_payment()` clamps `$amount` to `0`, writes the marker, returns without a ledger row.
3. The reverse-withdrawal product is virtual but not downloadable, so `needs_processing()` is true and the order stays `processing` — which is in the status list above.
4. That order counts toward the awaiting total forever. Any newly accrued debt is unpayable.
5. `get_awaiting_payment_message()` offers no pay link, because `needs_payment()` is false on an order that was already paid. The vendor is told to "complete or cancel it" with no way to do either.

Only an admin completing or cancelling the order clears it. This is the same lockout class as the one closed in aa2ab82 — that commit handled the *credited* case, this is the *settled-without-a-row* case.

Reproduced with a temporary test (seeded 100 debt, order for 100 left in `processing`, ledger cleared by hand, `woocommerce_payment_complete` fired, then 40 of new debt):

```
AWAITING COUNT: 1 TOTAL: 100
REJECTED: payment-already-awaiting — Your due balance is already covered by payment
order #19, which has not been completed yet. Please complete or cancel it before paying again.
Failed asserting that false is true.   // add_payment_to_cart( 40 ) against 40.0 of real debt
```

Suggested fix:

```php
foreach ( $orders as $order ) {
    // A payment settled without a ledger row is already resolved; counting it would deduct it forever.
    if ( 'yes' === $order->get_meta( Order::PAYMENT_SETTLED_META ) ) {
        continue;
    }

    if ( self::has_reverse_withdrawal_payment_in_order( $order ) ) {
        $candidates[ $order->get_id() ] = $order;
    }
}
```

Worth a regression test mirroring `test_already_credited_order_is_not_awaiting`, with the clamped-to-zero order in place of the credited one.

includes/ReverseWithdrawal/Helper.php
```diff
@@ -715,6 +715,148 @@ public static function get_formatted_billing_type( $billing_type = '' ) {
...
+            [
+                'type'        => 'shop_order',
+                'customer_id' => $vendor_id,
+                /*
+                 * Only statuses an order can still reach `completed` from. `failed` belongs here because the customer
+                 * can retry payment on it; `cancelled` and `refunded` do not, and holding them back would strand the
+                 * vendor behind an order they can no longer act on.
+                 */
+                'status'      => [ 'wc-pending', 'wc-processing', 'wc-on-hold', 'wc-failed' ],
+                'limit'       => -1,
```
kzamanbd, line 773: **Two things on this query.**

**Performance.** `'limit' => -1` hydrates every pending / processing / on-hold / failed order where the vendor is the customer, and the loop below then loads line items for each one to check the meta. Vendors who also shop on the marketplace can carry real volume here, and this runs on both the AJAX path and the REST route on every add-to-cart. Worth measuring on a store with a busy vendor account.

A hard `limit` is not free — under-counting the awaiting total is exactly what lets a payment over-credit — but the completion-time clamp in `Order::insert_payment()` is the backstop for that, so a bounded limit with a `date_created` floor is defensible if the numbers justify it.

**Stale orders have no self-service exit.** An abandoned checkout leaves a `pending` order in this list. WooCommerce only auto-cancels unpaid orders when hold-stock is enabled, and the reverse-withdrawal base product is virtual with stock management off — so on a lot of installs it never cancels and the vendor is blocked until an admin intervenes. The pay link in `get_awaiting_payment_message()` covers the vendor who can still pay it, but not the one whose gateway failed. Consider ignoring candidates older than the hold-stock window, or letting the vendor cancel the order straight from the message.

includes/ReverseWithdrawal/Helper.php
```diff
@@ -715,6 +715,148 @@ public static function get_formatted_billing_type( $billing_type = '' ) {
...
+        // Resolve every already-credited order in one query rather than asking per order.
+        $settled = ( new Manager() )->all(
+            [
+                'trn_id'    => array_keys( $candidates ),
+                // Manager::all() validates this against the transaction type map, so it has to stay a plain string.
+                'trn_type'  => 'vendor_payment',
+                'vendor_id' => [ $vendor_id ],
+                'return'    => 'vendor_transaction',
+                // Never 1: at that exact value all() collapses the result to a single row instead of a list.
+                'per_page'  => count( $candidates ) + 1,
```
kzamanbd: The comment is accurate and the workaround is correct, but it documents a landmine rather than removing it. The real defect is in `Manager::all()` (`Manager.php:234`), which collapses its result to a single row whenever `per_page === 1` — every future caller asking for a list will hit the same trap, and the next one may not have a regression test to catch it.

If that collapse is load-bearing for `get_store_balance()` (which passes `per_page => 1` deliberately), the cleaner fix is a `return` mode that never collapses, rather than each caller padding its page size.

includes/ReverseWithdrawal/Helper.php
```diff
@@ -729,7 +871,56 @@ public static function add_payment_to_cart( $amount ) {
...
+        if ( $due <= 0 ) {
+            return new WP_Error(
+                'payment-already-awaiting',
+                // Naming the order matters: without it a vendor blocked by a stale order has no way to work out why.
+                self::get_awaiting_payment_message( $awaiting_orders ),
+                [ 'status' => 400 ]
+            );
+        }
+
+        // Compare as scaled integers so paying the exact outstanding balance is not rejected by IEEE-754 drift.
```
kzamanbd, line 956: No extensibility hook on the cap. Pro, or a marketplace with its own debt rules, cannot adjust the ceiling without replacing the whole method. Dokan convention would want an escape hatch here, before the comparison:

```php
/**
 * Filters the maximum reverse withdrawal payment a vendor may make.
 *
 * @since DOKAN_SINCE
 *
 * @param float $due       Outstanding balance, less anything already awaiting completion.
 * @param int   $vendor_id
 */
$due = (float) apply_filters( 'dokan_reverse_withdrawal_payable_amount', $due, $vendor_id );
```

The clamp in `Order::insert_payment()` still backstops whatever a filter returns, so this does not reopen the hole.

includes/REST/ReverseWithdrawalController.php
```diff
@@ -281,6 +281,24 @@ public function get_vendor_due_status_permissions_check( $request ) {
...
+     *
+     * @return bool
+     */
+    public function add_to_cart_permissions_check( $request ) {
+        /*
+         * Deliberately stricter than the read routes: `dokan_is_user_seller()` resolves vendor staff to their parent,
+         * so staff would pass while the order lands under their own customer id and credits the wrong ledger. The
+         * AJAX caller already gates on `dokandar`, which staff do not have — match it so both callers agree.
+         */
+        return is_user_logged_in() && current_user_can( 'dokandar' );
```
kzamanbd: Returning a bare `false` means vendor staff get a generic `rest_forbidden` with no indication that this is a deliberate policy change — and given the PR body calls this out as a behaviour change, the people who hit it are the ones who most need the reason.

A `WP_Error` matches the newer Dokan controllers and gives them something actionable:

```php
return new WP_Error(
    'dokan_rest_cannot_add_to_cart',
    __( 'Only vendors can make a reverse withdrawal payment.', 'dokan-lite' ),
    [ 'status' => rest_authorization_required_code() ]
);
```

Non-blocking — it is consistent with `get_vendor_due_status_permissions_check()` as written.

includes/ReverseWithdrawal/Order.php
```diff
@@ -13,6 +13,16 @@
...
 class Order {
+
+    /**
+     * Marks an order whose payment was resolved without writing a ledger row.
+     *
+     * @since DOKAN_SINCE
+     *
+     * @var string
+     */
+    const PAYMENT_SETTLED_META = '_dokan_reverse_withdrawal_payment_processed';
```
kzamanbd: Constant name and meta value disagree — `PAYMENT_SETTLED_META` vs `..._payment_processed`. Cosmetic, but this value gets persisted to order meta the moment it ships, so it is the last chance to align them (`_dokan_reverse_withdrawal_payment_settled`) before it becomes a migration.

kzamanbd (changes requested):
Requesting changes on the blocking item only — the [clamped-to-zero lockout](https://github.com/getdokan/dokan/pull/3353#discussion_r-see-inline) in `Helper::get_awaiting_payment_orders()`, detailed in the inline comment on that loop. Reproduced locally; a vendor whose payment is clamped to zero at `woocommerce_payment_complete` leaves an order sitting in `processing` with no ledger row, and can never pay newly accrued debt without an admin completing or cancelling it by hand.

Everything else I raised is a follow-up and does not need to block the merge:

- `limit => -1` on the awaiting-orders query, and stale `pending` orders with no self-service exit
- the `per_page + 1` workaround around `Manager::all()`
- no extensibility filter on the cap
- bare `false` from `add_to_cart_permissions_check()`
- `PAYMENT_SETTLED_META` name vs value

Verified on `4d540ebfc`: `composer phpcs` clean on the changed files, `PaymentAmountReconciliationTest` 15/15 pass under wp-env with HPOS on. The core approach — guarding the shared sink, clamping again at completion — is right.

includes/ReverseWithdrawal/Helper.php
MdAsifHossainNadim: Fixed in bdfa562ed. Confirmed before touching anything, with a more ordinary trigger than a hand-cleared ledger — an admin manual credit landing between checkout and payment completion:

```
marker='yes' | ledger row=false | status=processing | needs_payment=false
balance=40 | awaiting count=1 | awaiting total=100
pay 40 => payment-already-awaiting :: "…order #19… complete or cancel it before paying again."
```

Your reading of the trap is exactly right, including the cruellest part: `needs_payment()` is false on an already-paid `processing` order, so the message cannot even offer a pay link while telling the vendor to do something only an admin can do.

Applied your suggested skip. Worth recording why it is safe rather than a new hole: `Order::insert_payment()` reads the same marker before crediting, so an order excluded here can never reach the ledger later — the two guards agree, which is what makes dropping it from the awaiting total correct.

Added `test_settled_order_without_a_ledger_row_is_not_awaiting`, mirroring `test_already_credited_order_is_not_awaiting` as you suggested. It drives the real `woocommerce_payment_complete` path and asserts the marker, the absent ledger row, the empty awaiting list, and that newly accrued debt stays payable. **Mutation-checked** — neutralising the skip fails exactly that test and nothing else.

After: `awaiting count=0, total=0, pay 40 => ACCEPTED`. `PaymentAmountReconciliationTest` 16/16.

includes/REST/ReverseWithdrawalController.php
MdAsifHossainNadim: Fixed in 0ca2562e1 — took your snippet as written.

```
vendor=true | customer=dokan_rest_cannot_add_to_cart/403 | staff=dokan_rest_cannot_add_to_cart/403 | guest=dokan_rest_cannot_add_to_cart/401
```

Agreed on the reasoning: since the PR body calls the staff exclusion out as a deliberate behaviour change, the people who hit it are the ones who most need to be told why.

includes/ReverseWithdrawal/Order.php
MdAsifHossainNadim: Fixed in 0ca2562e1 — now `_dokan_reverse_withdrawal_payment_settled`.

You are right that this was the last free moment: nothing has been released with the old key, so it is a rename today and a migration the moment it ships. Verified the new key is written and the old one is absent, and grepped both Lite and Pro for the old literal — no other references.

includes/ReverseWithdrawal/Helper.php
MdAsifHossainNadim: Agreed on both halves, and I would like to take them as a follow-up rather than in this PR.

On the limit: your own caveat is the reason to be careful — under-counting the awaiting total is exactly the failure mode the completion clamp exists to catch, so trading a correctness property for an unmeasured performance win is the wrong order of operations. Worth profiling on a store with a busy vendor account first, then choosing between a bounded limit with a `date_created` floor and an order-level marker meta that makes the query filterable server-side. The marker route has its own catch: it only covers orders created after deploy, leaving in-flight ones invisible to the deduction at upgrade time.

On stale orders: correct that the base product is virtual with stock management off, so hold-stock never cancels them on most installs. The clamped-to-zero case you raised separately was the sharpest edge of this and is now fixed; what remains is the genuinely-unpaid stale order, where a vendor-facing cancel action is probably the better answer than a time window.

Leaving this thread open so it does not get lost — I will file it as an issue and link it here.

includes/ReverseWithdrawal/Helper.php
MdAsifHossainNadim: Agreed that the workaround documents a landmine rather than removing it, and that `Manager::all()` is where the defect actually lives.

I would still keep the padding in this PR. The collapse is load-bearing for `get_store_balance()`, which passes `per_page => 1` deliberately, so the real fix is the non-collapsing return mode you describe — and that changes shared query semantics for every existing caller. Doing that inside an embargoed security fix widens the blast radius of a patch that needs to land narrow and be easy to review.

Your point about the next caller not having a regression test to catch it is the real risk, so this should not just sit. Filing it as a follow-up and linking it here.

includes/ReverseWithdrawal/Helper.php
MdAsifHossainNadim: I would rather not add this one, and it is worth putting the reason on the record since a filter was already considered and dropped earlier in this PR's history.

Your point that the clamp backstops the filter is right about the *ledger* — nothing a filter returns can push the balance negative, because `Order::insert_payment()` clamps again at completion. But the ledger is not the only thing at stake. A site that raises the cap lets a vendor pay real money above their debt through a real gateway, and then the credit is trimmed at completion and they get an order note about it. Money taken, credit withheld, no refund path. That is a worse failure than the rigidity the filter removes.

If we do want the escape hatch, the condition is that it applies at both ends — the cap in `add_payment_to_cart()` and the clamp in `insert_payment()` — so a marketplace with its own debt rules gets a consistent ceiling rather than a payable amount the ledger then refuses to honour. That is a design change worth its own discussion rather than a one-line filter in a security fix.

Happy to be overruled if you still want it; just flagging that the half-applied version is the dangerous shape.

includes/ReverseWithdrawal/Helper.php
MdAsifHossainNadim: Agreed, and fixed at the source in 8400745ce rather than padded around.

`Manager::all()` now returns a list for `vendor_transaction` at every page size — the `per_page === 1` collapse is gone. `balance` keeps its collapse, because its callers (`get_store_balance()`, the opening-balance lookup) ask for exactly one aggregate row on purpose. I checked every caller in Lite and Pro before touching it: nothing consumed the collapsed `vendor_transaction` shape, and `get_store_transactions()` — whose page size comes straight from the REST request — was actually returning a bare row to any client asking for one per page, so this is a fix there too.

The `+ 1` padding and its comment are dropped, and `test_transaction_list_keeps_its_shape_at_one_per_page` pins the shape. Mutation-checked: restoring the collapse fails that test plus the two settled-order tests, which now lean on the fix directly.

includes/ReverseWithdrawal/Helper.php
MdAsifHossainNadim: Added in 7cab742e8 as `dokan_reverse_withdrawal_payable_amount`, with the signature you proposed plus a `$context` argument.

The one thing I changed from the snippet: it is applied at **both** ends — when the cap is checked in `Helper::add_payment_to_cart()` (`'cart'`) and again when the ledger is credited in `Order::insert_payment()` (`'completion'`). Filtering only the cart side would let a raised ceiling accept a payment the completion clamp then trims, so the vendor's money is taken and the credit withheld. With both ends reading the same filter, whatever ceiling the cart accepts is what the ledger honours.

Two tests cover the pair: raising the ceiling is honoured end to end (vendor credited what they actually paid, and the default ceiling returns once the filter is removed), and lowering it only at completion still clamps — so the filter cannot be used to sneak credit past the ledger from one side.

includes/ReverseWithdrawal/Helper.php
MdAsifHossainNadim: Taking the two halves separately.

**Stale orders — fixed in a0a545621.** The rejection now includes WooCommerce's own cancel URL for the first awaiting order in a status the customer is allowed to cancel from (`woocommerce_valid_order_statuses_for_cancel`, so `pending`/`failed` by default). You are right that hold-stock never cancels these — the base product is virtual with stock management off — so without this the vendor stayed blocked until an admin noticed. Covered by `test_awaiting_payment_rejection_offers_a_way_out`, which asserts the order number, the pay URL and the cancel URL are all in the message.

**`limit => -1` — measured before deciding, and leaving it as is for this PR.** The query is bounded by *unfinished* orders for one customer, and each candidate now costs one `get_items()` read plus a single ledger query for the whole set (the per-order lookups were already collapsed to one). I did prototype the alternative — an order-level marker written at checkout so the query can filter server-side — and pulled it back out: it needs `meta_query`, which only the HPOS datastore accepts (the post datastore ignores it and WooCommerce warns), it cannot see orders placed before the upgrade so a legacy sweep is still required, and it adds a persisted meta and an option to a security patch that should stay narrow. Your own caveat is the deciding one: under-counting the awaiting total is exactly what lets a payment over-credit, so any bound has to be proven safe on real numbers before it goes in. Happy to open a follow-up with the prototype attached if you want it profiled on a store with a busy vendor account.

MdAsifHossainNadim: Correction on the cancel link, fixed in 7650ffec2 — the version I resolved this with did not actually work.

`get_cancel_order_url()` goes through `wp_nonce_url()`, which HTML-escapes the query string. In a plain-text message that reads as `&amp;order_id=…`, so PHP parsed the keys as `amp;order_id` / `amp;_wpnonce` and `WC_Form_Handler::cancel_order()` failed its `isset( $_GET['order_id'] )` guard and returned silently. My test missed it because `order_id=19` still appears in the string, right after `amp;`.

Now uses `get_cancel_order_url_raw()`. The test asserts the exact raw URL is in the message and that no `&amp;` reaches it — verified the tightened assertion fails on the old line and passes on the new one. Also drove the link end to end: pulled it out of the message the way a vendor would, replayed it through `WC_Form_Handler::cancel_order()` as the vendor, order went `pending → cancelled`, and the next payment attempt was accepted.

kzamanbd (approved):

PR #3358, opened 2026-07-30
fix(product-list): show the real product type instead of labeling every unknown type Simple

akzmoudud:
### All Submissions:

* [x] My code is tested

The vendor dashboard product list (`/dashboard/new/#products/`) reported **every** product type it did not recognise as "Simple", with the `Package` icon. A vendor who created a subscription product saw it listed as a simple product.

`getProductTypeLabel()` branched on `grouped` / `external` / `variable` and returned `__( 'Simple' )` for everything else; `ProductTypeIcon()` did the same with `Package`. The legacy PHP row template never had this bug — [`products-listing-row.php`](https://github.com/getdokan/dokan/blob/develop/templates/products/products-listing-row.php) falls back to the type slug (`ucfirst()`) for types it does not know. The React port dropped that fallback.

This PR restores it and improves on it:

* **Label** resolves in three tiers — Lite's own labels → the vendor's localized `dokan_product_types` map → a humanized slug. The middle tier is the important one: that map is already localized to JS from PHP, so a module-registered type carries the module's own translated label (`subscription` → "Simple subscription"), with no per-type code in Lite and no duplicated strings.
* **Icon** becomes a map lookup, with entries for the module-owned types whose products remain in the catalog after the module is deactivated (`subscription`, `variable-subscription`, `auction`). `auction` uses the same `Hammer` glyph the auction module uses, so toggling that module doesn't change the icon — only the extras the module layers on top.
* **New filter** `dokan_product_type_label` for overrides, matching the existing `dokan_product_status_label` pattern.
* **Guard**: a row arriving without a `type` now falls back to `simple` rather than throwing inside the cell renderer and blanking the whole listing.
* `PRODUCT_TYPE_OPTIONS` is derived from the same label map instead of repeating the four strings, so the Type column and the Type filter cannot drift apart. Same four options, same order, same text domain — verified the strings still extract with `wp i18n make-pot`.

Modules that need richer output keep overriding the whole cell through `dokan_product_list_table_fields`, unchanged — the auction module still supplies its own tooltip and greys the icon for closed and failed auctions.

**No behaviour change for bookings.** `useProducts.ts` is untouched, so `exclude_types: [ 'booking' ]` is unchanged and booking products still never appear in this list. The booking module renders through its own `ProductsTab` component, which imports nothing from `ProductList`.

### Related Pull Request(s)

* N/A

### Closes

* Closes getdokan/plugin-internal-tasks#2183

1. Activate WooCommerce Subscriptions and the **Product Subscription** (`vsp`) module.
2. As a vendor, go to `/dashboard/new/#products/` and create a Simple subscription product.
3. **Before:** the Type column shows the simple-product icon, tooltip "Simple". **After:** a recurring icon, tooltip "Simple subscription".
4. Deactivate the `vsp` module and reload. The row stays correct — tooltip falls back to "Subscription" (the module's own translated string is gone with the module) and the icon remains the subscription glyph rather than reverting to the simple one.
5. Repeat with an auction product and the `simple-auction` module toggled off: the row shows a hammer and "Auction" instead of claiming to be Simple. With the module on, the auction module's own column override applies exactly as before.
6. Confirm the four core types (simple, variable, grouped, external) are unchanged in both the Type column and the Type filter dropdown.
7. Confirm booking products still do not appear in this list.

### Changelog entry

*Fix: vendor dashboard product list labelled every non-core product type as "Simple"*

Previously, the new vendor dashboard product list recognised only simple, variable, grouped and external products. Every other type — subscriptions, auctions and any type added by a module or third-party plugin — was displayed as "Simple" with the simple-product icon, so vendors could not tell what a product actually was.

Now the list shows the real product type. Types registered through `dokan_product_types` display the registering module's own translated label, and any other unrecognised type falls back to a readable name derived from its slug instead of being misreported as Simple. Product types whose module has been deactivated keep a meaningful icon, since their products remain in the catalog.

### Before Changes

A Simple subscription product listed in the vendor dashboard shows the simple-product icon with the tooltip "Simple". Auction products do the same.

### After Changes

The same product shows a recurring icon with the tooltip "Simple subscription"; auction products show a hammer with "Auction".

### PR Self Review Checklist:

* [x] Code follows the code style guidelines
* [x] Naming is clear
* [x] KISS — three-tier lookup with a documented reason for each tier
* [x] DRY — the four core labels exist once and feed both the column and the filter
* [x] Readable — the nested `if` chains are replaced by two lookup maps
* [x] No performance concerns — map lookups per row, no added requests
* [x] No grammar errors

### Notes for the reviewer

Two deliberate limitations, both matching the legacy template's behaviour:

1. The humanized-slug fallback is not translatable — there is no `__()` call site for a slug Lite has never seen. It applies only when no module has registered a label for that type. The legacy template has the identical gap (`ucfirst()` with no `__()`).
2. The Type **filter** dropdown still lists only the four core types plus whatever modules append via `dokan_product_list_type_options`. So a type whose module is deactivated is visible in the list but not filterable. Fixing that needs a server-side change and is out of scope here.

`PRODUCT_TYPE_ICONS` naming module-owned type slugs in Lite is a conscious trade-off: a deactivated module's JS never runs, so the module cannot supply an icon for products that still exist. The alternative — letting those rows fall back to the generic icon — reintroduces a milder version of the reported bug. A future `dokan_product_type_icon` filter would let modules own this properly.

## Summary by CodeRabbit

* **New Features**
  * Product lists now support additional product types, including subscriptions, variable subscriptions, auctions, and other variants.
  * Product type icons and labels are displayed dynamically based on the available product types.
  * Unknown product types are presented with readable names instead of raw identifiers.
  * Product filters now automatically include all supported product types.
* **Bug Fixes**
  * Improved consistency between product type labels, icons, and filtering options.

### DIFF
--- a/src/dashboard/products/ProductList.tsx
+++ b/src/dashboard/products/ProductList.tsx
@@ -4,7 +4,17 @@ import { useToast } from '@getdokan/dokan-ui';
 import { addAction, applyFilters, removeAction } from '@wordpress/hooks';
 import { Fill } from '@wordpress/components';
 import { useNavigate } from 'react-router-dom';
-import { Boxes, Package, ExternalLink, LayoutGrid, Plus } from 'lucide-react';
+import {
+    Boxes,
+    Package,
+    ExternalLink,
+    LayoutGrid,
+    Plus,
+    RefreshCw,
+    Repeat,
+    Hammer,
+} from 'lucide-react';
+import type { LucideIcon } from 'lucide-react';
 import {
     DataViews,
     DokanBadge,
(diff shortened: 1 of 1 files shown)


MdAsifHossainNadim (approved):
## Verdict: ✅ Approve

### The exact issue
The new vendor dashboard product list (`/dashboard/new/#products/`) only knew four product types. `getProductTypeLabel()` branched on `grouped` / `external` / `variable` and returned "Simple" for everything else, and `ProductTypeIcon()` did the same with the `Package` glyph. So a WooCommerce Subscriptions product, an auction, or any module/third-party type was reported to the vendor as a simple product. The legacy PHP row template never had this gap — `products-listing-row.php` falls back to `ucfirst( $type )` for unknown types — and the React port dropped that fallback (getdokan/plugin-internal-tasks#2183).

### Why the fix is correct
- **Three-tier label lookup**: Lite's own four labels → the vendor's localized `window.dokan.product_types` map → a humanized slug. The middle tier is the one that matters: that map is already localized in `Assets::enqueue_front_scripts()` through the `dokan_product_types` filter, and Pro's `Products::set_default_product_types()` turns Lite's `[ 'simple' ]` list seed into a real map, so module labels ("Simple subscription", "Auction") flow through with no per-type code in Lite. With Lite alone the seed stays a list, `getVendorProductTypes()` correctly returns `{}`, and the humanized slug tier carries it.
- **Icon map** with `subscription` / `variable-subscription` / `auction` entries keeps a meaningful glyph for products whose module is deactivated (their products remain in the catalog while the module's JS no longer runs). Unknown types fall back to `Package` as before.
- **`PRODUCT_TYPE_OPTIONS`** is derived from the same label map, so the Type column and the Type filter cannot drift; same four options, same order, same text domain.
- **`item.type || 'simple'`** guard means a row without a type no longer throws inside the cell renderer.
- New **`dokan_product_type_label`** filter mirrors the existing `dokan_product_status_label` pattern and carries `@since DOKAN_SINCE`.
- The auction module's whole-cell override via `dokan_product_list_table_fields` is untouched and still wins when the module is on. `useProducts.ts` is untouched, so `exclude_types: [ 'booking' ]` still applies.

### Testing scope
Verified locally (Lite + Pro 5.1.2, WC 11.0.1, WC Subscriptions 8.3.0, WC Simple Auctions activated for the test) with the PR applied on `develop` and `npm run build`, toggling the Pro modules through wp-cli:

| Row | Before | After |
|---|---|---|
| WC Subscriptions product, `vsp` module **off** | package icon, tooltip "Simple" | refresh icon, tooltip "Subscription" (humanized tier) |
| Same product, `vsp` module **on** | — | refresh icon, tooltip "Simple subscription" (module's own label) |
| Auction product, `auction` module **off** | package icon, "Simple" | hammer, "Auction" |
| Auction product, `auction` module **on** | hammer, "Auction" (module override) | unchanged |
| Simple / Variable / Grouped / External rows | package / boxes / grid / external-link with "Simple" / "Variable" / "Grouped" / "External/Affiliate" | unchanged |
| Type filter dropdown (modules off) | Simple, Variable, Grouped, External/Affiliate | same four, same order |
| Type filter dropdown (`auction` on) | + Auction (appended by the module) | + Auction |
| Booking product | not listed | not listed |

- ESLint clean on the file; `tsc` 964 → 964 errors (none in this file); the branch merges cleanly onto `develop`.
- The two console warnings on this page (RawHTML `<div>` inside `<p>` in the title cell, and a dokan-ui menu key warning) reproduce on `develop` without this PR.

### Non-blocking notes
- CodeRabbit's open comment asks for `@since DOKAN_SINCE` on the two new module-private docblocks (`getVendorProductTypes`, `PRODUCT_TYPE_ICONS`). The new filter already carries it; author's call, but please resolve/dismiss the thread so it doesn't sit on the merge queue.
- The two limitations disclosed in the PR body (humanized fallback not translatable; Type filter can't select a type whose module is off) are accepted — both match the legacy behaviour and need server-side work to improve.
- Not exercised live: the `product_subscription` module's per-vendor trimming of the type map (prio 99). A vendor whose pack no longer allows a type they already own would get the humanized label rather than the module's — still correct, never "Simple".

MdAsifHossainNadim (approved):
Duplicate of my review above (posted twice by a network retry) — see that one for the full verdict, reasoning and testing scope.

dev-shahed:
@akzmoudud 

QA'd on Lite+Pro 5.1.1. Fix works: subscription → `RefreshCw` + "Simple subscription", `variable-subscription` → `Repeat`, `product_pack` → "Product pack" (was "Simple"). With `vsp`/`auction` off, icons survive and labels fall back to "Subscription" / "Auction" — steps 4 and 5 reproduce. Core types, Type filter and booking exclusion unchanged.

Two before merge:

1. **Rebase.** Branch is based on the 5.0.11-era `develop`. As-is against Pro 5.1.1 the dashboard throws `TypeError: Cannot read properties of undefined (reading 'useToast')` from `product-adv/vendor-dashboard.js`. Gone once merged onto current `develop` — not caused by this diff, but the branch as it stands isn't what tested green.
2. **Changelog wording.** WooCommerce normalizes unregistered types to `simple` before the REST payload, so a type like `custom_widget_thing` still renders "Simple". The humanize tier only reaches WooCommerce-registered types.

Non-blocking: the Type cell has no accessible name (icon-only, `aria-hidden` SVG, non-focusable trigger) — the column announces as empty. Pre-existing, but this PR is what makes it carry real information. `role="img" aria-label={ label }` on the existing trigger reuses the label already computed for the tooltip.

PR #3360, opened 2026-08-06
test(e2e): fix four CI flakes at root cause and make the Stripe block…

shohan0120:
The nightly `develop` e2e run had been red since 2026-07-31. Two different problems were tangled together in those failures and needed different fixes. Both are now fixed, and the branch also rebalances the shard baseline and adds an end-of-shard isolation re-run to CI.

No assertion was weakened and no test was skipped to achieve this. One test that was passing for the wrong reason is now stricter, not looser.

### 1. Stripe block checkout — 12 failures to 0

Root cause, reported by CI once the suite was made to describe its own failure: `Block validation said: Your mobile phone number is invalid.`

The seeded billing phone was the reserved fictional `(555) 555-5555`. Stripe Link validates the billing phone as a *mobile* number and rejects that range, so WooCommerce Blocks never submitted the form. `POST /wc/store/v1/checkout` never fired, no order was created, and the settle poll could only ever report `"none"`. The card was never the problem.

* Seed a real-format mobile number, and fill the guest phone explicitly in `fillBlockGuestDetails()` with fail-loud assertions.
* **Impossible wait budget.** Post-click waits were 60s + 120s = 180s minimum inside specs configured at 150s, so the poll that decides pass/fail could never complete on CI — the test died mid-poll and reported the poll's placeholder rather than a verdict. The SPA redirect now gets 15s on CI and the full 60s locally where it is the fast path; the two affected specs move to 240s to match the sibling money specs.
* **The suite was blind to its own failure mode.** After clicking Place Order it only polled the REST API, so "the Store API rejected the payment" and "the click was a no-op" surfaced identically. `placeBlockOrderExpectReceived()` now records every Store API checkout POST and reads back the block notice banner.
* **Closes a test that passed for the wrong reason.** `SE-GUEST-03` was passing on a missing-phone form error without ever submitting the declined card. The decline assertion now matches by exclusion — the error must not be a form-validation message — because real declines surface as "Something went wrong. Please contact us to get assistance.", which a payment-keyword match rejected.

### 2. Cross-spec state pollution — the four "flakes" that were not flakes

Four tests failed on CI and passed locally in isolation, across two runs. They were never timing flakes: each is a *different spec earlier on the same shard* leaving global state dirty. A clean-environment isolation run cannot reproduce this by construction, because a pristine database is exactly the condition under which the victim passes. Three earlier attempts that treated them as races did not work.

| Victim | Polluter (earlier on the same shard) | Leak |
|---|---|---|
| `adminDataViewsMigration` Store Reviews tabs | `adminModules.spec.ts` | the "reset" derived its original state from catalogue *presence*, which is always true, so it deactivated the module instead of restoring it |
| `productAdvertising` checkout | `paymentsPage.ts` | `togglePaymentMethod()` blind-flipped despite its "enable" comment, disabling an already-enabled offline gateway |
| `newAuction` React editor | `productsDetails.spec.ts` | pinned `vendor_product_editor` to `legacy` and never restored it, unlike its three sibling specs |
| `newProducts` empty Draft tab | `newProductForm.spec.ts` | saved a draft product and never deleted it |

A sweep for the same bug class found six more, all fixed here:

* `productsDetails.spec.ts` also flips its seeded product to `draft` and deleted none of its seven seeded products.
* `setting.spec.ts` ends the terms-and-conditions test on `off` when the seeded default is `on` — the leak `singleStore.spec.ts` was carrying a defensive re-seed for. Fixed at both ends.
* `vendorSettings.spec.ts` kept its catalog and min-max resets *inside a test*, so a skip, a failure or a `--grep` left them applied.
* `newWithdrawB15.spec.ts` left the disbursement schedule configured.
* `noticeAndPromotionPage.ts` — unrelated. The promotion feed can serve more than one live promotion, so the slider renders n slides and a bare `toBeVisible()` throws a strict-mode violation. It now asserts every match, which is stricter than the single-element assertion it replaces.

Two traps worth knowing when fixing this class: `dbUtils.updateOptionValue` deep-**merges**, so it can never remove a key a spec added — use `setOptionValue` for a true restore. And a restore placed inside a test only runs if that test runs *and* passes.

### 3. Shard rebalance

`utils/shard-durations.json` regenerated (the baseline was from 2026-07-15, 190 specs, now 195). Measured spread went from 71.2 min to roughly 0.1 min estimated, confirmed against real CI wall-clock: shards now finish in 15–21 min rather than 24–48 min.

### 4. CI — end-of-shard isolation re-run

After a shard finishes, only the tests that failed are re-run without their shard-mates, and that second pass decides the shard result.

* No recorded test failures means no re-run. If the first pass died from a crash or timeout rather than assertions, `--last-failed` would select zero tests and trivially pass; the gate keeps the shard red instead.
* The re-run overrides the reporters, because the blob reporter would emit a second set of results for the same tests and `merge-reports` would double-count them. The merged HTML report therefore keeps showing the first-pass failures.
* It runs before the coverage step, which overwrites `.last-run.json`.
* Every rescued test is written to the step summary, naming shard-state pollution as the likely cause.

**Trade-off, stated deliberately.** Playwright's in-place `retries` cannot green a deterministic shard-context failure, because all three attempts run against the same dirty state. This can, and the re-run inherits retries. The four pollution bugs in section 2 would have been reported green-with-a-warning rather than red under this policy. That was an explicit decision, not an oversight.

### Verification

| Run | Commit | Result |
|---|---|---|
| 31073883028 (nightly, pre-work) | `43ab2b728` | 17 failures |
| 31241990565 | `626a8fe4a` | 5 failures, 0 Stripe |
| 31249977219 | `3949b5306` | 12/12 shards green |
| 31252141880 | `b077d3878` | 12/12 shards green |

Each state-restore fix was verified by running the polluter and reading the global state back — modules, offline gateways, product editor, terms-and-conditions, catalog mode, withdraw disbursement and draft count — rather than by re-running the victim, which proves nothing. `tsc --noEmit` holds at the develop baseline of 82 throughout.

### Known gaps

* The isolation re-run has never executed in CI: both green runs had zero failing shards, so all three new steps skipped. It is verified locally against a throwaway spec only.
* `productsDetails` "vendor can create product tags" is a pre-existing flake measured at 20% raw failure locally, and it consumed both CI retries in one run. Not fixed here.
* `newProducts` "vendor can view product list page" is a new retry-masked flake, first seen inside a green run.

### All Submissions:

* [x] My code is tested

Test-only and CI-only. No plugin source is touched. See sections 1–4 above.

### Related Pull Request(s)

* None

### Closes

* Closes getdokan/plugin-internal-tasks#2211

1. Trigger the E2E workflow on this branch and confirm 12/12 shards pass (runs 31249977219 and 31252141880 already do).
2. For the state-restore fixes specifically, run a polluter on its own and read the global back, for example:
   `npx playwright test --project=e2e_tests --workers=1 tests/e2e/admin/adminModules.spec.ts`
   then confirm the Store Reviews module is still active. Repeat for `payments`, `productsDetails` and `newProductForm` against their respective globals. Re-running the victim test proves nothing, because it passes on a clean database either way.

### Changelog entry

*Fix e2e suite failures caused by Stripe block-checkout phone validation and cross-spec shard state leaks*

Previously the nightly e2e run was red: the Stripe block checkout never submitted because the seeded billing phone was rejected as a non-mobile number, and four further tests failed because other specs on the same shard left global options, modules and seeded products dirty. Now the checkout submits, every spec restores the global state it changes, the shard baseline is balanced, and a shard's failures are re-run in isolation before the shard is called red.

### Before Changes

Nightly `develop` e2e red since 2026-07-31, 17 failures at its worst. Shard wall-clock spread 15.7–86.9 min.

### After Changes

12/12 shards green on two consecutive runs. Shard wall-clock 15–21 min.

### PR Self Review Checklist:

* DRY: Don't Repeat Yourself.
* Performance issues
* Grammar errors.

### FOR PR REVIEWER ONLY:


tests/pw/tests/e2e/stripe-express/stripeExpressEdge.spec.ts
```diff
@@ -236,7 +236,7 @@ test.describe('Stripe Express — edge & resilience @pro', () => {
         const longUnicodeFirst = 'Renée-Ünüçödé-名前-' + 'あ'.repeat(12);
         const longUnicodeLast = 'Tëst-Sürñámé-日本語-' + 'Ω'.repeat(12);
         const longUnicodeAddr = 'Straße Ñoño 日本語 番地 ' + 'Ä'.repeat(48);
-        const billing = { first_name: longUnicodeFirst, last_name: longUnicodeLast, company: '', address_1: longUnicodeAddr, address_2: '', city: 'New York', state: 'NY', postcode: '10003', country: 'US', email: 'customer1@email.com', phone: '(555) 555-5555' };
+        const billing = { first_name: longUnicodeFirst, last_name: longUnicodeLast, company: '', address_1: longUnicodeAddr, address_2: '', city: 'New York', state: 'NY', postcode: '10003', country: 'US', email: 'customer1@email.com', phone: '+14152367890' };
```
shohan0120: shouldnt we put this on testdata?

shohan0120: Agreed — done in 2bd8801ac.

It turned out to be wider than this one line: the phone was hardcoded in **seven** places across five files — three billing seeds in `payloads.ts`, three guest-checkout fills in `stripeExpressGuest.spec.ts`, the page object's customer fixture and the orders page fixture. All of them now reference a single exported `MOBILE_TEST_PHONE` in `utils/payloads.ts`.

I put the reason for the value on the constant rather than leaving it bare. Stripe Link validates the billing phone as a *mobile* number and rejects the reserved fictional `(555) 555-5555` range; when it does, WooCommerce Blocks refuses to submit at all, so no Store API request fires and no order is created — the test fails with no visible card error. Without that note the value looks arbitrary and the obvious tidy-up back to a fictional number would silently break the Blocks checkout again.

Side benefit: `+14152367890` is a plausibly-real San Francisco number. It is seed data and nothing is ever sent to it, but if we want to swap it for something safer (libphonenumber's canonical `+14155552671`, inside the reserved 555 range but with a valid exchange) that is now a one-line change instead of a hunt across five files.

No behaviour change — same value everywhere it was already used. `tsc` holds at the 82-error develop baseline with a byte-identical error set.

Note: `paypalMarketplaceGuest.spec.ts` has the same literal but is not in this PR, so I left it for the PayPal branch rather than pulling an unrelated file in here.

PR #3367, opened 2026-08-10
fix(orders): apply Hide Customer Info to the vendor order export

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

With **Dokan → Settings → Selling Options → Hide Customer Info** enabled, `MiscHooks::hide_customer_info_from_vendor_order_export()` dropped only two columns from the vendor order CSV — `billing_email` and `customer_ip`. Everything else survived: the customer's first name, last name, full name, phone, and the complete billing *and* shipping address. The vendor order details page hides that whole block, so the export contradicted the setting it exists to honour.

The callback now drops every `billing_*` and `shipping_*` column plus `customer_ip`. `customer_note` stays, matching the details page, which keeps showing it. `order_shipping` and `order_shipping_cost` are untouched — they carry the `order_` prefix and are marketplace data, not customer data.

Matching by prefix rather than listing the columns keeps this from drifting: the canonical list lives in `dokan_order_csv_headers()` in another file, and a column added there later — or added by Pro or a third party earlier in the same `dokan_csv_export_headers` chain — is covered without touching this method.

Prefix matching cuts both ways, though: `shipping_*` is not exclusively customer data, so a shipment-tracking module adding `shipping_tracking_number` would be dropped silently and read as a bug in that module. The drop list is therefore filterable via `dokan_hidden_customer_info_csv_columns`.

The callback also moves from priority 20 to 999. At 20 it never saw columns another callback registered later, so a third party adding `billing_vat_number` at priority 30 leaked it while the setting was on. 999 covers every realistic registration — verified that a `billing_*` column added at 10, 20, 50, 100, 500 or 998 is now dropped — while still leaving room for a deliberate later hook. (`PHP_INT_MAX` was considered and rejected: it is used three times across Lite and Pro against twenty uses of 99, and it leaves nobody any room to run after us.)

That late priority is also what makes the dedicated filter worth having: the column list, rather than hook ordering, is the supported way to put a legitimate `shipping_*` column back.

`@return mixed` on the method was always wrong — it returns an array on every path — and is corrected while the docblock is being edited.

### Tests

`tests/php/src/Orders/HideCustomerInfoExportTest.php` covers four cases:

| Case | Asserts |
|---|---|
| Setting off | `billing_email`, `billing_full_name`, `shipping_address_1`, `customer_ip` all still present |
| Setting on | no surviving column is customer data; `order_shipping`, `order_shipping_cost` and `customer_note` are deliberately kept |
| Third-party columns | `billing_vat_number` (priority 10) and `shipping_tracking_number` (priority 999) both dropped, `vendor_internal_reference` kept |
| Filter | `shipping_tracking_number` can be kept via `dokan_hidden_customer_info_csv_columns` while `billing_email` still goes |
| Filter payload | the hidden-column list arrives as a plain 0-based list |

### Closes

* Closes getdokan/dokan-pro#5962
* Closes getdokan/plugin-internal-tasks#2224

1. Go to **WP Admin → Dokan → Settings → Selling Options** and enable **Hide Customer Info**.
2. Log in as a vendor, open **Vendor Dashboard → Orders**, and click **Export All**.
3. Open the CSV. It should contain only: Order No, Order Items, Shipping method, Shipping Cost, Payment method, Order Total, Earnings, Order Status, Order Date, Customer Note — no name, email, phone, address or IP.
4. Repeat with **Export Filtered** after applying a status or date filter — same columns.
5. Turn the setting back off and export again: all 33 columns return, unchanged from before this PR.
6. Check the rows line up with the headers in both modes (10 values per row with the setting on, 33 with it off).

Both the legacy and React vendor dashboards post to the same `Dashboard\Templates\Orders::handle_order_export()` handler, so testing either covers both.

### Changelog entry

*Fix - "Hide Customer Info" was not applied to the vendor order export*

With the **Hide Customer Info** selling option enabled, customer information was correctly hidden on the vendor order details page but the vendor's exported order CSV still contained the customer's name, phone and full billing and shipping addresses. The export now drops every billing and shipping column along with the customer IP, so it matches what the setting hides everywhere else.

### Before Changes

Setting on, CSV header row (31 columns):

```
Order No, Order Items, Shipping method, Shipping Cost, Payment method, Order Total, Earnings, Order Status, Order Date,
Billing Company, Billing First Name, Billing Last Name, Billing Full Name, Billing Phone, Billing Address 1, Billing Address 2,
Billing City, Billing State, Billing Postcode, Billing Country, Shipping Company, Shipping First Name, Shipping Last Name,
Shipping Full Name, Shipping Address 1, Shipping Address 2, Shipping City, Shipping State, Shipping Postcode, Shipping Country,
Customer Note
```

### After Changes

Setting on, CSV header row (10 columns):

```
Order No, Order Items, Shipping method, Shipping Cost, Payment method, Order Total, Earnings, Order Status, Order Date, Customer Note
```

Setting off is unchanged at all 33 columns.

### Note for reviewers

**On the `'off' === $opt` vs `'on' !== $opt` convention.** Deliberately kept as `'off' === $opt`, matching `templates/orders/details.php`, rather than converging on Pro's `dokan_maybe_hide_customer_info()` spelling. For a value that is neither `on` nor `off`, `'on' !== $opt` would export the customer data, while this fails closed and withholds it. On a privacy control the safe direction is to withhold, and an unexpected missing column is an annoyance where an unexpected present column is a leak. There is now a comment saying so. Happy to flip it if the team would rather have the three call sites agree even at that cost — the three genuinely do disagree today and that is worth its own cleanup.

Two adjacent gaps found while investigating, both out of scope here:

* `includes/REST/OrderController.php` returns `billing`, `shipping`, `customer_ip_address` and `customer_note` on the vendor-scoped `/dokan/v1/orders` endpoint with no `hide_customer_info` check. The setting is a UI curtain rather than an access control, and the React vendor dashboard reads that endpoint — worth its own ticket.
* Pro defines a shared gate for this setting, `dokan_maybe_hide_customer_info()` with its `dokan_should_render_customer_info` filter, but the export (in Lite) re-reads the raw option instead. A site overriding that filter gets it honoured on the details page and in emails but not in the CSV. Routing the export through the same gate would be the deeper fix; it crosses the Lite/Pro boundary, so it is left for a follow-up.
* `src/dashboard/orders/OrderList.tsx` sends `customer_id` as a POST field while `Dashboard\Templates\Orders::handle_order_export()` reads it from `$_GET`, so the React dashboard's **Export Filtered** silently ignores the customer filter. Pre-existing and unrelated to this change.

### DIFF
--- a/includes/Order/MiscHooks.php
+++ b/includes/Order/MiscHooks.php
@@ -22,8 +22,8 @@ class MiscHooks {
      * @since 3.8.0
      */
     public function __construct() {
-        // remove customer info from order export based on setting
-        add_filter( 'dokan_csv_export_headers', [ $this, 'hide_customer_info_from_vendor_order_export' ], 20, 1 );
+        // Remove customer info from order export based on setting. Runs last so columns any other callback adds are covered too.
+        add_filter( 'dokan_csv_export_headers', [ $this, 'hide_customer_info_from_vendor_order_export' ], PHP_INT_MAX, 1 );
 
         add_filter( 'woocommerce_rest_prepare_shop_order_object', [ $this, 'add_vendor_info_in_rest_order' ], 10, 1 );
         add_filter( 'wp_count_posts', [ $this, 'modify_vendor_order_counts' ], 10, 1 ); // no need to add hpos support for this filter
@@ -32,11 +32,8 @@ class MiscHooks {
     /**
      * Remove customer sensitive information while exporting order
      *
-     * Every `billing_*` and `shipping_*` column plus `customer_ip` counts as customer data and is
-     * dropped when the setting is on. `customer_note` stays, matching the vendor order details page.
-     * Matching by prefix covers columns added later without touching this method, but this callback
-     * runs at priority 20, so columns registered on a later priority are not seen. Use the filter
-     * below to adjust the list either way.
+     * `customer_note` is kept, matching the vendor order details page. `shipping_*` is not always
+     * customer data, so the columns this drops are filterable.
      *
      * @since 3.8.0 Moved this method from Order/Hooks.php file
      *
@@ -52,13 +49,16 @@ class MiscHooks {
             return $headers;
         }
 
-        $hidden_columns = array_filter(
-            array_keys( $headers ),
-            function ( $column_key ) {
-                return 'customer_ip' === $column_key
-                    || str_starts_with( $column_key, 'billing_' )
-                    || str_starts_with( $column_key, 'shipping_' );
-            }
+        $hidden_columns = array_keys(
+            array_filter(
+                $headers,
+                function ( $column_key ) {
+                    return 'customer_ip' === $column_key
+                        || str_starts_with( $column_key, 'billing_' )
+                        || str_starts_with( $column_key, 'shipping_' );
+                },
+                ARRAY_FILTER_USE_KEY
+            )
         );
 
         /**
--- a/includes/Order/MiscHooks.php
+++ b/includes/Order/MiscHooks.php
@@ -22,8 +22,8 @@ class MiscHooks {
      * @since 3.8.0
      */
     public function __construct() {
-        // Remove customer info from order export based on setting. Runs late so columns other callbacks add are covered too.
-        add_filter( 'dokan_csv_export_headers', [ $this, 'hide_customer_info_from_vendor_order_export' ], 999 );
+        // Runs past the conventional 999 so columns registered there are covered too.
+        add_filter( 'dokan_csv_export_headers', [ $this, 'hide_customer_info_from_vendor_order_export' ], 9999 );
 
         add_filter( 'woocommerce_rest_prepare_shop_order_object', [ $this, 'add_vendor_info_in_rest_order' ], 10, 1 );
         add_filter( 'wp_count_posts', [ $this, 'modify_vendor_order_counts' ], 10, 1 ); // no need to add hpos support for this filter
@@ -32,8 +32,7 @@ class MiscHooks {
     /**
      * Remove customer sensitive information while exporting order
      *
-     * `customer_note` is kept, matching the vendor order details page. `shipping_*` is not always
-     * customer data, so the columns this drops are filterable.
+     * `customer_note` is kept, matching the vendor order details page.
      *
      * @since 3.8.0 Moved this method from Order/Hooks.php file
      *
@@ -44,7 +43,7 @@ class MiscHooks {
     public function hide_customer_info_from_vendor_order_export( $headers ) {
         $hide_customer_info = dokan_get_option( 'hide_customer_info', 'dokan_selling', 'off' );
 
-        // Fail closed: the setting is a switcher, so anything other than an explicit off keeps customer data out of the export.
+        // Fail closed: only an explicit off lets customer data into the export.
         if ( 'off' === $hide_customer_info ) {
             return $headers;
         }
@@ -69,7 +68,10 @@ class MiscHooks {
          * @param array $hidden_columns Column keys to remove from the export.
          * @param array $headers        All export column keys mapped to their labels.
          */
-        $hidden_columns = apply_filters( 'dokan_hidden_customer_info_csv_columns', $hidden_columns, $headers );
+        $filtered_columns = apply_filters( 'dokan_hidden_customer_info_csv_columns', $hidden_columns, $headers );
+
+        // A malformed return falls back to the computed list rather than exporting the data it was meant to hide.
+        $hidden_columns = is_array( $filtered_columns ) ? array_filter( $filtered_columns, 'is_string' ) : $hidden_columns;
 
         return array_diff_key( $headers, array_flip( $hidden_columns ) );
     }
--- a/includes/Order/MiscHooks.php
+++ b/includes/Order/MiscHooks.php
@@ -32,6 +32,8 @@ class MiscHooks {
     /**
      * Remove customer sensitive information while exporting order
      *
+     * Drops every billing and shipping column plus the customer IP when the setting is on.
+     *
      * @since 3.8.0 Moved this method from Order/Hooks.php file
      *
      * @param array $headers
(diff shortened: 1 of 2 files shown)


MdAsifHossainNadim:
## Verdict: Approve ✅

Re-reviewed through `4f22795c5`. All three points from the first pass are resolved, and the follow-up commits went further than asked. One thing the latest commit introduces is worth a look before merge — see follow-up 1.

### What changed

- **`dokan_hidden_customer_info_csv_columns` filter** added as a documented escape hatch — `shipping_*` is not always customer data, so the drop list has to stay adjustable.
- **Hook priority `20` → `PHP_INT_MAX` → `999`**, so columns registered by other callbacks are covered too.
- **New test file** `tests/php/src/Orders/HideCustomerInfoExportTest.php` — 5 tests: setting off, setting on, third-party columns dropped by prefix (default priority and a later one), columns filtered back in, and the filter receiving a plain list.
- Docblock corrected (`@return mixed` → `@return array`), `$column` → `$column_key`, redundant `accepted_args` dropped, and the `'off' ===` check now carries an explicit fail-closed rationale.

### Verified

- `order_shipping` / `order_shipping_cost` carry the `order_` prefix, so shipping method and cost survive the filter. Counts check out: 33 columns → 10 with the setting on.
- Rows can't misalign — `dokan_order_csv_export()` builds each line from the *filtered* headers.
- `str_starts_with()` is safe on the PHP 7.4 floor (WP core polyfills it since 5.9; Dokan requires WP 6.9).
- Blast radius is vendor-only: the filter's consumer chain ends at `Dashboard\Templates\Orders::handle_order_export()`. No admin export path, and nothing else in Lite or Pro hooks `dokan_csv_export_headers`.
- Both the legacy and React dashboards post to that same handler, so the test plan covers both.
- Test scaffolding is sound: `MiscHooks` is constructed eagerly at boot (`dokan-class.php:325`), `WP_UnitTestCase` restores `$wp_filter` between tests, and `hide_customer_info` isn't remapped by `dokan_admin_settings_rearrange_map()`, so writing `dokan_selling` directly is correct.

### Follow-ups

**1. Priority `999` collides with the house "run late" idiom.** Moving off `PHP_INT_MAX` is well reasoned, and with the escape-hatch filter in place there's no need for an "I win" priority — agreed on both. But `999` is the specific value that collides: it's already used 7 times in Lite as the "run late" value (`VendorNavMenuChecker`, `ReverseWithdrawal\Order`, `ProductCache`, `Core`), so it's the most likely number an ecosystem developer reaches for. At equal priority WordPress runs callbacks in registration order, and this one registers at plugin load — so a third-party `billing_*` column registered at 999 runs *after* us and survives into the export. That's exactly why the test case had to move from 999 down to 100 in this commit.

Two ways to close it, either is fine:

- Pick a late value that doesn't collide (`9999` is already used in Lite), since the sanctioned opt-out is now the filter and the priority only needs to beat *accidental* registrations; or
- Keep `999` and make the boundary explicit — a test asserting that a `billing_*` column registered at 1000 deliberately survives, plus a line in the docblock. Right now the boundary is encoded only in the test's choice of number, so it's indistinguishable from a bug.

Leaning toward the first: one explicit opt-out beats an explicit one plus a silent one.

**2. `array_flip()` trusts the new filter's return value.** A third-party callback returning `null`, `false` or a string fatals mid-download, after the `Content-Disposition` header is already sent. `array_filter( (array) apply_filters( ... ), 'is_string' )` closes it, and it's free to do while the filter is still new.

**3. The fail-closed behaviour is documented but untested.** A case for a non-canonical value (`''`, `'yes'`, `1`) would lock in the contract the comment now asserts. Worth noting that Pro's `dokan_maybe_hide_customer_info()` fails *open* (`'on' !== $opt`), so the asymmetry is deliberate on one side only.

Nit: dropping `accepted_args` is correct, but the two neighbouring `add_filter` calls in the same constructor still spell out `10, 1`, so the file is now inconsistent on the convention the commit cites.

### Before merge

- PHPUnit wasn't run locally (no `vendor/` in the checkout; `composer install` fires the Mozart hook that rewrites tracked `lib/`). CI hasn't run on any of the four commits — worth a re-trigger so the suite is on record.
- PR has no labels yet. Suggested: `Type: Bug`, `[Orders]`, `[Vendor Dashboard]`, `focus: order`.

MdAsifHossainNadim:
## Verdict: ✅ Approved

Re-reviewed through `c26d9b034`.

### What we did

With **Hide Customer Info** enabled, the vendor order CSV previously dropped only `billing_email` and `customer_ip` — the customer's name, phone and full billing *and* shipping address still shipped in the file, contradicting the setting the order details page already honours. The export now drops every `billing_*` / `shipping_*` column plus `customer_ip`, matching prefix rather than a hardcoded list so columns added later (by core, Pro, or a third party) are covered without touching this method. `customer_note` stays, matching the details page; `order_shipping` / `order_shipping_cost` carry the `order_` prefix and survive as marketplace data. `dokan_hidden_customer_info_csv_columns` is the escape hatch for a legitimate `shipping_*` column such as a tracking number.

Three follow-ups from the first review round are closed in `c26d9b034`:

- **Priority `999` → `9999`.** `999` is the house "run late" value (6 uses in Lite) and `MiscHooks` registers at plugin boot, so at equal priority WordPress ran a third party's callback *after* ours and its `billing_*` column survived into the export.
- **`array_flip()` no longer trusts the filter.** A callback returning `null`/`false`/a string fataled mid-download, after `Content-Disposition` was sent. Worth noting a plain `(array)` cast would have been *worse* than the fatal — `(array) null` is an empty list, so a malformed return would have exported every customer column. It now falls back to the computed list, matching the fail-closed direction of the setting check above it.
- **The fail-closed contract is now asserted**, not just described in a comment.

### Test scopes

`tests/php/src/Orders/HideCustomerInfoExportTest.php` covers: setting off leaves all 33 columns; any hiding value (`on`, `''`, `yes`, `1`) leaves exactly the 10 non-customer columns; third-party `billing_*`/`shipping_*` columns dropped at the default priority **and at 999**, while an unrelated column survives; columns filterable back in; the filter payload arriving as a plain list; malformed filter returns still hiding; and an array with a junk entry honouring its valid entries.

Verified against the real stack (WP + WooCommerce + Dokan, real orders), both directions:

| Check | Result |
| --- | --- |
| Column count, setting off → on | 33 → 10, no row misalignment either way |
| Rendered CSV, setting on | zero billing/shipping/IP headers present |
| Third-party column at priority 10 / 999 | both dropped |
| Column registered at 9999 | survives — the deliberate opt-out |
| Vendor dashboard, orders list + order details | render clean in both states, no console errors, customer block hidden on / restored off |

**Before merge:** the PHPUnit suite has still never executed against this file — Lite has no PR-triggered CI, and Docker was unavailable locally, so the behaviour above was verified through the real stack rather than through a test runner. Please run `npm run test:phpunit` once with Docker up so the suite is on record.

Two optional items left deliberately for the author: switching the filter contract from a 0-based list to a keyed subset of `$headers` (would delete the `array_flip` and the `is_string` scrub — free now, breaking after release), and `protected $is_unit_test = true;` on the test class (no test class in the repo currently takes that branch, so it wants a green run first).

PR #3370, opened 2026-08-11
fix(product-editor): enforce the required "Downloadable Files" field

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

A vendor can submit a Virtual + Downloadable product with **no downloadable file attached**, even when *Dokan → Product Form Manager → Downloadable Options → Downloadable Files* is toggled **Required**. The product saves and lands in the review queue with `downloadable = yes` and 0 files.

Two separate gaps let it through:

**1. Blank repeater rows counted as a value (client side)**

`FileUploadEdit` renders one empty row up front and pushes every row it holds — blank ones included — into the form data. The required check in `isEmpty()` only asked whether the array had entries:

```ts
if ( Array.isArray( v ) ) {
    return v.length === 0;
}
```

So the field starts correctly invalid (`downloads: []`), but the moment the vendor clicks **Add New** or types into a row, `downloads` becomes `[ { id: '', name: '', file: '' } ]` — length 1 — the rule passes and **Save Changes** unlocks.

Two changes close it:

- `isEmpty()` now recurses, so a collection whose entries are all themselves empty is empty. There were three copies of this predicate in the module — `isEmpty` in `field-config/validations`, and `isEmptyValue` in `utils.tsx` backing the `empty`/`not_empty` **dependency** comparisons — so they collapse into one exported `isEmpty()` in `utils.tsx`; otherwise a blank download row would be "empty" to validation and "not empty" to dependency resolution at the same time. Recursion is limited to plain objects, because `Object.values( new Date() )` is `[]` and a `Date` must not read as no value.
- `FileUploadEdit` now hands the form only the rows that carry a file, while still rendering every row. It is the one place that knows `file` is the meaningful key, so a row the vendor has named but not attached a file to no longer satisfies the rule — previously that passed the browser check and died at the server with a 400. Form data now also matches what WooCommerce persists, since `save_downloadable_files()` discards file-less rows anyway.

**2. Nothing enforced the required flags server side**

`ProductControllerV3` handed the payload straight to WooCommerce, which silently drops file rows with no `file` — so the product saved as downloadable with zero downloads.

`validate_required_downloads()` now rejects such a save before the product is written. It hangs off a single `prepare_object_for_database()` override rather than a guard in each endpoint — WooCommerce routes create, update and batch alike through `save_object()`, so one override covers all three.

**It judges the assembled product, not the payload.** WooCommerce turns a product downloadable from `downloadable` alone and only reads `downloads` when that key is present, so a payload-shaped check is trivially dodged — `POST /dokan/v3/products {"downloadable":true}` with no `downloads` key returned `201` with `downloadable = yes` and zero files. Checking `$product->is_downloadable() && ! $product->get_downloads()` after `parent::prepare_object_for_database()` closes that, and because WooCommerce has already dropped file-less rows by then, the same single check covers blank rows and name-only rows too. Nothing is persisted until `save_object()` calls `save()` on what the method returns.

Scope still keys off the request — a save touching neither `downloads` nor `downloadable` (quick edit, batch status changes) is out of scope, so products that predate the rule stay editable.

The rule is read from the form schema, so it follows whatever the Product Form Manager is set to — including per-product-type `requireds`/`visibilities` and a renamed field label — and stays inert while the field is optional. Per-type resolution lives on `FormSchema::get_label()` / `::is_required()` / `::is_visible()`, the PHP counterparts of `resolveLabel` / `resolveRequired` / `resolveVisibility` in `utils.tsx`.

**Extension point:** `prepare_object_for_database()` now ends with `dokan_rest_pre_insert_product_object`, following the convention already used by `OrderController`, `CustomersController` and Pro's Coupon / Variation / ManualOrders controllers. Returning a `WP_Error` rejects the save, so extensions that inject required fields into the product form can enforce them without patching this controller. It mirrors WooCommerce's `woocommerce_rest_pre_insert_product_object` — which Pro's simple-auction and subscription modules already use for this job — but is scoped to vendor dashboard saves.

**Why this enforces one field and not every required field:** the schema marks several fields required that a partial form never renders yet still submits empty — `description` is required by default and the quick-create modal omits it while `initForm` still seeds and POSTs it as `''` — so a blanket "all required fields" check would reject every quick create. That constraint is recorded in the method's docblock so the next reader doesn't "fix the altitude" and break product creation.

### Closes

* Closes getdokan/dokan-pro#5963
* Closes https://github.com/getdokan/plugin-internal-tasks/issues/2223
* Client report: getdokan/client-issue#509

1. As admin, go to **Dokan → Product Form Manager → Downloadable Options** and turn **Required** on for **Downloadable Files**.
2. Log in as a vendor and open **Products → Add new product** (new React editor).
3. Fill in title, description, price and category, then tick **Downloadable** and **Virtual**.
4. Leave **Downloadable Files** empty and click **Add New** once or twice to add blank rows.
   * **Save Changes stays disabled**, the section shows *1 field needs attention* and the field shows *Please fill out this field.*
5. Type a **name** into a row but no file → the name stays put and **Save Changes is still disabled**.
6. Type or choose a file URL in that row → the error clears, **Save Changes** enables, and the product saves with the download attached.
7. Server side, `PUT dokan/v3/products/<id>` with `downloadable: true` and `downloads: []`, rows whose `file` is empty, or **no `downloads` key at all**, returns `400 dokan_rest_product_download_required`.
8. Regression checks: turn **Required** back off — a downloadable product saves with no file again; quick create and a partial `PUT` that omits `downloads` keep working.

Verified against a live install (Lite 5.0.12 / Pro Business 5.0.11, WC 10.x):

| case | result |
|---|---|
| no rows / one blank row / name-only row | `400 dokan_rest_product_download_required` |
| `downloadable: true` with **no `downloads` key** — create and update | `400` (was `201` / `200` before this) |
| same via `POST /products/batch` | rejected per item |
| file attached | 200, saved with 1 download |
| not downloadable · partial `PUT` without `downloads` · quick-create draft | 200 / 201 |
| field left optional, or hidden for this product type | 200 |
| product that predates the rule: price-only or status-only save | 200 |
| product that predates the rule: turning `downloadable` off | 200 |

Client side: `[]`, blank rows and a **row carrying only a name** all keep **Save Changes** disabled; typing that name stays visible in the row while the form data stays empty, and adding the file enables Save and persists one download.

Blast radius of the shared `isEmpty()` change was measured rather than assumed: old vs new implementation diffed over every array/object-valued schema field across 40 real products (14 fields, 28 distinct values) — **0 behaviour changes**. The only differences are `[ { …all blank } ]`, `[ '' ]` and `{}`, none of which any Lite or Pro field produces.

### Changelog entry

*Fixed: required "Downloadable Files" field was not enforced*

Vendors could submit a downloadable product without attaching any file even though the Product Form Manager marked Downloadable Files as required, so incomplete downloadable products reached the review queue. Empty file rows no longer count as a filled-in value, and the requirement is now enforced on the server as well as in the browser.

### Before Changes

With **Downloadable Files** required and left empty, clicking **Add New** cleared the validation error and enabled **Save Changes**. The product saved with `downloadable = yes` and `downloads = 0`.

### After Changes

Blank rows no longer satisfy the requirement — **Save Changes** stays disabled and the field keeps showing *Please fill out this field.* until a file is attached. A crafted REST save is rejected with `400 dokan_rest_product_download_required`.

### PR Self Review Checklist:

* [x] Code is following code style guidelines
* [x] Have not left commented out code
* [x] Have not left any debug/console statements

### DIFF
--- a/includes/REST/ProductControllerV3.php
+++ b/includes/REST/ProductControllerV3.php
@@ -359,9 +359,9 @@ class ProductControllerV3 extends WC_REST_Products_Controller {
          *
          * @since DOKAN_SINCE
          *
-         * @param WC_Data|WP_Error $product  Product assembled from the request, not yet saved.
-         * @param WP_REST_Request  $request  Full details about the request.
-         * @param bool             $creating Whether a new product is being created.
+         * @param WC_Data         $product  Product assembled from the request, not yet saved.
+         * @param WP_REST_Request $request  Full details about the request.
+         * @param bool            $creating Whether a new product is being created.
          */
         return apply_filters( "dokan_rest_pre_insert_{$this->post_type}_object", $product, $request, $creating );
     }
@@ -383,6 +383,9 @@ class ProductControllerV3 extends WC_REST_Products_Controller {
      * enforcing every required field here would reject those saves outright. Extensions that need their own
      * rules can hook `dokan_rest_pre_insert_product_object`.
      *
+     * Product variations are out of reach here: WooCommerce saves them through WC_REST_Product_Variations_Controller,
+     * which never enters this controller, so a downloadable variation is guarded in the browser only.
+     *
      * @since DOKAN_SINCE
      *
      * @param WP_REST_Request $request Full details about the request.
@@ -398,15 +401,20 @@ class ProductControllerV3 extends WC_REST_Products_Controller {
             return null;
         }
 
-        if ( ! $product->is_downloadable() || $product->get_downloads() ) {
+        if ( ! $product->is_downloadable() ) {
             return null;
         }
 
-        // Only a save with nothing attached needs the schema, which is expensive to build.
+        // WooCommerce drops rows on empty() alone, so a whitespace-only file still arrives here - and still saves when the field is optional.
+        foreach ( $product->get_downloads() as $download ) {
+            if ( '' !== trim( (string) $download->get_file() ) ) {
+                return null;
+            }
+        }
+
+        // Only a save with nothing attached needs the schema, which is expensive to build - and never its resolved values.
         $product_type = $product->get_type();
-        $schema       = dokan()->product_editor->get_schema( $product->get_id() );
-        $matches      = wp_list_filter( $schema, [ 'id' => Elements::DOWNLOADS ] );
-        $field        = reset( $matches );
+        $field        = FormSchema::get_field( dokan()->product_editor->get_schema(), Elements::DOWNLOADS );
 
         if ( ! $field || ! FormSchema::is_required( $field, $product_type ) || ! FormSchema::is_visible( $field, $product_type ) ) {
             return null;
--- a/includes/REST/ProductControllerV3.php
+++ b/includes/REST/ProductControllerV3.php
@@ -324,9 +324,11 @@ class ProductControllerV3 extends WC_REST_Products_Controller {
      * Prepare a single product for create or update.
      *
      * Every save funnels through here - WooCommerce routes create, update and batch alike through
-     * save_object() - so form-level validation is hooked once here instead of per endpoint.
+     * save_object() - so form-level validation runs once here instead of per endpoint, and it judges
+     * the product WooCommerce assembled rather than the raw payload. Nothing is persisted until
+     * save_object() calls save() on what this returns.
      *
-     * @since 5.0.13
+     * @since DOKAN_SINCE
      *
      * @param WP_REST_Request $request  Full details about the request.
      * @param bool            $creating Whether a new product is being created.
@@ -334,7 +336,34 @@ class ProductControllerV3 extends WC_REST_Products_Controller {
      * @return WC_Data|WP_Error
      */
     protected function prepare_object_for_database( $request, $creating = false ) {
-        return $this->validate_required_downloads( $request ) ?? parent::prepare_object_for_database( $request, $creating );
+        $product = parent::prepare_object_for_database( $request, $creating );
+
+        if ( is_wp_error( $product ) ) {
+            return $product;
+        }
+
+        /**
+         * Filters form-level validation of a vendor product save.
+         *
+         * Return a WP_Error to reject the save. Extensions that add required fields to the product
+         * form can enforce them here instead of patching the controller.
+         *
+         * @since DOKAN_SINCE
+         *
+         * @param WP_Error|null   $error    Error to reject the save with, null to allow it.
+         * @param WC_Data         $product  Product assembled from the request, not yet saved.
+         * @param WP_REST_Request $request  Full details about the request.
+         * @param bool            $creating Whether a new product is being created.
+         */
+        $error = apply_filters(
+            'dokan_rest_product_validate_save',
+            $this->validate_required_downloads( $request, $product ),
+            $product,
+            $request,
+            $creating
+        );
+
+        return is_wp_error( $error ) ? $error : $product;
     }
 
     /**
@@ -344,36 +373,40 @@ class ProductControllerV3 extends WC_REST_Products_Controller {
      * read back from the schema rather than hard-coded. Browser-side validation alone can be bypassed, which
      * is how incomplete downloadable products reach the review queue.
      *
+     * Judged on the assembled product rather than on the payload, because the payload can be shaped to dodge
+     * the rule: WooCommerce turns a product downloadable from `downloadable` alone and only reads `downloads`
+     * when that key is sent, and it drops file-less rows in save_downloadable_files(). A save that leaves
+     * both fields alone, such as quick edit, stays out of scope.
+     *
      * Deliberately scoped to this one field: the schema marks several fields required that a partial form
      * never renders yet still submits empty - the quick-create modal omits the required Description - so
-     * enforcing every required field here would reject those saves outright.
+     * enforcing every required field here would reject those saves outright. Extensions that need their own
+     * rules can hook `dokan_rest_product_validate_save`.
      *
-     * @since 5.0.13
+     * @since DOKAN_SINCE
      *
      * @param WP_REST_Request $request Full details about the request.
+     * @param WC_Data         $product Product assembled from the request, not yet saved.
      *
      * @return WP_Error|null Error when the required file is missing, null otherwise.
      */
-    protected function validate_required_downloads( $request ) {
-        $downloads = $request->get_param( Elements::DOWNLOADS );
+    protected function validate_required_downloads( $request, $product ) {
+        $touches_downloads = null !== $request->get_param( Elements::DOWNLOADS )
+            || null !== $request->get_param( Elements::DOWNLOADABLE );
 
-        // A save that leaves the downloadable fields alone, such as quick edit, is none of our business.
-        if ( ! is_array( $downloads ) || ! dokan_string_to_bool( $request->get_param( Elements::DOWNLOADABLE ) ) ) {
+        if ( ! $touches_downloads || ! $product instanceof WC_Product ) {
             return null;
         }
 
-        foreach ( $downloads as $download ) {
-            if ( ! empty( $download['file'] ) ) {
-                return null;
-            }
+        if ( ! $product->is_downloadable() || $product->get_downloads() ) {
+            return null;
         }
 
         // Only a save with nothing attached needs the schema, which is expensive to build.
-        $product_type = $request->get_param( Elements::TYPE );
-        $product_type = is_string( $product_type ) && '' !== $product_type ? $product_type : Elements::PRODUCT_TYPE_SIMPLE;
-
-        $schema = dokan()->product_editor->get_schema( (int) $request->get_param( 'id' ) );
-        $field  = current( wp_list_filter( $schema, [ 'id' => Elements::DOWNLOADS ] ) );
+        $product_type = $product->get_type();
+        $schema       = dokan()->product_editor->get_schema( $product->get_id() );
+        $matches      = wp_list_filter( $schema, [ 'id' => Elements::DOWNLOADS ] );
+        $field        = reset( $matches );
 
         if ( ! $field || ! FormSchema::is_required( $field, $product_type ) || ! FormSchema::is_visible( $field, $product_type ) ) {
             return null;
--- a/includes/REST/ProductControllerV3.php
+++ b/includes/REST/ProductControllerV3.php
@@ -325,8 +325,10 @@ class ProductControllerV3 extends WC_REST_Products_Controller {
      *
      * Every save funnels through here - WooCommerce routes create, update and batch alike through
      * save_object() - so form-level validation runs once here instead of per endpoint, and it judges
-     * the product WooCommerce assembled rather than the raw payload. Nothing is persisted until
-     * save_object() calls save() on what this returns.
+     * the product WooCommerce assembled rather than the raw payload. The product itself is not written
+     * until save_object() calls save() on what this returns, though the parent call can already have
+     * created attachment records when a payload sends images by URL; the vendor form only ever sends
+     * attachment ids.
      *
      * @since DOKAN_SINCE
      *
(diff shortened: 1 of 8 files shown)


MdAsifHossainNadim:
## ✅ Approve

Verified locally against `bad26ef8b` — PHPUnit `dokan-product-editor` 14/14 and `dokan-product-controller-v3` 20/20 green, `composer phpcs` clean, no new ESLint errors vs `develop`.

### What was broken

A vendor could save a Virtual + Downloadable product with **no file attached** even with *Downloadable Files* set to Required — two independent gaps:

1. **Client:** blank repeater rows counted as a value, so clicking **Add New** unlocked **Save Changes**.
2. **Server:** nothing enforced the required flag at all.

### How it's fixed

- **`isEmpty()` now recurses**, so a list of all-blank rows reads as empty. The three copies of that predicate collapsed into one shared helper, so validation and dependency resolution can't disagree.
- **`FileUploadEdit` hands the form only rows that carry a file**, so the browser check and the server judge the same list — no more "typed a name, Save unlocked, server said no".
- **`ProductControllerV3::prepare_object_for_database()` rejects a downloadable save with no file.** It judges the *product WooCommerce assembled*, not the raw payload — which is what closes the payload-shaped holes: `{"downloadable": true}` with no `downloads` key, `{"downloads": []}` with no `downloadable`, and the create path all reach the same verdict. One override covers create, update **and** batch.
- **The rule reads the form schema** via `FormSchema::is_required()` / `::is_visible()` / `::get_label()`, so it follows the Product Form Manager, honours per-product-type `requireds`/`visibilities` and a renamed label, and stays inert while the field is optional. Quick edit and partial saves are untouched.
- **Extensions get the conventional `dokan_rest_pre_insert_product_object` seam**, matching `OrderController` / `CustomersController` and mirroring WooCommerce's own filter.

### Addressed during review

Hardcoded `@since 5.0.13` → `DOKAN_SINCE` · the three payload-shaped bypasses · a false "nothing is persisted" claim in the docblock (`set_product_images()` does write) · hook renamed to the house convention before release. Test coverage went **4 → 14** cases.

### Non-blocking follow-ups

- A whitespace-only `file` value still counts as an attached file (WC skips rows only on `empty()`).
- `get_schema()` costs ~6 queries / ~24 ms on the reject path; the guard reads only definition keys, so a `$with_values = false` flag would zero that out.
- `get_label()` and the three resolvers have no direct unit tests.
- Product **variations** go through a different controller, so the guard doesn't cover them.
- Lite CI has no PHPUnit job — these 14 tests won't run automatically on future PRs. Worth raising separately.

includes/REST/ProductControllerV3.php
```diff
@@ -320,6 +320,112 @@ public function update_item( $request ) {
...
+     */
+    protected function validate_required_downloads( $request, $product ) {
+        $touches_downloads = null !== $request->get_param( Elements::DOWNLOADS )
+            || null !== $request->get_param( Elements::DOWNLOADABLE );
+
+        if ( ! $touches_downloads || ! $product instanceof WC_Product ) {
+            return null;
+        }
+
+        if ( ! $product->is_downloadable() || $product->get_downloads() ) {
```
kzamanbd: **WARNING — variations bypass this rule.**

`FormSchema::get_schema()` builds `$digital_field_visibilities` with `PRODUCT_TYPE_VARIATION => true`, so the form marks Downloadable Files visible — and potentially required — for variations. But WooCommerce V3 doesn't save variations through the products controller: grepping `class-wc-rest-products-controller.php` for `variations` turns up only the schema entry and the "To manipulate product variations you should use the `/products/<product_id>/variations/<id>` endpoint" rejection at `:763`. Variation saves route through `WC_REST_Product_Variations_Controller`, which never reaches this method.

So a downloadable **variation** saved with no file still goes through, with the browser check as the only guard — the same shape as the bug this PR closes.

Not a blocker, and the hook convention you introduced hands you the follow-up for free: Pro's variation controller would emit `dokan_rest_pre_insert_product_variation_object` under the same `{$this->post_type}` pattern. Suggest either naming the boundary here in the docblock, next to the existing "one field, not every field" note, or opening a follow-up so it doesn't get rediscovered as a bug later.

includes/REST/ProductControllerV3.php
```diff
@@ -320,6 +320,112 @@ public function update_item( $request ) {
...
+
+        // Only a save with nothing attached needs the schema, which is expensive to build.
+        $product_type = $product->get_type();
+        $schema       = dokan()->product_editor->get_schema( $product->get_id() );
+        $matches      = wp_list_filter( $schema, [ 'id' => Elements::DOWNLOADS ] );
+        $field        = reset( $matches );
+
+        if ( ! $field || ! FormSchema::is_required( $field, $product_type ) || ! FormSchema::is_visible( $field, $product_type ) ) {
+            return null;
+        }
```
kzamanbd: **SUGGESTION — this can match a section, not just the field.**

`get_schema()` returns a flat list of **sections and fields sharing a single `id` namespace**, so a section — or an item an extension adds through `dokan_product_editor_prepared_schema` — that happens to use the id `downloads` would be picked up by `reset()` instead of the real field. A section carries no `requireds` map, so `is_required()` returns `false` and the rule silently switches itself off. Fails open, and quietly.

Cheap to close:

```php
$matches = wp_list_filter( $schema, [ 'id' => Elements::DOWNLOADS, 'type' => 'field' ] );
```

includes/REST/ProductControllerV3.php
```diff
@@ -320,6 +320,112 @@ public function update_item( $request ) {
...
+        /**
+         * Filters a vendor product prepared for the database, before it is saved.
+         *
+         * Scoped to vendor dashboard saves, unlike WooCommerce's own
+         * `woocommerce_rest_pre_insert_product_object`. Return a WP_Error to reject the save, which
+         * is how extensions enforce required fields they add to the product form.
+         *
+         * @since DOKAN_SINCE
+         *
+         * @param WC_Data|WP_Error $product  Product assembled from the request, not yet saved.
```
kzamanbd: **SUGGESTION** — `@param WC_Data|WP_Error $product` overstates what actually arrives here. Both error paths return before the filter runs (`is_wp_error( $product )` at `:343`, `is_wp_error( $invalid )` at `:349`), so the value passed in is always `WC_Data`. `@param WC_Data $product` reads truer, and the "return a `WP_Error` to reject the save" contract is already spelled out in the prose just above.

includes/REST/ProductControllerV3.php
```diff
@@ -320,6 +320,112 @@ public function update_item( $request ) {
...
+        $schema       = dokan()->product_editor->get_schema( $product->get_id() );
+        $matches      = wp_list_filter( $schema, [ 'id' => Elements::DOWNLOADS ] );
+        $field        = reset( $matches );
+
+        if ( ! $field || ! FormSchema::is_required( $field, $product_type ) || ! FormSchema::is_visible( $field, $product_type ) ) {
+            return null;
+        }
+
+        $label = FormSchema::get_label( $field, $product_type );
+        $label = '' !== $label ? $label : __( 'Downloadable Files', 'dokan-lite' );
```
kzamanbd, line 424: **SUGGESTION (low)** — `$label` is an admin-configured string from Pro's Product Form Manager, and it goes into the `WP_Error` message unescaped. Admin-only input, so the exposure is small, but if any client ever renders that message as HTML it becomes an injection path. `esc_html( $label )` costs nothing here.

includes/REST/ProductControllerV3.php
```diff
@@ -320,6 +320,112 @@ public function update_item( $request ) {
...
+     * @param bool            $creating Whether a new product is being created.
+     *
+     * @return WC_Data|WP_Error
+     */
+    protected function prepare_object_for_database( $request, $creating = false ) {
+        $product = parent::prepare_object_for_database( $request, $creating );
+
+        if ( is_wp_error( $product ) ) {
+            return $product;
+        }
```
kzamanbd, line 345: Noting the orphan-attachment trade-off you already called out in the docblock, just so it stays a conscious call rather than something rediscovered later: the parent runs first, and `wc_rest_upload_image_from_url()` can create attachment records before the rejection. Unreachable from the vendor form since it only sends ids, but a crafted payload combining `images[].src` with a missing download leaves an orphan attachment behind on every `400`. Fine as documented — no change requested.

src/dashboard/product-editor/utils.tsx
```diff
@@ -69,16 +69,41 @@ export function fieldValueForProduct( item: FormItem ): any {
...
     if ( typeof val === 'string' ) {
         return val.trim() === '';
     }
     if ( Array.isArray( val ) ) {
-        return val.length === 0;
+        return val.every( isEmpty );
+    }
+    if ( isPlainObject( val ) ) {
+        return Object.values( val ).every( isEmpty );
     }
```
kzamanbd: **SUGGESTION** — the recursion has no cycle guard and no depth cap. Not reachable from JSON-derived form data today, but `exports.ts:29` now puts `isEmpty` on the public `@dokan/product-editor` surface, where an extension can hand it an arbitrary object graph and get a stack overflow instead of a boolean. A `WeakSet` of visited objects, or a simple depth cap, is cheap insurance for a function that's now API.

Separately, worth one more docblock line: this also makes `{}` and `[ '' ]` newly empty. The 40-product / 14-field measurement in the description is good evidence for Lite and Pro schema fields, but it can't cover fields a third party injects through `dokan_product_editor_prepared_schema` — saying so explicitly would save the next reader from assuming the blast radius was exhaustive.

src/dashboard/product-editor/exports.ts
```diff
@@ -26,6 +26,7 @@ export { default as CustomField, getValidationError } from './components/CustomF
 // Utils
 export {
     getField,
+    isEmpty,
```
kzamanbd, line 29: Good call exporting this rather than leaving Pro to reimplement the predicate — it's the piece that keeps validation and dependency resolution from disagreeing about what "no value" means. Just flagging that it makes `isEmpty` public API, which is what motivates the recursion-guard note over in `utils.tsx`.

kzamanbd (approved):
Approving. Strong PR — the diagnosis, the layering and the test coverage all hold up under checking. Findings inline are minor; one scope gap worth a decision from you, but not a reason to hold this.


- **The single override really does cover create, update and batch.** `create_item():283` and `update_item():308` both call `parent::`, which routes through `save_object()` → `prepare_object_for_database()`, and `save_object()` returns the `WP_Error` untouched.
- **A partial `PUT` really is out of scope.** This one nearly isn't: `downloadable` carries `'default' => false` in WooCommerce's schema (`class-wc-rest-products-controller.php:1354`), which would have made `$touches_downloads` always true. WordPress applies schema defaults only for `CREATABLE` (`wp-includes/rest-api.php:3415`), and `downloads` has no default at all, so the scope check survives. Batch update builds `PUT` sub-requests with no defaults; batch create sets them explicitly. All four paths behave.
- **"WooCommerce discards file-less rows anyway"** — accurate, `save_downloadable_files()` does `if ( empty( $file['file'] ) ) continue;`.
- **PHP ↔ TS resolver parity** — `FormSchema::is_required()` / `is_visible()` / `get_label()` use the same `??` chains and the same fallbacks as `resolveRequired` / `resolveVisibility` / `resolveLabel` in `utils.tsx:20-41`. Genuine counterparts, not approximations.
- **The `isEmpty` consolidation is clean** — no remaining importer of the old `field-config/validations` export anywhere in Lite or Pro.
- **`FileUploadEdit`'s new UI/data divergence is safe.** I went looking for a bug here and didn't find one: I expected the pre-existing re-sync effect at `:26-30` to wipe blank rows now that `publish()` filters them out of form data. It doesn't, because `field.value` comes from `formItems` (the fetched schema) and `useProductEditor.ts:49-64` memoises fields on `[ formItems, currentProductType ]`, while `onChange` writes to the product store. The array reference stays stable, so the effect doesn't re-fire on keystrokes. Worth knowing it rests on that, though — if fields ever get rebuilt from live form data, the name-only row case breaks.

Test file is the right shape: `DokanTestCase`, two `@group`s, two `@covers`, factories with `set_seller_id()`, and 14 tests covering blank rows, missing-key create *and* update, batch, per-type `requireds`/`visibilities`, the extension filter, and three regression cases. `@since DOKAN_SINCE` used correctly throughout.

dev-shahed:
Can’t create another type of product if the downloadable product is required, so it doesn't works now.
One validation should be: if the downloadable product is required by the manager and the user checks the Downloadable checkbox, a file must be uploaded before the product can be saved or created.

@MdAsifHossainNadim bhai.

PR #3375, opened 2026-08-12
fix(dashboard): hide the Downloadable Product Permission section on orders without downloadable products

akzmoudud:
### All Submissions:

* [x] My code is tested

On the vendor dashboard → order details page, the **Downloadable Product Permission** panel rendered on every order, including physical-only orders where it is always empty. This PR renders the panel only when it is relevant:

* the order contains at least one downloadable item (`WC_Order::has_downloadable_item()`), **or**
* the order already has download-permission rows (cheap existence check via the `customer-download` data store, `limit: 1`).

The second condition matters: permissions can exist on an order whose product was later made non-downloadable (or was granted manually) — the section stays visible there so vendors can still see and revoke existing grants.

A new filter `dokan_order_details_show_download_permissions( bool $show, WC_Order $order )` (tagged `@since DOKAN_SINCE`) lets extensions/themes force the section on or off without overriding the template.

Note: WooCommerce core shows its equivalent wp-admin meta box unconditionally; we deliberately deviate for the vendor dashboard since the empty panel confuses vendors selling physical goods, and admins keep the full tool in wp-admin. Also adds a **Download Permission** entry to the `CONTEXT.md` glossary capturing this distinction.

### Closes

* Closes getdokan/dokan-pro#6055
* Discussion: getdokan/dokan-pro#6052

1. As a vendor, open an order containing only physical products → the Downloadable Product Permission section is **not** rendered.
2. Open an order containing a downloadable product → the section renders as before (permission rows + Grant Access).
3. On a downloadable order with granted permissions, edit the product and untick **Downloadable**, then reload the order details → the section is **still** rendered so the existing grants can be revoked.
4. Add `add_filter( 'dokan_order_details_show_download_permissions', '__return_true' );` → the section renders on every order again.

### Changelog entry

*fix(dashboard): hide the Downloadable Product Permission section on orders without downloadable products*

Previously the vendor dashboard order details page always rendered the Downloadable Product Permission section, even for orders with no downloadable products, where it was permanently empty and confusing. Now the section only appears when the order contains a downloadable item or already has download permissions granted; a new `dokan_order_details_show_download_permissions` filter can override the behaviour.

### Before Changes

Every order details page showed the (empty) section — see https://snipboard.io/hQYPI7.jpg

### After Changes

Physical-only orders: section absent. Downloadable orders and orders with existing permissions: unchanged. Verified all three states plus the filter override as a vendor on a local site (template render + browser).

https://claude.ai/code/session_01PHqzbMyKg4rJ4wzkjhMWcc

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -47,6 +47,15 @@ A return of money to the customer, recorded against a Vendor order and mirrored
 up to its Order. A refund created directly on a parent Order is outside the
 model: vendor accounting does not see it.
 
+**Download Permission**:
+A grant of file access to the customer, recorded against an order — created
+automatically when a paid order contains a downloadable product, or granted
+manually by the Vendor or admin. Independent of the order's current items: a
+permission can exist on an order with no downloadable item (granted manually,
+or the product changed since purchase), and a downloadable item confers no
+access until a permission exists.
+_Avoid_: Download access, download link
+
 ### Money
 
 **Commission**:
--- a/templates/orders/details.php
+++ b/templates/orders/details.php
@@ -192,6 +192,34 @@ $hide_customer_info = dokan_get_option( 'hide_customer_info', 'dokan_selling', '
 
             <div class="clear"></div>
 
+            <?php
+            // Permissions can exist without a downloadable item in the order (granted
+            // manually, or the product was made non-downloadable later) — keep the
+            // section visible in that case so existing grants stay manageable.
+            $show_download_permissions = $order->has_downloadable_item();
+
+            if ( ! $show_download_permissions && 0 !== $order->get_id() ) {
+                $existing_download_permissions = WC_Data_Store::load( 'customer-download' )->get_downloads(
+                    array(
+                        'order_id' => $order->get_id(),
+                        'limit'    => 1,
+                        'return'   => 'ids',
+                    )
+                );
+
+                $show_download_permissions = ! empty( $existing_download_permissions );
+            }
+
+            /**
+             * Filters whether the Downloadable Product Permission section is shown on the order details page.
+             *
+             * @since DOKAN_SINCE
+             *
+             * @param bool      $show_download_permissions Whether to render the section.
+             * @param \WC_Order $order                     Order object.
+             */
+            if ( apply_filters( 'dokan_order_details_show_download_permissions', $show_download_permissions, $order ) ) :
+                ?>
             <div class="" style="width: 100%">
                 <div class="dokan-panel dokan-panel-default">
                     <div class="dokan-panel-heading"><strong><?php esc_html_e( 'Downloadable Product Permission', 'dokan-lite' ); ?></strong></div>
@@ -202,6 +230,7 @@ $hide_customer_info = dokan_get_option( 'hide_customer_info', 'dokan_selling', '
                     </div>
                 </div>
             </div>
+            <?php endif; ?>
         </div>
     </div>
 


MdAsifHossainNadim (changes requested):
## Verdict: 🔴 Request changes

The implementation itself is clean and idiomatic, and the parts worth doubting check out: `get_downloads()` genuinely supports `limit` and `return => 'ids'` (defaults `-1` / `'objects'`), so the existence check is as cheap as intended and only runs when `has_downloadable_item()` is false; it mirrors WooCommerce core's own pattern; the `0 !== $order->get_id()` guard matches the existing one in `templates/orders/downloadable.php:10`; and the filter name, `@since DOKAN_SINCE` tag and CONTEXT.md glossary entry all follow convention. No standards or escaping issues.

Three things before this can merge.

---

### 🔴 1. The premise is wrong — that panel is never empty, and this removes a working capability

#6055 says "on those orders the section is always empty," and the PR repeats it. That isn't what the template does. `templates/orders/downloadable.php` renders the permission list **and then, unconditionally**, a toolbar containing a "Search for a downloadable product…" select2 and a working **Grant Access** button (lines 49–68). That search is order-agnostic — it finds any downloadable product, not just items in this order.

So on a physical-only order the vendor today has a functional control for granting a bonus file, or a digital add-on paid for offline. After this PR that control is gone — and because the fallback condition is *"already has permissions"*, the state is unreachable: a vendor can never create the **first** manual grant on such an order. Only the new filter or wp-admin brings it back.

The sharpest evidence is this PR's own glossary entry, which says permissions are "granted manually by the Vendor or admin" and are "Independent of the order's current items." That accurately describes why this gate is too aggressive. It is also the real reason WooCommerce core keeps its metabox unconditional — the PR notes the deviation but frames it as cosmetic, when it actually withdraws a capability.

Worth settling with @anik-fahmid on the issue. Options that keep the win without the loss: render the panel but collapsed / with an empty-state line; hide the permission *list* while keeping Grant Access; or make hiding an explicit setting.

### 🔴 2. This breaks a currently-passing E2E test

`tests/pw/tests/e2e/orders/orders.spec.ts:90` — *"vendor can view order details"* (`@lite @vendor`) — calls `viewOrderDetails()`, which asserts the Downloadable Product Permission panel is **visible** (`ordersPage.ts:725-726`, with a comment explaining which sub-elements were deliberately excluded). The order under test comes from `createTestOrder()` (`ordersPage.ts:396`), which uses `PRODUCT_ID`, seeded from `payloads.createProduct()` → `type: 'simple'`, not downloadable (`payloads.ts:412-413`, `_env.setup.ts:208`).

That is exactly the physical-only case, so the panel will be absent and the assertion will fail. The checklist ticks "My code is tested" but `tests/pw/` is untouched. Either move that assertion into a downloadable-order test or seed a downloadable order for it.

Heads-up: this will **not** show up on the PR — Lite's E2E workflow is daily-cron/manual-dispatch only, so it lands as a nightly failure attributed to whatever ran next.

### 🟠 3. Unfixed twin in Pro

`dokan-pro/modules/vendor-subscription-product/templates/subscription/subscription-details.php:190-198` renders the identical unconditional panel for subscription details (passing `$subscription` as `$order`), with a React equivalent at `SubscriptionBody.tsx:626`. A vendor selling subscriptions still sees the same panel and the same complaint. Worth a companion PR, or at least a tracked follow-up so the fix isn't half-applied.

### Architectural note

The decision block sits inside `templates/orders/details.php`, which is a **user-overridable** template. Two consequences: any theme that already copied this template keeps the old behaviour and silently misses the fix (worth a changelog line), and the logic is effectively frozen for those sites.

Better as a small helper — e.g. `dokan_order_show_download_permissions( WC_Order $order ): bool` in `includes/` — with the template reduced to a one-line call. That makes it unit-testable, keeps the filter in PHP where it belongs, and gives the Pro subscription template something to reuse instead of duplicating twenty lines. If the React order-details redesign lands later, a helper is what it can call; a template-embedded conditional is not.

### Minor

When the section shows via the permissions fallback, `orders/downloadable.php` re-runs the same `get_downloads()` query. Negligible, but passing the result through would avoid it.

akzmoudud:
Closing this PR without merging.

The linked issue (getdokan/dokan-pro#6055) was closed as "not a bug": the Downloadable Product Permission section is rendered on every order by design — it mirrors WooCommerce core, which shows the same meta box on every order in wp-admin so that downloads can be granted manually to any order (e.g. a bonus file on a physical-product order).

Hiding the section based on the order's items would remove that Grant Access capability for vendors, as noted in the review. Since the current behaviour matches WooCommerce and the issue is closed, no change is needed.

Thanks for the work on this, @akzmoudud — the template cleanup and filter idea are worth revisiting if product decides to make the section collapsible in the future.

CC: @anik-fahmid bhai, @MdAsifHossainNadim bhai

PR #3376, opened 2026-08-12
fix(product-editor): paginate async_select fields so all tags/options are reachable

akzmoudud:
### All Submissions:

* [x] My code is tested

Fixes two defects in the new product editor's `async_select` fields (Tags, and Pro's Upsells / Cross-sells / Grouped products), reported in getdokan/dokan-pro#6026:

**1. Options were capped at the first 20 (the reported bug).** `fetchOptions()` in `AsyncSelectEdit.tsx` fetched a single hardcoded `per_page=20` page with no `page` parameter and no scroll handler, so a store with 100+ tags could never browse beyond the first 20. The legacy jQuery form has paginated this dropdown via select2 since v3.0.10; the React rewrite dropped that behavior.

**2. The non-creatable variant could not search at all (discovered while validating).** The debounced function passed as `AsyncSelect`'s `loadOptions` returns the *previous* invocation's promise while a call is pending. react-select therefore attached to a stale, already-resolved promise, rendered "No options", cleared its request token, and discarded the real server response when it arrived 300ms later. With "vendors can create tags" **off**, tag search was completely broken — the server found the tag, the UI threw it away.

**The fix** replaces the per-variant fetch logic with a shared `usePaginatedOptions` hook that explicitly owns option state:

- Page 1 is fetched lazily when the menu first opens, so fields the vendor never opens cost no requests; scrolling the menu list near its end appends the next page for the current search term.
- Pagination is driven by native `scroll` events on the menu list rather than react-select's `onMenuScrollToBottom`. That prop is routed through `ScrollCaptor`, whose `captureMenuScroll` defaults to `!isTouchCapable()` — so it is disabled outright on Android, iOS and touch-screen laptops — and even where it is active it binds only `wheel`/`touchstart`/`touchmove`, meaning scrollbar drag and keyboard navigation never reach it. Listening for `scroll` covers every mechanism. (A `MenuList` override is not usable here: dokan-ui's `TaggableSelect` spreads caller props and then replaces `components` with its own.)
- Clearing the input — blur, menu close, or selecting an option — resets the search, so reopening a field after picking a searched option shows the unfiltered list again rather than one still filtered by the previous term.
- A failed request is reported separately from a genuinely empty last page: the paging state is left intact so the next scroll or menu open retries, and the failure is surfaced to the vendor instead of being indistinguishable from "this store has no options".
- Typing blanks the list immediately so the loading state carries the feedback during the debounce, instead of stale results sitting there looking ignored.
- A new search term resets to page 1 and replaces the list.
- Out-of-order responses are discarded via a request id, and appended pages are deduped by term id.
- `hasMore` is inferred from receiving a full page rather than the `X-WP-TotalPages` header, so proxies that strip custom headers cannot silently disable pagination (cost: one empty request when the total is an exact multiple of 20).
- The non-creatable variant now uses the state-driven `Select` instead of `AsyncSelect`, because `AsyncSelect` owns its options internally and cannot have pages appended — this is also what eliminates the stale-promise search bug.

No backend changes: the WooCommerce/Dokan REST endpoints already supported `page`/`per_page`.

### Closes

* Closes getdokan/dokan-pro#6026

1. Create 100+ product tags (WP Admin → Products → Tags).
2. Log in as a vendor → Vendor Dashboard → Add New Product (new product editor).
3. Open the **Tags** dropdown: the first 20 tags appear, and scrolling the menu keeps loading pages of 20 until every tag is reachable. **Check each scroll mechanism**, since they do not share a code path in react-select: mouse wheel, dragging the menu's scrollbar, keyboard arrow keys, and a touch device (or DevTools device emulation) — touch is the case react-select's own `onMenuScrollToBottom` cannot serve at all.
4. Type a term beyond the first 20 (e.g. the 115th tag's name): it is found and rendered. Repeat with admin setting "vendors can create tags" both on and off.
5. Search for a tag, select it, then reopen the dropdown: the list is back to page 1 unfiltered, not still filtered by the term you typed.
6. With Pro active, verify the Linked Products fields (Upsells / Cross-sells, and Grouped products with product type "Grouped") also paginate on scroll with 20+ products.

### Changelog entry

*fix: Product tag selector did not show all tags in the vendor dashboard product form*

Previously the product editor's tag/linked-product dropdowns loaded only the first 20 options with no way to browse the rest, and (when tag creation was disabled for vendors) typing in the Tags field always showed "No options" even though the tag existed. Now the dropdowns load further pages as the vendor scrolls, and searching works in both configurations.

### Before Changes

Opening the Tags dropdown with 120 tags showed only "Tag 001"–"Tag 020"; scrolling loaded nothing more. With vendor tag creation off, typing "115" fired the server search (which returned Tag 115) but the menu displayed "No options".

### After Changes

Opening the Tags dropdown shows the first 20 tags and wheel-scrolling appends pages of 20 until all 120 are listed; typing "115" renders "Tag 115".

### DIFF
--- a/src/dashboard/product-editor/components/AsyncSelectEdit.tsx
+++ b/src/dashboard/product-editor/components/AsyncSelectEdit.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable import/no-extraneous-dependencies */
-import { AsyncSelect, Select } from '@src/components';
+import { Select } from '@src/components';
 import { TaggableSelect } from '@getdokan/dokan-ui';
 import CustomField, { getValidationError } from './CustomField';
 import { components as rsComponents } from 'react-select';
@@ -9,7 +9,13 @@ import { addQueryArgs } from '@wordpress/url';
 import { debounce } from '@wordpress/compose';
 import { applyFilters } from '@wordpress/hooks';
 import { decodeEntities } from '@wordpress/html-entities';
-import { useEffect, useMemo, useState } from '@wordpress/element';
+import {
+    useCallback,
+    useEffect,
+    useMemo,
+    useRef,
+    useState,
+} from '@wordpress/element';
 
 type FieldProps = {
     data: any;
@@ -36,27 +42,112 @@ const decodeLabel = ( option: any ) =>
 const decodeValue = ( value: any ) =>
     Array.isArray( value ) ? value.map( decodeLabel ) : decodeLabel( value );
 
-// Fetch `{ value, label }` options from an endpoint, optionally filtered by `search`.
+const OPTIONS_PER_PAGE = 20;
+
+// Fetch one page of `{ value, label }` options from an endpoint, optionally
+// filtered by `search`. `hasMore` is inferred from a full page rather than the
+// X-WP-TotalPages header so a proxy stripping custom headers can't disable
+// pagination; the only cost is one empty request when the total is an exact
+// multiple of the page size.
 const fetchOptions = async (
     endpoint?: string,
-    search = ''
-): Promise< any[] > => {
+    search = '',
+    page = 1
+): Promise< { options: any[]; hasMore: boolean } > => {
     if ( ! endpoint ) {
-        return [];
+        return { options: [], hasMore: false };
     }
     try {
         const result: any = await apiFetch( {
-            path: addQueryArgs( endpoint, { search, per_page: 20 } ),
+            path: addQueryArgs( endpoint, {
+                search,
+                per_page: OPTIONS_PER_PAGE,
+                page,
+            } ),
         } );
-        return Array.isArray( result )
+        const options = Array.isArray( result )
             ? result.map( ( item: any ) => ( {
                   value: item.id,
                   label: decodeEntities( item.name ),
               } ) )
             : [];
+        return { options, hasMore: options.length === OPTIONS_PER_PAGE };
     } catch {
-        return [];
+        return { options: [], hasMore: false };
     }
+};
+
+// Options state shared by the non-tree async fields: page 1 replaces the list,
+// scrolling the open menu to its end appends the next page for the current
+// search term, and a new search resets back to page 1.
+const usePaginatedOptions = ( endpoint?: string ) => {
+    const [ options, setOptions ] = useState< any[] >( [] );
+    const [ isLoading, setIsLoading ] = useState( false );
+    // Mutable request state: lets the scroll handler read the latest
+    // page/search without re-rendering, and `requestId` discards responses
+    // that arrive after a newer request has been issued.
+    const stateRef = useRef( {
+        search: '',
+        page: 1,
+        hasMore: false,
+        loading: false,
+        requestId: 0,
+    } );
+
+    const load = useCallback(
+        ( search: string, page: number ) => {
+            const requestId = ++stateRef.current.requestId;
+            stateRef.current.loading = true;
+            setIsLoading( true );
+            fetchOptions( endpoint, search, page ).then( ( result ) => {
+                if ( requestId !== stateRef.current.requestId ) {
+                    return; // superseded by a newer search/page request
+                }
+                stateRef.current = {
+                    ...stateRef.current,
+                    search,
+                    page,
+                    hasMore: result.hasMore,
+                    loading: false,
+                };
+                setIsLoading( false );
+                setOptions( ( previous ) => {
+                    const merged =
+                        page === 1
+                            ? result.options
+                            : [ ...previous, ...result.options ];
+                    // Terms can shift between pages while browsing (e.g. a tag
+                    // created mid-session); keep the first occurrence of each.
+                    const seen = new Set();
+                    return merged.filter(
+                        ( option ) =>
+                            ! seen.has( option.value ) &&
+                            seen.add( option.value )
+                    );
+                } );
+            } );
+        },
+        [ endpoint ]
+    );
+
+    const onSearch = useMemo(
+        () => debounced( ( input: string ) => load( input, 1 ) ),
+        [ load ]
+    );
+
+    const onMenuScrollToBottom = () => {
+        const { search, page, hasMore, loading } = stateRef.current;
+        if ( hasMore && ! loading ) {
+            load( search, page + 1 );
+        }
+    };
+
+    useEffect( () => {
+        load( '', 1 );
+        return () => onSearch.cancel();
+    }, [ load, onSearch ] );
+
+    return { options, isLoading, onSearch, onMenuScrollToBottom };
 };
 
 // Flatten a nested tree into ordered options carrying their depth level.
@@ -185,31 +276,15 @@ const CreatableSelectField = ( {
     onChange,
     validity,
 }: FieldProps ) => {
-    const [ options, setOptions ] = useState< any[] >( [] );
-    const search = useMemo(
-        () =>
-            debounced( ( input ) =>
-                fetchOptions( field.api_endpoint, input ).then( setOptions )
-            ),
-        [ field.api_endpoint ]
-    );
-
-    useEffect( () => {
-        let active = true;
-        fetchOptions( field.api_endpoint ).then(
-            ( opts ) => active && setOptions( opts )
-        );
-        return () => {
-            active = false;
-            search.cancel();
-        };
-    }, [ field.api_endpoint, search ] );
+    const { options, isLoading, onSearch, onMenuScrollToBottom } =
+        usePaginatedOptions( field.api_endpoint );
 
     return (
         <CustomField field={ field } error={ getValidationError( validity ) }>
             <TaggableSelect
                 isMulti={ field.multiple }
                 options={ options }
+                isLoading={ isLoading }
                 value={ decodeValue( data[ field.id ] ?? [] ) }
                 placeholder={ field.placeholder }
                 // Server already filters by `search`; keep all returned options.
@@ -217,9 +292,10 @@ const CreatableSelectField = ( {
                 isValidNewOption={ makeIsValidNewOption( field, data ) }
                 onInputChange={ ( input: string, meta: any ) => {
                     if ( meta.action === 'input-change' ) {
-                        search( input );
+                        onSearch( input );
                     }
                 } }
+                onMenuScrollToBottom={ onMenuScrollToBottom }
                 onChange={ ( value: any ) =>
                     onChange( { [ field.id ]: value ?? [] } )
                 }
@@ -228,31 +304,33 @@ const CreatableSelectField = ( {
     );
 };
 
-// Upsells/cross-sells etc.: plain async multi-select with debounced search.
+// Upsells/cross-sells etc.: multi-select with debounced server search. Options
+// are state-driven (rather than AsyncSelect's internal loadOptions cache) so
+// scrolling the menu can append further pages.
 const AsyncMultiSelectField = ( {
     data,
     field,
     onChange,
     validity,
 }: FieldProps ) => {
-    const loadOptions = useMemo(
-        () =>
-            debounced( ( input, callback ) =>
-                fetchOptions( field.api_endpoint, input ).then( callback )
-            ),
-        [ field.api_endpoint ]
-    );
-
-    useEffect( () => () => loadOptions.cancel(), [ loadOptions ] );
+    const { options, isLoading, onSearch, onMenuScrollToBottom } =
+        usePaginatedOptions( field.api_endpoint );
 
     return (
         <CustomField field={ field } error={ getValidationError( validity ) }>
-            <AsyncSelect
+            <Select
                 isMulti={ field.multiple }
-                cacheOptions
-                defaultOptions
+                options={ options }
+                isLoading={ isLoading }
                 value={ decodeValue( data[ field.id ] ) }
-                loadOptions={ loadOptions }
+                // Server already filters by `search`; keep all returned options.
+                filterOption={ null }
+                onInputChange={ ( input: string, meta: any ) => {
+                    if ( meta.action === 'input-change' ) {
+                        onSearch( input );
+                    }
+                } }
+                onMenuScrollToBottom={ onMenuScrollToBottom }
                 onChange={ ( value: any ) =>
                     onChange( { [ field.id ]: value } )
                 }
--- a/src/dashboard/product-editor/components/AsyncSelectEdit.tsx
+++ b/src/dashboard/product-editor/components/AsyncSelectEdit.tsx
@@ -9,6 +9,7 @@ import { addQueryArgs } from '@wordpress/url';
 import { debounce } from '@wordpress/compose';
 import { applyFilters } from '@wordpress/hooks';
 import { decodeEntities } from '@wordpress/html-entities';
+import { __ } from '@wordpress/i18n';
 import {
     useCallback,
     useEffect,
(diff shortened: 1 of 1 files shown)


akzmoudud:
**Before the fix**

- The Tags dropdown loaded only the first 20 tags in a single `per_page=20` request. There was no way to load more — scrolling did nothing, so any tag beyond the first 20 could not be browsed.
- When "vendors can create tags" was off, searching was also broken: the debounced loader handed react-select a stale, already-resolved promise, so the menu showed "No options" and the real server response was discarded — even though the server found the tag.

**After the fix**

- Opening the dropdown shows the first 20 options, and scrolling to the bottom keeps appending the next page until every option is reachable.
- Searching works in both configurations, and search results paginate on scroll too.

**Why this approach**

- **Scroll pagination instead of a bigger `per_page`:** WooCommerce caps `per_page` at 100, so raising the limit only moves the problem (a store with 150 tags breaks again). Loading pages on demand scales to any store size and keeps the initial request as light as it is today. It also matches what the legacy select2 form has always done, so both forms now behave consistently.
- **A controlled `Select` instead of `AsyncSelect`:** `AsyncSelect` owns its loaded options internally, so there is no way to append page 2 to what it is showing. Owning the options in our own state makes appending possible — and it removes the stale-promise bug as a side effect, since we no longer rely on `loadOptions`' return-value contract.
- **`hasMore` inferred from a full page instead of the `X-WP-TotalPages` header:** some proxies strip custom headers, which would silently disable pagination. Treating "we got a full page" as "there may be more" costs at most one empty request and cannot be broken by the environment.
- **No backend changes:** the REST endpoints already support `page`/`per_page`, so the whole fix stays in one component.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -36,29 +42,114 @@ const decodeLabel = ( option: any ) =>
...
+        () => debounced( ( input: string ) => load( input, 1 ) ),
+        [ load ]
+    );
+
+    const onMenuScrollToBottom = () => {
+        const { search, page, hasMore, loading } = stateRef.current;
+        if ( hasMore && ! loading ) {
+            load( search, page + 1 );
+        }
+    };
```
kzamanbd: **ERROR — pagination never fires on touch devices, and never on keyboard or scrollbar drag.**

`onMenuScrollToBottom` reaches react-select through `ScrollCaptor`, which is gated on `captureMenuScroll`. That prop defaults to `!isTouchCapable()` (`react-select/dist/Select-*.dev.js:1016`), and `isTouchCapable()` is just `document.createEvent('TouchEvent')` succeeding (`index-*.dev.js:251`). On Android, iOS, and any touch-screen laptop that returns `true`, so capture is disabled and this callback is **never invoked** — the reported bug stays unfixed for those vendors.

Even where capture is on, `useScrollCapture` binds only `wheel`, `touchstart` and `touchmove` (`Select-*.dev.js:668-671`). Dragging the scrollbar to the bottom does nothing, and neither does keyboard navigation — react-select's `focusOption` wraps back to index 0 at the end of the list, so a keyboard-only vendor can never reach page 2.

Neither wrapper overrides the default (`src/components/Select.tsx:65-89`, dokan-ui `TaggableSelect.tsx:65`), and test step 3 says "scroll the menu with the mouse wheel" — so this would pass review on desktop.

Suggest driving pagination from a `MenuList` `onScroll` instead, which fires for every scroll mechanism:

```tsx
const makeMenuList = ( onReachBottom: () => void ) => ( props: any ) => (
    <rsComponents.MenuList
        { ...props }
        innerProps={ {
            ...props.innerProps,
            onScroll: ( event: any ) => {
                const { scrollTop, scrollHeight, clientHeight } =
                    event.currentTarget;
                if ( scrollHeight - scrollTop - clientHeight < 24 ) {
                    onReachBottom();
                }
            },
        } }
    />
);
```

Then pass `components={ { MenuList: … } }` on both fields rather than `onMenuScrollToBottom`.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -36,29 +42,114 @@ const decodeLabel = ( option: any ) =>
...
             ? result.map( ( item: any ) => ( {
                   value: item.id,
                   label: decodeEntities( item.name ),
               } ) )
             : [];
+        return { options, hasMore: options.length === OPTIONS_PER_PAGE };
     } catch {
-        return [];
+        return { options: [], hasMore: false };
     }
```
kzamanbd: **ERROR — a single failed request permanently disables pagination for the rest of the session.**

The `catch` collapses a network failure into the same shape as a genuinely empty last page: `hasMore: false`. `load()` then writes that into `stateRef` (with `page` already advanced), and since `onMenuScrollToBottom` is guarded by `hasMore` and nothing retries, the field is stuck.

Concrete failure: a vendor scrolls to page 3, that request 5xx's or the connection blips. The list freezes at 60 tags — no error, no spinner, and further scrolling does nothing for the remainder of the session. The only escape is typing a search term to force a page-1 reset. If page 1 is the request that fails, the field reads "No options" permanently.

This is new: before this PR there was no `hasMore` latch to poison. Suggest distinguishing failure from an empty page:

```ts
    } catch {
        return { options: [], hasMore: false, failed: true };
    }
```

and bailing out of the state write on `failed` so `search`/`page`/`hasMore` survive and the next scroll retries.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -36,29 +42,114 @@ const decodeLabel = ( option: any ) =>
...
+                    return; // superseded by a newer search/page request
+                }
+                stateRef.current = {
+                    ...stateRef.current,
+                    search,
+                    page,
+                    hasMore: result.hasMore,
+                    loading: false,
+                };
+                setIsLoading( false );
```
kzamanbd, line 164: Follow-on from the `catch` above — this is where the poisoned `hasMore` lands. With a `failed` flag on the result, the guard would be:

```ts
fetchOptions( endpoint, search, page ).then( ( result ) => {
    if ( requestId !== stateRef.current.requestId ) {
        return; // superseded by a newer search/page request
    }
    if ( result.failed ) {
        // Keep search/page/hasMore so the next scroll retries this page.
        stateRef.current.loading = false;
        setIsLoading( false );
        return;
    }
    …
} );
```

Worth surfacing the failure to the vendor too — right now a 5xx and "this store has no tags" are indistinguishable in the UI.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -185,41 +276,26 @@ const CreatableSelectField = ( {
...
                 placeholder={ field.placeholder }
                 // Server already filters by `search`; keep all returned options.
                 filterOption={ null }
                 isValidNewOption={ makeIsValidNewOption( field, data ) }
                 onInputChange={ ( input: string, meta: any ) => {
                     if ( meta.action === 'input-change' ) {
-                        search( input );
+                        onSearch( input );
                     }
                 } }
```
kzamanbd: **ERROR — the search term is never cleared, so the list stays filtered after the menu closes.**

This only handles `meta.action === 'input-change'`, but react-select resets its own `inputValue` to `''` on three other actions, all of which land here:

- `setValue` → `onInputChange( '', { action: 'set-value' } )` (`Select-*.dev.js:1270`)
- `onMenuClose` → `action: 'menu-close'` (`:1835`)
- `onInputBlur` → `action: 'input-blur'` (`:1583`)

Because none of those are handled, `stateRef.current.search` keeps the last typed term and `options` keeps that term's results. `src/components/Select.tsx:80-81` defaults both `blurInputOnSelect` and `closeMenuOnSelect` to `true`, so this fires on **every** selection, not just an edge case.

Repro: type `shirt` → pick `shirt-blue` → menu closes → reopen Tags. The input is empty, but the menu shows only shirt matches, and scrolling paginates `shirt`. The vendor can't get back to the full list without typing something and clearing it — which is close to the original complaint in getdokan/dokan-pro#6026.

Suggest resetting in the same handler (needs `load` returned from the hook):

```tsx
onInputChange={ ( input: string, meta: any ) => {
    if ( meta.action === 'input-change' ) {
        onSearch( input );
    } else if ( meta.prevInputValue ) {
        // react-select clears its input on blur/menu-close/set-value;
        // drop the stale search so the next open shows page 1 again.
        onSearch.cancel();
        load( '', 1 );
    }
} }
```

Resetting on `onMenuOpen` instead would work equally well. Same fix needed on `AsyncMultiSelectField`.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -228,31 +304,33 @@ const CreatableSelectField = ( {
...
+                isLoading={ isLoading }
                 value={ decodeValue( data[ field.id ] ) }
-                loadOptions={ loadOptions }
+                // Server already filters by `search`; keep all returned options.
+                filterOption={ null }
+                onInputChange={ ( input: string, meta: any ) => {
+                    if ( meta.action === 'input-change' ) {
+                        onSearch( input );
+                    }
+                } }
```
kzamanbd: Same stale-search problem as `CreatableSelectField` — see the comment there. Both variants need the reset.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -228,31 +304,33 @@ const CreatableSelectField = ( {
...
+            <Select
                 isMulti={ field.multiple }
-                cacheOptions
-                defaultOptions
+                options={ options }
+                isLoading={ isLoading }
                 value={ decodeValue( data[ field.id ] ) }
-                loadOptions={ loadOptions }
+                // Server already filters by `search`; keep all returned options.
+                filterOption={ null }
```
kzamanbd, line 466: **WARNING — `filterOption={ null }` is new for this field and makes typing look ignored for ~300 ms.**

With client filtering off, the menu keeps rendering the previously loaded page verbatim while the 300 ms debounce plus the round trip elapse — so the first few keystrokes appear to do nothing. `AsyncSelect` filtered loaded options locally, so this is a small regression on the linked-products fields specifically.

Either keep the default client filter as a pre-filter over the loaded page (the server result is a superset of what matches locally anyway), or blank the list on `input-change` so `isLoading` carries the feedback.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -36,29 +42,114 @@ const decodeLabel = ( option: any ) =>
...
+        const { search, page, hasMore, loading } = stateRef.current;
+        if ( hasMore && ! loading ) {
+            load( search, page + 1 );
+        }
+    };
+
+    useEffect( () => {
+        load( '', 1 );
+        return () => onSearch.cancel();
+    }, [ load, onSearch ] );
```
kzamanbd, line 277: **WARNING — description/behaviour mismatch.** The PR body says "Opening the menu loads page 1", but this loads page 1 on *mount*: every `async_select` field fires a request when the product editor renders, whether or not the vendor ever opens it (four requests with Pro active).

Behaviour is unchanged from before this PR, so not a regression — just fix the wording. That said, moving the initial load to `onMenuOpen` would be the real win here, and it would compose nicely with the stale-search reset.

Minor, in the same effect: `options` isn't cleared when `endpoint` changes, so the old endpoint's options stay visible until the new page 1 resolves.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -36,29 +42,114 @@ const decodeLabel = ( option: any ) =>
...
     Array.isArray( value ) ? value.map( decodeLabel ) : decodeLabel( value );
 
-// Fetch `{ value, label }` options from an endpoint, optionally filtered by `search`.
+const OPTIONS_PER_PAGE = 20;
+
+// Fetch one page of `{ value, label }` options from an endpoint, optionally
+// filtered by `search`. `hasMore` is inferred from a full page rather than the
+// X-WP-TotalPages header so a proxy stripping custom headers can't disable
+// pagination; the only cost is one empty request when the total is an exact
+// multiple of the page size.
```
kzamanbd: **SUGGESTION — the stated rationale isn't the real constraint.** `apiFetch` without `parse: false` never exposes response headers at all, so `X-WP-TotalPages` isn't available regardless of what any proxy does. The full-page inference is the right choice; the comment should just say that reading the header would mean opting into `parse: false` and unwrapping every response by hand.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -36,29 +42,114 @@ const decodeLabel = ( option: any ) =>
...
+                            ? result.options
+                            : [ ...previous, ...result.options ];
+                    // Terms can shift between pages while browsing (e.g. a tag
+                    // created mid-session); keep the first occurrence of each.
+                    const seen = new Set();
+                    return merged.filter(
+                        ( option ) =>
+                            ! seen.has( option.value ) &&
+                            seen.add( option.value )
+                    );
```
kzamanbd: **SUGGESTION** — the dedupe leans on `seen.add()` returning a truthy `Set` to double as the `filter` predicate. It works, but it reads as a puzzle. A `Map` keyed by `value` says the same thing plainly:

```ts
return [ ...new Map( merged.map( ( o ) => [ o.value, o ] ) ).values() ];
```

(`Map` keeps first-insertion order, and the later duplicate overwrites the value at the same position — if keeping the *first* occurrence matters for a label that changed mid-session, a short `for` loop is clearer still.)

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
```diff
@@ -36,29 +42,114 @@ const decodeLabel = ( option: any ) =>
...
+                } );
+            } );
+        },
+        [ endpoint ]
+    );
+
+    const onSearch = useMemo(
+        () => debounced( ( input: string ) => load( input, 1 ) ),
+        [ load ]
+    );
```
kzamanbd, line 178: **SUGGESTION** — `debounced` is declared at line 213, below this use. Safe at runtime since the hook only runs during render, but it reads backwards; move `debounced` above `usePaginatedOptions`.

Also: `onMenuScrollToBottom` just below is the one member of this hook that isn't memoized, so it gets a fresh identity on every render while `load` and `onSearch` are stable. Worth a `useCallback` for consistency.

kzamanbd (changes requested):
Thanks for the thorough write-up — the diagnosis is right on both counts, and the stale-`loadOptions`-promise analysis in particular is a good catch. Moving to a state-driven `Select` is the correct call.

Verified locally: ESLint clean, `tsc --noEmit` clean, no PHP touched. Both endpoints (`/wc/v3/products/tags`, `/dokan/v2/products?product_type=…`) do accept `page`/`per_page`, and `addQueryArgs` merges correctly into the Pro endpoints' existing query string — so "no backend changes" holds.

Three defects in the new hook that are reachable in normal vendor use, though. Details inline; summary:

1. **Pagination never fires on touch devices.** `onMenuScrollToBottom` is gated on react-select's `captureMenuScroll`, which defaults to `!isTouchCapable()`. On Android/iOS and touch-screen laptops the callback is never wired, so the bug this PR closes is still live for those vendors. The test steps only cover the mouse wheel, so QA would pass and it would ship broken.
2. **One failed request permanently disables pagination for the session.** The `catch` collapses failure into `hasMore: false`, and nothing ever retries.
3. **The search term is never reset**, so the option list stays filtered after the menu closes — which happens on every single selection.

Plus a couple of smaller UX/readability notes. Happy to talk through any of it.

dev-shahed:
@akzmoudud  Dropdown keeps the previous search's results after the menu is closed, so the full list is no longer browsable

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
akzmoudud: Done in e9556eb0. `fetchOptions` now returns a `failed` flag distinct from an empty last page, and `load()` bails out of the state write on it — `search`/`page`/`hasMore` survive, so the next scroll or menu open retries. Also surfaced to the vendor as you suggested: `noOptionsMessage` reports "Could not load options. Please try again." rather than looking like an empty store.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
akzmoudud: Done in e9556eb0, via the second option you offered: `onInputChange` blanks the list and sets `isLoading` on `input-change`, so the loading state carries the feedback through the debounce + round trip instead of the previous page sitting there looking ignored. Kept `filterOption={ null }` so appended pages are never client-filtered out.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
akzmoudud: Done in e9556eb0 — took the "real win" rather than just fixing the wording: page 1 is now fetched on first `onMenuOpen`, not on mount. Verified 0 tag requests before the menu is opened (previously 4 with Pro active). `options` are also cleared when `endpoint` changes. The PR description has been updated to match.

src/dashboard/product-editor/components/AsyncSelectEdit.tsx
akzmoudud: Done in e9556eb0 — `debounced` now sits above `usePaginatedOptions`. The memoization point resolved itself more strongly: `onMenuScrollToBottom` is gone entirely (replaced by the scroll-listener approach for the touch/keyboard reasons in the other thread), and every member the hook returns is memoized — `onInputChange`/`onMenuOpen`/`onMenuClose`/`noOptionsMessage` via `useCallback`, plus internal `load`/`loadMore` and `onSearch`.

kzamanbd (approved):

PR #3380, opened 2026-08-14
fix(cache): remove the redundant group flush from invalidate_group()

### LINKED ISSUE #3369: invalidate_group() flushes a cache group that is never written to, causing a full Redis keyspace scan on every call
### Bug Description

# Upstream bug report for Dokan

Ready to paste into a new issue at <https://github.com/getdokan/dokan/issues>.
Fill in the two bracketed values before sending.

Keep the tone as it is. The report asks for a two line deletion and shows the
evidence for why it is safe, which is the fastest path to it being accepted.

---

## Title

```
invalidate_group() flushes a cache group that is never written to, causing a full Redis keyspace scan on every call
```

### Summary

`Traits\ObjectCache::invalidate_group()` calls `wp_cache_flush_group()` on a
group name that Dokan never writes to. With a Redis or Memcached object cache
the call is not free: it triggers a full keyspace scan that deletes nothing.

On our production store one WooCommerce checkout made 114 of these calls and
spent **28.5 seconds** with Redis blocked. Removing the call changes no
behavior, because the `wp_cache_set()` on the following line is the entire
invalidation mechanism.

### The code

`includes/Traits/ObjectCache.php`:

```php
public static function invalidate_group( $group ) {
    $group     = static::get_cache_group_with_prefix( $group );
    $supported = wp_cache_supports( 'flush_group' );

    if ( $supported ) {
        wp_cache_flush_group( $group . '_prefix' );   // flushes GROUP "{$group}_prefix"
    }

    return wp_cache_set( $group . '_prefix', static::get_time_prefix(), $group );
    //                   ^^^^^^^^^^^^^^^^^^                             ^^^^^^
    //                   KEY "{$group}_prefix"              in GROUP "{$group}"
}
```

The flush targets a **group** called `{$group}_prefix`. The `wp_cache_set()`
writes a **key** called `{$group}_prefix` into the group `{$group}`, which is
its third argument. These are two different groups, and nothing in Dokan ever
writes to the first one.

### Evidence from a production Redis instance

The key exists in group `dokan_cache_withdraws_seller_55054`:

```
$ redis-cli --scan --pattern '*dokan_cache_withdraws_seller_55054*'
<prefix>wp:dokan_cache_withdraws_seller_55054:dokan_cache_withdraws_seller_55054_prefix
        └─ group ─────────────────────────── └─ key ──────────────────────────────────
```

The group being flushed, `dokan_cache_withdraws_seller_55054_prefix`, holds
nothing:

```
$ redis-cli --scan --pattern '*:dokan_cache_withdraws_seller_55054_prefix:*'
(empty)
```

### Why it is expensive

Redis has no native concept of a cache group, so Redis Object Cache implements
`flush_group()` as a Lua script that walks the keyspace with `SCAN ... MATCH`.
`MATCH` filters server side but is not an index, so the entire keyspace is
examined on every call regardless of how many keys match.

Measured on our instance, with 58,259 keys:

| | |
| --- | --- |
| Cost per `invalidate_group()` | ~250 ms |
| Keys examined per call | 58,259 |
| Keys deleted per call | **0** |

Redis is single threaded, so each of those 250 ms windows blocks every other
client on the site. In the same period a `PING`, which does no work at all,
measured 207 ms at p95 because it was queued behind these scans.

### Impact in a real checkout

`WC_Order::save()` runs many times during a WooCommerce checkout, and each one
fires `woocommerce_update_order`, which Dokan hooks to invalidate several cache
groups. From a New Relic trace of a single checkout on our store:

| Function | Calls in one request |
| --- | --- |
| `WC_Order::save` | 24 |
| `DokanCache::invalidate_group` | **114** |
| `wp_cache_flush_group` | 114 |
| `WP_Object_Cache::execute_lua_script` | 114 |
| Redis `EVAL` | 114 |

114 calls at 250 ms is roughly **28.5 seconds of blocked Redis inside one
checkout**, and it degraded every other request on the site for the duration.
That checkout took 62.67 seconds end to end.

The cost scales linearly with keyspace size, so larger stores are affected more.
At 250,000 keys each call would cost about 1 second.

### Suggested fix

```diff
 public static function invalidate_group( $group ) {
-    $group     = static::get_cache_group_with_prefix( $group );
-    $supported = wp_cache_supports( 'flush_group' );
-
-    if ( $supported ) {
-        wp_cache_flush_group( $group . '_prefix' );
-    }
+    $group = static::get_cache_group_with_prefix( $group );

     return wp_cache_set( $group . '_prefix', static::get_time_prefix(), $group );
 }
```

### Why this is safe

The `wp_cache_set()` is the complete invalidation. It overwrites the group's
version prefix with a fresh timestamp, so every cache key derived from the old
prefix becomes unreachable immediately. That mechanism does not depend on the
flush in any way.

The flush cannot be contributing anything, because its target group is empty.

If the intent was to flush the real group, `wp_cache_flush_group( $group )`
would delete the very prefix key that the next line rewrites. That is still
redundant, and it would still cost a keyspace scan on Redis. The
`wp_cache_set()` on its own is both sufficient and O(1).

### A second, unrelated finding

In the same trace, `Dokan\Fees::get_shipping_fee_recipient()` called
`WC_Order::save()` and took **7.31 seconds**. A getter writing the whole order
row is surprising on its own, and each save fires `woocommerce_update_order`,
which feeds back into the invalidation described above. Worth a look as a
separate issue.

### Environment

| | |
| --- | --- |
| Dokan | Version 5.0.12  and Dokan Pro | Version 5.0.10  |
| WooCommerce | 11.0.0 |
| WordPress | 7.0.3 |
| PHP | 8.3.21 |
| Object cache | Redis Object Cache 2.8.0 (PhpRedis 6.2.0) |
| Redis | 7.2.5 |
| Keyspace | 58,259 keys |

### How we found it

New Relic APM. The checkout was invisible at first because WooCommerce routes
`?wc-ajax=checkout` through `template_redirect` rather than admin-ajax or REST,
so it reported under a generic transaction name. Once named separately, faceting
its spans by function name produced the identical 114 counts above and made the
chain obvious.

Happy to provide the full trace or run any additional diagnostics.

### Step To Reproduce

_No response_

### Expected Behaviour

See description.

### Actual Behaviour

See description.

### Additional Information

_No response_

```shell
| Dokan | Version 5.0.12  and Dokan Pro | Version 5.0.10  |
| WooCommerce | 11.0.0 |
| WordPress | 7.0.3 |
| PHP | 8.3.21 |
| Object cache | Redis Object Cache 2.8.0 (PhpRedis 6.2.0) |
| Redis | 7.2.5 |
| Keyspace | 58,259 keys |
```

mrabbani:
### All Submissions:

* [x] My code is tested

`Traits\ObjectCache::invalidate_group()` called `wp_cache_flush_group( $group . '_prefix' )` — but that group name is never written to. The invalidation prefix key is a **key** named `{$group}_prefix` stored **inside** the group `{$group}` (the third argument of the `wp_cache_set()` on the next line). The flush therefore never deleted a single key.

It was not free, though: persistent object caches implement `flush_group()` as a keyspace scan. Redis Object Cache runs a Lua script doing `SCAN ... MATCH` over **every** key in the database, and Redis is single-threaded, so each call blocks all other clients. On the reporting store (~58k keys) each call cost ~250 ms and one WooCommerce checkout triggered it 114 times — ~28.5 s of blocked Redis per checkout (#3369).

This PR removes the flush. The `wp_cache_set()` time-prefix rewrite is the complete invalidation mechanism: every key derived from the old prefix becomes unreachable, and orphaned entries are evicted by their own expiry (default 2 weeks). A comment now documents why no flush belongs there.

Also adds `tests/php/src/Cache/ObjectCacheTest.php` pinning the invalidation contract:
- invalidated group values become unreachable
- the group works again after invalidation
- other groups are untouched
- `invalidate_group()` never calls `flush_group()` (verified via a spy `WP_Object_Cache`; this test fails against the previous code)

### Closes

* Closes #3369

1. Enable Redis Object Cache on a test site (`wp plugin install redis-cache --activate && wp redis enable`).
2. Seed a large keyspace: `redis-cli eval "for i=1,60000 do redis.call('set','filler:'..i,'1') end return 'ok'" 0`
3. Time an invalidation: `wp eval '$t=microtime(true); \WeDevs\Dokan\Cache::invalidate_group("withdraws"); echo round((microtime(true)-$t)*1000,2)." ms\n";'`
4. Before this PR: ~30 ms locally (scales linearly with keyspace size). After: ~0.2 ms, constant.
5. Verify invalidation still works: `wp eval '\WeDevs\Dokan\Cache::set("k","v","withdraws"); \WeDevs\Dokan\Cache::invalidate_group("withdraws"); var_dump(\WeDevs\Dokan\Cache::get("k","withdraws"));'` → `bool(false)`.

### Changelog entry

*fix: Group cache invalidation no longer triggers a full Redis keyspace scan*

Previously every `Cache::invalidate_group()` call ran `wp_cache_flush_group()` against a group that never holds any key, forcing persistent object caches (Redis/Memcached) to scan the entire keyspace and delete nothing — up to ~28 s of blocked Redis in a single checkout on large stores. Invalidation now relies solely on the O(1) time-prefix rewrite, which has always been the actual invalidation mechanism. No behavior change; measured ~115× faster per call at a 60k keyspace.

### Before Changes

`invalidate_group()` at 60,121 Redis keys: **~30 ms per call**, full keyspace SCAN, 0 keys deleted. 114 calls per checkout on the reporting store → ~28.5 s of blocked Redis.

### After Changes

`invalidate_group()` at 60,121 Redis keys: **~0.2 ms per call**, no scan. Invalidation contract unchanged (covered by new unit tests).

https://claude.ai/code/session_016Ht3K5Bxqoc68UtdUAuJKU

### DIFF
--- a/includes/Traits/ObjectCache.php
+++ b/includes/Traits/ObjectCache.php
@@ -155,6 +155,9 @@ trait ObjectCache {
      * ```
      *
      * @since 3.3.2
+     * @since DOKAN_SINCE Removed the redundant `wp_cache_flush_group()` call. It targeted a group
+     *                    no key is ever written to, while forcing persistent backends such as
+     *                    Redis into a full keyspace scan on every call.
      *
      * @param string $group Group of caches to clear.
      *
@@ -163,11 +166,10 @@ trait ObjectCache {
     public static function invalidate_group( $group ) {
         $group = static::get_cache_group_with_prefix( $group );
 
-        $supported = wp_cache_supports( 'flush_group' );
-        if ( $supported ) {
-            wp_cache_flush_group( $group . '_prefix' );
-        }
-
+        // Rewriting the group's time prefix is the complete invalidation: every key
+        // derived from the old prefix becomes unreachable, and the orphaned entries
+        // are evicted by their own expiry. Do not add a wp_cache_flush_group() call
+        // here — on Redis/Memcached a group flush scans the entire keyspace.
         return wp_cache_set( $group . '_prefix', static::get_time_prefix(), $group );
     }
 
--- a/tests/php/src/Cache/ObjectCacheTest.php
+++ b/tests/php/src/Cache/ObjectCacheTest.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace WeDevs\Dokan\Test\Cache;
+
+use WeDevs\Dokan\Cache;
+use WeDevs\Dokan\Test\DokanTestCase;
+use WP_Object_Cache;
+
+/**
+ * Tests for the ObjectCache trait's group invalidation contract.
+ *
+ * @group cache
+ */
+class ObjectCacheTest extends DokanTestCase {
+
+    /**
+     * Invalidation must make every value cached under the group unreachable.
+     */
+    public function test_invalidated_group_values_become_unreachable() {
+        Cache::set( 'balance', 100, 'withdraws' );
+        $this->assertSame( 100, Cache::get( 'balance', 'withdraws' ) );
+
+        $this->assertTrue( Cache::invalidate_group( 'withdraws' ) );
+
+        $this->assertFalse( Cache::get( 'balance', 'withdraws' ) );
+    }
+
+    /**
+     * The group must accept and serve fresh values after an invalidation.
+     */
+    public function test_cache_works_again_after_invalidation() {
+        Cache::set( 'balance', 100, 'withdraws' );
+        Cache::invalidate_group( 'withdraws' );
+
+        Cache::set( 'balance', 200, 'withdraws' );
+
+        $this->assertSame( 200, Cache::get( 'balance', 'withdraws' ) );
+    }
+
+    /**
+     * Invalidating one group must not touch values cached in other groups.
+     */
+    public function test_invalidating_one_group_leaves_other_groups_intact() {
+        Cache::set( 'item_a', 1, 'withdraws' );
+        Cache::set( 'item_b', 2, 'vendors' );
+
+        Cache::invalidate_group( 'withdraws' );
+
+        $this->assertFalse( Cache::get( 'item_a', 'withdraws' ) );
+        $this->assertSame( 2, Cache::get( 'item_b', 'vendors' ) );
+    }
+
+    /**
+     * Invalidation must never flush a cache group: on persistent backends such as
+     * Redis, wp_cache_flush_group() costs a full keyspace scan per call, and the
+     * time-prefix rewrite already makes the group's old keys unreachable.
+     *
+     * @see https://github.com/getdokan/dokan/issues/3369
+     */
+    public function test_invalidate_group_never_flushes_a_cache_group() {
+        global $wp_object_cache;
+
+        $original = $wp_object_cache;
+
+        $spy = new class() extends WP_Object_Cache {
+            /**
+             * Groups passed to flush_group() during the test.
+             *
+             * @var array
+             */
+            public $flushed_groups = [];
+
+            public function flush_group( $group ) {
+                $this->flushed_groups[] = $group;
+
+                return parent::flush_group( $group );
+            }
+        };
+
+        $wp_object_cache = $spy; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+
+        try {
+            Cache::set( 'balance', 100, 'withdraws' );
+            Cache::invalidate_group( 'withdraws' );
+        } finally {
+            $wp_object_cache = $original; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+        }
+
+        $this->assertSame( [], $spy->flushed_groups );
+    }
+}


MdAsifHossainNadim:
## Verdict: ✅ Approve

### What changes

`Cache::invalidate_group()` drops the `wp_cache_flush_group( $group . '_prefix' )` call. `{$group}_prefix` is a **key stored inside** group `{$group}`, never a group of its own — so the flush deleted nothing while forcing Redis/Memcached into a full keyspace `SCAN` on every call. The `wp_cache_set()` time-prefix rewrite that remains has always been the complete invalidation: old keys become unreachable instantly and expire on their own. Plus `tests/php/src/Cache/ObjectCacheTest.php` pinning the contract, including a spy test that fails against the old code.

### Testing scope

A/B verified against `develop` on a live site — Redis Object Cache 2.6.5, WP 7.1, Dokan 5.0.15 + Pro:

| Check | `develop` | This PR |
| --- | --- | --- |
| `invalidate_group()` @ 617-key keyspace | 9.04 ms, **0 keys deleted** | **0.07 ms**, no scan |
| `flush_group` calls per single order-save cycle | 32, every one deleting 0 keys | **0** |

Behaviour proven unchanged:

- set → get → invalidate → miss → re-set → hit: identical results on both trees
- invalidating one group leaves sibling groups intact
- cross-request invalidation through real Redis across 4 separate PHP processes
- vendor dashboard order status change (Processing → On-hold): row, status tab counts and sidebar badge all refresh correctly; Redis `EVAL` counter unchanged (33 → 33) across that request
- swept Lite + Pro + sibling plugins: no other `wp_cache_flush_group()` call site exists, so no Pro companion is needed

### Non-blocking suggestions

1. Set `protected $is_unit_test = true;` on `ObjectCacheTest` — these four tests need neither the `WP_REST_Server` nor the seeded users that `DokanTestCase::set_up()` boots.
2. Add a cleanup line to the test instructions; step 2 leaves 60k TTL-less keys behind → `redis-cli --scan --pattern 'filler:*' | xargs redis-cli del`.
3. For whoever merges: `phpcs.yml` and `e2e_api_tests.yml` are daily-scheduled only (no PR trigger) and Lite has no PHPUnit workflow, so please run `npm run test:phpunit` locally first. The pre-existing `OrderStatusChangeTest` failures on `develop` are unrelated.

Worth its own follow-up issue (out of scope here): `Fees::get_shipping_fee_recipient()` (`includes/Fees.php:171`) calls `$order->save()` from a getter — 7.31 s in the reporter's trace, and each save re-enters this same invalidation path. `get_tax_fee_recipient()` shares the pattern.

MdAsifHossainNadim (approved):

Issue #232, opened 2018-04-23
dokan social login issue error 500


mohamadhegazy:
1- valid error 500 issue appeared when tried to login with social media 

it appears also if the user cancel the login via social media 

it happens fresh wordpress 4.9.5 and dokan 2.7.8 and pro 2.7.5 and reported to support team and they validate it

2- also if disabled dokan social login it still conflict with woo-commerce social login 2.4.1

mohamadhegazy:
2- this is happening because you declared Hybrid_Auth class in the main dokan-pro.php and in libs also

i changed all Hybrid_Auth to Hybrid_Auth2 in all pro plugin files and the issue solved so i can use another social login plugin, but issue number 1 is still exist.

i think if you gonna implement social login for pro users by default, you have to use another class name as all social login plugins uses the same class, and we as pro or business users  we have the right to use your plugin and still uses our own if some of us wishes, specially when there's an issue with yours, as wedevs dosn't issue updates that fast, and we cant edit the same files over and over each time there's an update
thank you

itsachalsingh:
Hello wedevs

when you resolve this. I am also getting the same error

Thank you

shabnam611:
Hello,

Dokan social login is a pro version feature. So, please create support ticket and report your issue with screenshot, so that our support team can take a look at your issues.

On March 2018, FaceBook has brought some changes, which made us to fix those issues with Facebook for dokan social login. The latest version of Dokan pro is 2.8.3 and the latest version of Dokan lite is 2.8.5. 

Please make sure you are using the latest version of Dokan lite and Dokan pro. 

Thanks!

PR #3389, opened 2026-08-24
fix(components): read the plugin-ui external lazily in the DataViews wrapper

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

#3382 wrapped plugin-ui's `DataViews` in `DokanDataViews.tsx` and, in doing so, evaluated `@wedevs/plugin-ui` **at module scope** twice: a top-level `const ForwardedDataViews = PluginUIDataViews as …` alias, and an eager `Object.assign( DokanDataViews, { Pagination: PluginUIDataViews.Pagination, … } )`. `@wedevs/plugin-ui` is a webpack external that compiles to the `window.dokan.pluginUI` global (`webpack-dependency-mapping.js`, `SHARED_UI_EXTERNALS`), so those two statements compile to *reads of that global while the bundle is still initialising*. On any page where the global is not there at that moment, the module throws `TypeError: Cannot read properties of undefined (reading 'DataViews')` uncaught, and the **entire bundle containing it dies** — three of them: `product-editor-utils.js`, `frontend.js` and `dokan-admin-dashboard.js`, exactly the three in getdokan/dokan-pro#6099.

**Why the global is missing even though `plugin-ui.js` is enqueued first** — reproduced locally on builds whose `?ver=` hashes are byte-identical to the issue report. The dependency graph does order `plugin-ui.js` correctly; what kills it is print order *within* the page:

1. `plugin-ui.js` runs and sets `window.dokan.pluginUI` ✓ — then `components.js` initialises fine (which is why it is *not* among the throwing bundles)
2. `wp_localize_script( 'dokan-vue-vendor', 'dokan', … )` (`includes/Assets.php:103`) prints `var dokan = {…}`, which **replaces `window.dokan` wholesale**, wiping `pluginUI`, `dokanUI`, `components`, `utilities`
3. `product-editor-utils.js`, `frontend.js` and `dokan-admin-dashboard.js` initialise after that block, and #3382's eager reads hit `undefined.DataViews` → three uncaught TypeErrors

The clobber itself is long-standing, systemic (three active `wp_localize_script( …, 'dokan', … )` sites) and harmless to every other module, because the shared-bundle contract has always been *everything lazy*: webpack's harmony re-exports and in-render reads never touch the globals during module init. The wrapper was the sole violation of that contract in both Lite and Pro (verified with an AST scan of every importer of `@wedevs/plugin-ui`, `@getdokan/dokan-ui` and the `@dokan/*` externals family — zero other module-scope reads). So the right-altitude fix is to restore the contract in the wrapper, not to touch the localize (renaming/merging the decade-old public `dokan` object is a separate hardening with its own blast radius; and it would not subsume this fix anyway, since third-party enqueue order can still produce pages where the global is legitimately absent).

**The fix:**

* The type-cast alias moves inside the component, so the external is first read at render time — when plugin-ui is necessarily loaded.
* The five sub-components (`Pagination`, `Layout`, `Search`, `Filters`, `BulkActionToolbar`) are forwarded through **lazy `Object.defineProperties` getters** instead of eager `Object.assign` values. They stay enumerable, keep returning the same (stable) component references, and remain reachable for runtime consumers via `window.dokan.components.DataViews.*`. `Object.assign` with getter objects or spreads is *not* equivalent — both evaluate getters at module scope, which is the bug.
* The export is typed `as typeof PluginUIDataViews`, i.e. exactly the type the barrel exposed before #3382, reusing plugin-ui's own function-plus-namespace declaration for the statics.
* The in-file comment records why the reads must stay lazy, with the issue reference, so the "obvious cleanup" back to `Object.assign` is not made blindly.

This PR is deliberately a one-file change. The wider invariant — *never evaluate a binding imported from `@wedevs/plugin-ui`, `@getdokan/dokan-ui` or any `@dokan/*` external at module scope* — is worth writing down in the shared-bundle headers and the frontend docs, but that is documentation churn rather than part of this regression fix and can follow separately.

Everything #3382 shipped is preserved: the `dokan-lite`-domain labels (`No data found`, `Add Filter` / `Remove filter`, the destructive-confirm message), caller overrides, and Pro's no-rebuild consumption through `window.dokan.components`.

### Related Pull Request(s)

* #3382 — the PR that introduced the eager reads

### Closes

* Closes getdokan/dokan-pro#6099

1. Build from this branch: `npm ci && npm run build` (a cold build matters — a warm `node_modules/.cache/webpack` from an older tree can silently produce a non-externalised bundle that hides the bug).
2. Log in to wp-admin, open **Dokan → Settings**, and open the browser console.
3. **Before (develop):** three uncaught `TypeError: Cannot read properties of undefined (reading 'DataViews')` — one each from `product-editor-utils.js`, `frontend.js`, `dokan-admin-dashboard.js`. **After (this branch):** none.
4. Confirm the Dokan admin Dashboard (`admin.php?page=dokan`) still renders its React UI.
5. Confirm no regression of #3382 on a vendor-dashboard list, e.g. `/dashboard/new/#/withdraw-requests` with an empty tab: the list renders, the empty state still shows the (translatable) `No data found`, and the sub-components still resolve — quick console probe on that page:

```js
typeof dokan.components.DataViews             // "function"
typeof dokan.components.DataViews.Pagination  // "function" (getter, resolved lazily)
Object.keys( dokan.components.DataViews )     // ["Pagination","Layout","Search","Filters","BulkActionToolbar"]
```

### Changelog entry

*Fix – Uncaught JavaScript errors while Dokan's bundles load*

A recent change made the vendor-dashboard DataViews wrapper read the shared plugin-ui bundle's global while JavaScript bundles were still initialising. On pages where that global was not available at that instant, three bundles (`product-editor-utils.js`, `frontend.js`, `dokan-admin-dashboard.js`) each crashed on load with an uncaught `TypeError: Cannot read properties of undefined (reading 'DataViews')` — in wp-admin and on the storefront — silently breaking any feature living in those bundles on the affected pages. The wrapper now defers every read until the component or its sub-components are actually used, so the bundles initialise cleanly everywhere, while the translatable list labels the wrapper exists for keep working exactly as before.

### Before Changes

On `develop` (byte-identical `?ver=` hashes to the issue: `541eb079…`, `45ebaadb…`, `bf4b63cb…`), wp-admin **Dokan → Settings**, `window.dokan.pluginUI` undefined at bundle-init time — executing each bundle against the live page state:

| Bundle | Result |
|---|---|
| `product-editor-utils.js` | `TypeError: Cannot read properties of undefined (reading 'DataViews')` |
| `frontend.js` | `TypeError: Cannot read properties of undefined (reading 'DataViews')` |
| `dokan-admin-dashboard.js` | `TypeError: Cannot read properties of undefined (reading 'DataViews')` |
| `components.js` | no throw (initialises before the `var dokan = {…}` overwrite) |

### After Changes

Same page, same missing global, fixed build — verified live:

| Check | Result |
|---|---|
| `product-editor-utils.js` / `frontend.js` / `dokan-admin-dashboard.js` init with `window.dokan.pluginUI` undefined | no throw ✅ |
| Compiled output audit | every `.DataViews` access sits inside the component body or a getter closure; module scope only holds the harmless external require ✅ |
| Dokan admin Dashboard (`admin.php?page=dokan`) | React UI renders ✅ |
| Vendor dashboard `/dashboard/new/#/withdraw-requests`, Pending (0) | list renders, empty state shows `No data found` from the `dokan-lite` domain ✅ |
| `DataViews.Pagination/Layout/Search/Filters/BulkActionToolbar` | all resolve through the getters and stay enumerable ✅ |
| `npx tsc` | project baseline unchanged (no errors in the touched files); ESLint clean ✅ |

### DIFF
--- a/src/components/dataviews/DokanDataViews.tsx
+++ b/src/components/dataviews/DokanDataViews.tsx
@@ -2,14 +2,14 @@ import { DataViews as PluginUIDataViews } from '@wedevs/plugin-ui';
 import type { DataViewsProps } from '@wedevs/plugin-ui';
 import { __ } from '@wordpress/i18n';
 
-// plugin-ui's generic defers a conditional on Item that TS cannot forward; runtime props are unchanged.
-const ForwardedDataViews = PluginUIDataViews as (
-    props: any // eslint-disable-line @typescript-eslint/no-explicit-any
-) => JSX.Element;
-
 // plugin-ui labels its chrome with core's `default` domain, which Dokan's language files can never reach.
 function DokanDataViews< Item >( props: DataViewsProps< Item > ) {
     const { actions, filter, emptyTitle } = props;
+
+    // plugin-ui's generic defers a conditional on Item that TS cannot forward; runtime props are unchanged.
+    const ForwardedDataViews = PluginUIDataViews as (
+        p: any // eslint-disable-line @typescript-eslint/no-explicit-any
+    ) => JSX.Element;
 
     // Only a destructive action reaches the confirm dialog, so leave every other action object untouched.
     const localizedActions = actions?.map( ( action ) =>
@@ -45,11 +45,20 @@ function DokanDataViews< Item >( props: DataViewsProps< Item > ) {
     );
 }
 
-// Callers render these directly, so they must survive the wrap.
-export default Object.assign( DokanDataViews, {
-    Pagination: PluginUIDataViews.Pagination,
-    Layout: PluginUIDataViews.Layout,
-    Search: PluginUIDataViews.Search,
-    Filters: PluginUIDataViews.Filters,
-    BulkActionToolbar: PluginUIDataViews.BulkActionToolbar,
-} );
+/*
+ * Callers render these directly, so they must survive the wrap — but only through lazy getters.
+ * `@wedevs/plugin-ui` is externalised to the `window.dokan.pluginUI` global, which not every
+ * page this module ships to defines, so reading it while the module initialises crashes the
+ * whole bundle (getdokan/dokan-pro#6099). The getters defer each read to first use, by which
+ * point rendering a DataViews guarantees plugin-ui is loaded.
+ */
+export default Object.defineProperties( DokanDataViews, {
+    Pagination: { enumerable: true, get: () => PluginUIDataViews.Pagination },
+    Layout: { enumerable: true, get: () => PluginUIDataViews.Layout },
+    Search: { enumerable: true, get: () => PluginUIDataViews.Search },
+    Filters: { enumerable: true, get: () => PluginUIDataViews.Filters },
+    BulkActionToolbar: {
+        enumerable: true,
+        get: () => PluginUIDataViews.BulkActionToolbar,
+    },
+} ) as typeof PluginUIDataViews;


MdAsifHossainNadim:
## Verdict: Approve ✅

Reviewed against `develop`, on a local build whose `?ver=` hashes match the ones in getdokan/dokan-pro#6099, plus an audit of the compiled bundles.

### Why this fix is correct

`@wedevs/plugin-ui` compiles to the `window.dokan.pluginUI` global. On `admin.php?page=dokan` the actual print order is:

`plugin-ui.js` (#113) → `components.js` (#150) → `var dokan = {…}` from `wp_localize_script( 'dokan-vue-vendor', 'dokan', … )` (#155) → `product-editor-utils.js` (#221), `frontend.js` (#226), `dokan-admin-dashboard.js` (#232)

So those three bundles initialise *after* `window.dokan` has been replaced wholesale. #3382's top-level cast and eager `Object.assign` both read `.DataViews` off that now-missing namespace during module initialisation — an uncaught throw that aborts the entire module factory. Hence exactly three dead bundles, and no fourth: `components.js` initialises before the overwrite.

Moving the cast inside the component and forwarding the five statics through `Object.defineProperties` getters removes every module-scope property read. The compiled output now holds only the harmless `e.exports = window.dokan.pluginUI` external require at module scope — verified in all three bundles. On `page=dokan`: no `DataViews` TypeErrors. No regression of #3382 — `page=dokan-dashboard` and `/dashboard/new/#/withdraw-requests` still render their DataViews lists with the `dokan-lite` labels. `tsc` baseline unchanged (961 errors, byte-identical set on both branches); ESLint clean.

Leaving the `wp_localize_script` clobber alone here is also the right call: it is a decade-old public object with three call sites and its own blast radius, and fixing it would not subsume this change anyway.

### Two non-blocking notes

**1. The comment promises more than the code delivers.** The getters defer the `.DataViews` read, but not the external itself: webpack requires module `7512` eagerly (`var c = n(7512)`) and caches its exports, so `c` is a *snapshot* of `window.dokan.pluginUI` taken at bundle-init. If that snapshot is `undefined`, the getter still throws when it is first read, and a later restore of the global cannot help it. Not reachable today — no React DataViews mounts on `page=dokan` — but `DokanDataViews.tsx:51` currently reads:

> The getters defer each read to first use, by which point rendering a DataViews guarantees plugin-ui is loaded.

Suggest rewording that (and the matching lines in both `src/externals/*.js` headers and the skill note) to what the change actually guarantees: a missing global no longer takes the whole bundle down with it.

**2. The statics are now non-configurable.** `Object.defineProperties` defaults to `configurable: false`, so `Object.assign( DataViews, { Pagination } )` throws where it worked before #3382, when these were plain data properties on plugin-ui's own object. Adding `configurable: true` to the five descriptors restores that, at no cost.

### Follow-up worth filing separately

`includes/Assets.php:103` should merge rather than replace, the way `includes/Admin/Dashboard/Dashboard.php:416` already does:

```php
wp_add_inline_script( $handle, 'window.dokanAdminDashboard = ' . wp_json_encode( $data ), 'before' );
```

One caveat for that PR: `wp_localize_script` stringifies scalars (`1` → `"1"`) and `wp_json_encode` does not, so it needs a sweep of the legacy Vue consumers first.

Also worth folding in there: `Dashboard::enqueue_scripts()` enqueues the full React admin bundle set on both `dokan` and `dokan-dashboard`, so ~2 MB of JS loads on the legacy Vue page where the SPA never mounts. Dropping `dokan-admin-dashboard` from `page=dokan` would remove this failure surface entirely.

PR #3391, opened 2026-08-25
fix: declare WooCommerce 11.0 compatibility

akzmoudud:
### All Submissions:

* [x] My code is tested
* [x] I've included related pull request(s)

> **Scope reduced.** The vendor balance / refund work that was previously in this PR has been **moved to its own pull request**, so the money code gets its own review and its own revert boundary — as raised in review. This PR is now the header bump only.

Bumps `WC tested up to` from `10.4.3` to `11.0.1`.

Both plugin headers declared `10.4.3` while sites are already running WooCommerce 11.x, so admins see the incompatibility notice on **Plugins → Installed Plugins** today. Nothing is actually broken by it — it is a stale declaration.

Feature declarations (`custom_order_tables`, `cart_checkout_blocks`) are unchanged and still correct. No new `FeaturesUtil::declare_compatibility()` call is needed — neither WooCommerce 11.1 feature uses that opt-in.

### Related Pull Request(s)

* Dokan Pro: https://github.com/getdokan/dokan-pro/pull/6108 — the remaining WooCommerce 11 fixes (CSV images, GTIN search, vendor order meta, booking line items)
* Refund / vendor balance work: https://github.com/getdokan/dokan-pro/pull/6117 — refunds created outside Dokan's own flow (WC REST API, WP-CLI, other plugins)

### Closes

* Refs getdokan/plugin-internal-tasks#2314 (item 5)

1. Run WooCommerce 11.x
2. Go to **Plugins → Installed Plugins**
3. Confirm the Dokan row no longer shows a WooCommerce incompatibility notice

### Changelog entry

*Fix — Declare WooCommerce 11.0 compatibility*

The plugin header declared it was tested up to WooCommerce 10.4.3, so sites running WooCommerce 11.x were shown an incompatibility notice on the plugins screen even though nothing was broken. The declaration is updated to 11.0.1.

### Compatibility testing

Verified on WooCommerce **8.5.2, 9.1.4, 10.4.3, 11.0.1 and 11.1.0-beta.1** — no fatals, admin and vendor panels clean on all five, both plugins stayed active throughout.


akzmoudud:
## Cross-version testing

Paired with getdokan/dokan-pro#6108. Since the refund gate depends on how Lite and Pro interact across independently versioned releases, it was verified from the declared floor (`WC requires at least: 8.5.0`) up to the 11.1 beta rather than on a single version.

| WooCommerce | Automated | Admin + vendor panels | Manual | debug.log |
|---|---|---|---|---|
| 11.1.0-beta.1 | 16/16 | 14/14 pages 200 | ✅ | 0 fatals |
| 11.0.1 | 16/16 | 14/14 pages 200 | ✅ | 0 fatals |
| 10.4.3 | 16/16 | 14/14 pages 200 | ✅ | 0 fatals |
| 9.1.4 | 16/16 | 14/14 pages 200 | ✅ | 0 fatals |
| 8.5.2 | 16/16 | 14/14 pages 200 | ✅ | 0 fatals |

Same plugin code throughout (dokan-lite 5.0.16 + dokan-pro 5.0.15), HPOS enabled. No fatals, no database errors, no Dokan errors. Both plugins stayed active on every version.

### Unit tests

`tests/php/src/Orders/RefundHandlerProGateTest.php` — **7/7 pass** via wp-env. It covers both version-skew branches (Pro class absent, and Pro class present but without `is_dokan_refund()`), marked vs unmarked refunds, refund-ID pass-through, the Lite-only path, and an unresolvable refund ID.

Full Lite suite: **433 tests, 431 pass**. The 2 failures are in `OrderStatusChangeTest` (`pending -> cancelled`) and were confirmed pre-existing by re-running them on a clean `develop` with this branch's changes stashed.

### The behaviour this fixes, measured

Before, on a real sub-order:

```
BEFORE: order total=15.00 refunded=0  status=processing
        vendor_balance: {"debit":"15.0000","status":"wc-processing"}

REST refund of the full 15.00 -> 201 created

AFTER:  order total=15.00 refunded=15 status=refunded
        vendor_balance: {"debit":"15.0000","status":"wc-refunded"}   <-- untouched
        dokan_refund rows: 0
        still counted as WITHDRAWABLE after full refund: YES -> 15.0000
```

After, the same REST path records the vendor's share:

```
{"trn_type":"dokan_orders","debit":"50.0000","credit":"0.0000","status":"wc-processing"}
{"trn_type":"dokan_refund","debit":"0.0000","credit":"10.0000","status":"approved"}
```

And a refund created through Pro's own flow is still handled exactly once:

```
marker meta: 'dokan'
is_dokan_refund(): true
credit before / after: 10 / 10
=> Lite skipped the Pro-created refund, no double adjustment
```

### Note for review

The gate **fails closed**. If Pro is active but predates `Refund::is_dokan_refund()`, Lite stands down exactly as it does today rather than risking a second debit. Lite and Pro ship independently, so new Lite paired with older Pro is a real deployment state — that path is covered by two of the seven unit tests.

The two PRs are therefore safe to merge in **either order**.

MdAsifHossainNadim (changes requested):
Reviewed against the paired Pro PR (getdokan/dokan-pro#6108) and getdokan/plugin-internal-tasks#2314, tracing the refund path through Lite, Pro and WooCommerce core. Everything below is sourced from code with `file:line` — I could not execute the suite (no Docker in my environment), so the one reproduction below is written as a test for you to run rather than a result I'm reporting.

**The mechanism is right and failing closed on version skew is the correct call.** Holding on #1 because the fix does not close the leak on the paths that produce it most often, and this PR is marked as closing item 1 of #2314.

---

## :red_circle: 1. Refunds with no line items still leak — including the wp-admin "set status to Refunded" path

`OrderRefundCommission::get_vendor_total_refund()` is built entirely from `$refund->get_items()` (`OrderCommission::calculate_for_refund()`, `includes/Commission/OrderCommission.php:144`). A refund created **without** `line_items` has no line items, no tax items and no shipping items, so the vendor share resolves to `0.0` and `RefundHandler.php:249` bails on `$vendor_earning_refund > 0`. **No balance row is written.**

The PR body treats this as a testing detail ("line items are required, since commission is allocated per item…"). It is not a testing detail — it is the majority of real refund traffic:

| Caller | Passes `line_items`? | Reference |
|---|---|---|
| `wc_order_fully_refunded()` — fires on **every** transition to `wc-refunded` | `'line_items' => array()` | `woocommerce/includes/wc-order-functions.php:906,928` |
| `wc/v2` REST refunds | never — the schema marks it `readonly` | `class-wc-rest-order-refunds-v2-controller.php:292-300` |
| `wc/v3` REST refunds | only when the caller supplies them; `{"amount":"15.00"}` alone is valid | `Version3/class-wc-rest-order-refunds-controller.php:55-63` |
| WooCommerce Stripe, refund initiated from the Stripe Dashboard | `order_id`, `amount`, `reason` only | `woocommerce-gateway-stripe/includes/class-wc-stripe-webhook-handler.php:778` |

The wp-admin path is the one that hurts. Admin sets a multi-vendor **parent** order to Refunded:

1. `Hooks::on_order_status_change()` cascades `update_status( 'refunded' )` to every sub-order (`includes/Order/Hooks.php:125-138`; `wc-processing → wc-refunded` and `wc-completed → wc-refunded` are both whitelisted at `:208-209`).
2. Each sub-order transition fires WooCommerce's `wc_order_fully_refunded()` → an **item-less** full refund.
3. Lite now handles it (unmarked, `DOKAN_SUBORDER_REFUND`) → vendor share `0.0` → nothing written.
4. The same handler then stamps the balance row `status = 'wc-refunded'` (`Hooks.php:172-181`) while `debit` still holds the pre-refund gross.
5. `dokan_withdraw_get_active_order_status()` **hardcodes** `'wc-refunded'` as an active status (`includes/Withdraw/functions.php:357`).

That is #2314 item 1 verbatim, reached through wp-admin, on the branch that closes it.

`Hooks::manage_refunded_for_order()` is the pre-existing compensator for this, but it is gated on `$_POST['post_type'] === 'shop_order'` plus a `dokan_change_status` nonce (`Hooks.php:294,307`) — inert under HPOS, inert over REST, and inert for the programmatic sub-order cascade.

Reproduction, as a test to drop into `RefundHandlerProGateTest`:

```php
public function test_item_less_full_refund_still_leaves_the_order_withdrawable() {
    global $wpdb;

    $order_id = $this->create_single_vendor_order();
    $order    = wc_get_order( $order_id );

    // Exactly what wc_order_fully_refunded() passes.
    $refund = wc_create_refund(
        [
            'order_id'   => $order_id,
            'amount'     => $order->get_total(),
            'reason'     => 'Order fully refunded.',
            'line_items' => [],
        ]
    );

    $this->assertNotWPError( $refund );

    $credit = (float) $wpdb->get_var(
        $wpdb->prepare(
            "SELECT COALESCE( SUM( credit ), 0 ) FROM {$wpdb->prefix}dokan_vendor_balance WHERE trn_id = %d AND trn_type = 'dokan_refund'",
            $order_id
        )
    );

    $this->assertGreaterThan( 0.0, $credit ); // expected to fail on this branch
}
```

**Suggested direction:** when the refund carries no items, prorate instead of returning `0.0` — the vendor's net earning for the order × `abs( $refund->get_total() ) / $order->get_total()`. If that is too large for this PR, that is a fine call — but then please say so explicitly in the body and leave #2314 item 1 open rather than closing it.

## :orange_circle: 2. Credits are keyed by order ID, never reversed, and not idempotent

`insert_into_balance_table()` writes `'trn_id' => $order->get_id()` (`RefundHandler.php:264`) — the refund ID appears nowhere in the row. Nothing in Lite or Pro hooks `woocommerce_refund_deleted` (zero occurrences in either plugin). So deleting a WooCommerce refund leaves the credit standing and the vendor permanently short, and a repeated `woocommerce_order_refunded` writes a second credit that nothing can detect.

This predates the PR, but the PR moves it from "only Pro's own flow writes here" to "every third-party refund writes here" on Pro sites. At minimum worth a tracked follow-up; ideally the refund ID goes into the row.

## :orange_circle: 3. Three writers now share this table with no reconciliation

`RefundHandler::insert_into_balance_table()`, `Hooks::manage_refunded_for_order()` (`Hooks.php:290` — credits the **whole** net earning, not the refunded portion) and Pro's `Refund::update_vendor_balance()` all insert `trn_type = 'dokan_refund'` rows for the same order. `manage_refunded_for_order()`'s duplicate check is presence-only (`Hooks.php:312-321`), so a partial credit written earlier by `handle_refund()` makes it skip entirely when the order later goes fully refunded, leaving the remainder un-credited. This PR widens the first writer without addressing that interaction.

## :yellow_circle: 4. Hard-coded Pro class name in Lite

`get_pro_refund_class()` returning `'\WeDevs\DokanPro\Refund\Refund'` is the **first** hard-coded Pro FQCN in `includes/` — there are currently zero. All 16 existing Lite → Pro touchpoints go through the `dokan_pro()` accessor. The refund path is already hook-driven end to end (`dokan_vendor_earning_in_refund`, `dokan_refund_should_insert_into_vendor_balance`, `dokan_refund_adjust_vendor_balance`), so a filter fits the codebase and keeps the identical fail-closed semantics:

```php
// Pro adjusts the refunds it creates; anything else still needs handling here.
if ( apply_filters( 'dokan_refund_is_handled_externally', dokan()->is_pro_exists(), $refund_id, $order_id ) ) {
    return;
}
```

An older Pro registers no filter → default `true` → Lite stands down, exactly as today. A current Pro returns `false` for unmarked refunds. That also removes `get_pro_refund_class()` (a seam whose only reason to exist is the test), removes the `method_exists()` autoload probe on every refund, removes the duplicate `wc_get_order( $refund_id )` (once inside `is_dokan_refund()`, again at `RefundHandler.php:41`), and lets gateways and wePOS opt out of Lite's handling too.

## :yellow_circle: 5. The tests cover the seam, not the behaviour

Five of the seven tests reach `should_handle_while_pro_active()` through reflection on an anonymous subclass, so they verify the helper rather than the contract. Nothing asserts *Pro active + marked refund → no row written* or *Pro active + unmarked refund → row written*, and the branch at `RefundHandler.php:38` is never executed. `test_gate_passes_the_refund_id_to_pro` asserts an implementation detail of a call Lite makes into Pro. With the filter seam in #4 all of this collapses to an `add_filter()` in the test plus a database assertion.

Separately, `assertGreaterThan( 0.0, $credit )` in `test_lite_only_site_still_records_the_refund_credit` passes whether the credit is the correct vendor share or the entire order total — please assert the expected amount.

## :yellow_circle: 6. REST/CLI refunds will not appear in Dokan's refund ledger

`wp_dokan_refund` is only written by `Refund::save()` (`dokan-pro/includes/Refund/Refund.php:442`). After this change the vendor's balance drops with no matching row, so Dokan → Refunds and the vendor's refund list show nothing explaining it. Reasonable as a deliberate limitation — worth one line in the PR body.

## :white_circle: Smaller

- **Pro's marker is a process-global flag** (dokan-pro#6108). `self::$creating_dokan_refund` stays `true` for the whole `wc_create_refund()` call, including the nested `update_status()` hooks, so anything creating a refund inside that window is stamped as Dokan's and then skipped by Lite *and* never adjusted by Pro. I could not find a live path that reaches it — `on_sub_order_change()` only ever promotes a parent to *completed* (`Hooks.php:426-429`) — so this is hardening, not a proven leak. Keying `mark_dokan_refund()` to `$args['order_id']` rather than a bare bool would close it.
- **The PR body carries Pro-only content.** The "Compatibility testing" section explains the GTIN column guard and the hidden order-item meta keys; neither ships here. Same for the branch name `fix/wc-11-compatibility` on a two-file change. Trimming it to what Lite actually contains would help whoever reads this at release time.
- **Pre-existing, now far more visible:** `RefundHandler.php:253-257` inverts the reason — when there *is* a reason it is overwritten with a generic string, and when there is not, `perticulars` is left empty. REST refunds almost always carry one. Follow-up, not this PR.
- **Widened blast radius:** `dokan_get_seller_id_by_order()` falls back to the product's `post_author` (`includes/Order/functions.php:347-353`), and admins hold `dokandar`. A refund on a plain admin-owned order now writes a `dokan_refund` credit against the admin's "vendor" balance on Pro sites where nothing was written before. Probably harmless — worth a sanity check.

---

**What is genuinely good here:** the diagnosis in #2314 is excellent, failing closed on version skew is the right trade (double-debiting really is worse than the original leak), routing both Pro `wc_create_refund()` call sites through a single helper is thorough — those are confirmed to be the only two in the plugin — and the header bump is correct.

akzmoudud:
Thanks — this was a better review than the PR deserved, and two findings were things I had seen and misread rather than missed. Consolidated response to all six points.

## 🔴 1. Item-less refunds — confirmed, fixed

You were right, and I had the evidence in front of me. I observed the zero vendor share during testing and wrote it off in the PR body as *"line items are required, since commission is allocated per item"* — presenting the bug as a testing note.

Every link verified:

| Claim | Verified |
|---|---|
| `wc_order_fully_refunded()` passes `line_items => array()` | `wc-order-functions.php:921` |
| `calculate_for_refund()` iterates `$refund->get_items()` | `OrderCommission.php:143` |
| Bail on zero share | `RefundHandler.php:249` |
| Parent → sub-order cascade | `Order/Hooks.php:125-138` |
| `wc-refunded` hard-coded active | `Withdraw/functions.php:357` |

Reproduced on the wp-admin path — setting a multi-vendor parent order to Refunded:

```
sub #373  refund #2004  total=-20.00  items=0   dokan_refund credit rows: 0
   balance row: {"debit":"20.0000","status":"wc-refunded"}  -> still payable 20.00
```

**Why my first re-run looked like you were wrong:** it showed credits being written. A backtrace found the cause —

```
#4 wc_create_refund   /woocommerce-germanized/includes/class-wc-gzd-order-helper.php:625
#5 WC_GZD_Order_Helper->create_refund_with_items
```

WooCommerce Germanized hooks `woocommerce_order_status_refunded` at priority 5 and supplies line items, pre-empting WooCommerce's item-less path. It is active on my test site, so every full-refund test was silently getting populated items. Deactivated: `items=0`, zero credit rows.

**Fix:** refunds with no items prorate — the vendor's earning × the share of the order total refunded. The earning is recalculated with refund adjustment **disabled**, because `VendorBalanceUpdateHandler` rewrites the order tables at `woocommerce_update_order` priority 80, before `woocommerce_order_refunded`; using the stored `net_amount` would read an already-zeroed value.

Since this is money, run with a **20% commission** so the earning differs from the order total and proration cannot look right by coincidence:

| Scenario | Order | Vendor earning | Credited |
|---|---|---|---|
| Item-less full | 100.00 | 80.00 | **80.00** (not 100) |
| Item-less half | 100.00 | 80.00 | **40.00** |
| Split 30% + 70% | 100.00 | 80.00 | **80.00** — never over |
| Item-based (50 of 100) | 100.00 | 80.00 | **40.00** — unchanged |
| With shipping | 115.00 | 95.00 | **95.00** |
| Over-refund attempt | 100.00 | 80.00 | **0.00** — WC rejects |

Your suggested test is in the suite and fails on the unfixed branch rather than passing vacuously — verified by stashing the production change.

## 🟠 2. Idempotency and reversal

**Idempotency — added, but I overstated it.** I originally reported *"a duplicate `woocommerce_order_refunded` credited twice (80 → 160)"* as observed behaviour. It was not: I produced the 160 by calling `do_action()` myself in a test, then reported it without checking whether anything real can. It cannot, as far as I can find — `wc_create_refund()` at `wc-order-functions.php:743` is the only production caller and fires once per refund, and no bundled plugin re-fires it. The guard is defensive hardening on a property the code should have had since 2017, not a fix for a failure anyone has hit.

You applied exactly the right standard to your own finding about Pro's process-global flag — *"I could not find a live path that reaches it ... so this is hardening, not a proven leak"* — and I failed to apply it to mine.

**One thread I could not close:** Pro's approval path fires the balance adjustment at `Refund.php:1064`, and I did not find an explicit pending-status guard preventing the same request being approved twice. If that path can run twice, the duplicate becomes real.

**Reversal — reproduced, still open.** Manual refund, no money moved, then deleted:

```
1. order created:   debit=80.00  credit= 0.00  -> vendor owed 80.00
2. manual refund:   debit=80.00  credit=80.00  -> vendor owed  0.00
                    (refunded_payment = false)
3. refund DELETED:  debit=80.00  credit=80.00  -> vendor owed  0.00

WooCommerce: total_refunded = 0.00   (self-corrects)
Dokan:       credit         = 80.00  (stranded)  -> vendor short $80
```

Confirmed exactly as you described. WooCommerce is immune because it never stores a derived total — `get_total_refunded()` sums the refunds that currently exist, so a deleted one stops counting.

I also traced the origin, having told you 4.0.0 earlier, which was wrong: the refund ID has **never** been stored, back to `7dd7e9ad` (2017, v2.5.3).

On *"ideally the refund ID goes into the row"* — the schema part is trivial (`dbDelta` at `Installer.php:457`, plus the `Upgrades/V_X_Y_Z.php` pattern). The work is that **five writers** create `dokan_refund` rows: Lite's `RefundHandler` and `Order/Hooks.php`, plus Stripe Express, Razorpay and PayPal Marketplace. If only mine populates it, reversal appears to work and silently fails for gateway refunds — which are exactly the ones where `refunded_payment = true` and reversing would be wrong anyway. Reversal must therefore be conditional on that flag, since `WC_AJAX::delete_refund()` deletes without checking it and without any gateway call.

I lean towards a separate PR, but will do it here if you prefer.

## 🟠 3. Three writers, no reconciliation

I skipped this point when reading. All three observations verify — `manage_refunded_for_order()` credits the whole net earning (`Hooks.php:323-330`), its duplicate check is presence-only (`:312-321`), and three writers share the table.

**The specific leak is closed by the proration change, because of hook ordering:**

```
class-wc-order.php:457   woocommerce_order_status_refunded  -> wc_order_fully_refunded() -> creates refund -> handle_refund() credits
class-wc-order.php:478   woocommerce_order_status_changed   -> manage_refunded_for_order() -> row exists -> returns
```

The refund is created **first**. Before this PR, `handle_refund()` resolved the item-less refund to `0.0` and wrote nothing, then `manage_refunded_for_order()` skipped because an earlier partial row existed → remainder lost, as you said. After, the remainder is credited and the third writer correctly stands down. Verified: item-based 40.00 then item-less 60.00 on a 100.00 order credits **80.00 total**.

The structural half stands: `manage_refunded_for_order()` would still credit the *whole* earning if it ever ran with no prior row, and it is `$_POST`/nonce-gated to the classic order screen, so inert under HPOS and REST. A landmine, not a live bug — worth tracking with the reversal work, since both stem from rows carrying no reference to what created them.

## 🟡 4. Hard-coded Pro FQCN — adopted your design

```php
if ( apply_filters( 'dokan_refund_is_handled_externally', dokan()->is_pro_exists(), $refund_id, $order_id ) ) {
    return;
}
```

Pro registers `Hooks::is_refund_handled_here()`. Fail-closed semantics identical, and it removes `get_pro_refund_class()`, the `method_exists()` autoload probe and the duplicate `wc_get_order()`.

## 🟡 5. Tests — rewritten to assert behaviour

They now drive real refunds through `wc_create_refund()` and assert database state: the filter contract, item-less full and partial refunds, the item-based path, duplicate events, and an unresolvable refund ID. `test_lite_only_site_still_records_the_refund_credit` asserts the expected amount rather than `> 0`.

**8/8 pass.** Full suite: 434 tests, 2 failures, both `OrderStatusChangeTest` (`pending → cancelled`), confirmed pre-existing on clean `develop`.

## 🟡 6. REST/CLI refunds absent from the Dokan ledger

Correct — now stated in the PR body under *Known limitations*, alongside the deletion gap and the missing migration.

## ⚪ Smaller

* **Pro's process-global flag** — fair. Both `wc_create_refund()` calls are sequential rather than nested, and I could not find a live path either, but keying to `$args['order_id']` is stricter.
* **PR body carrying Pro-only content** — fixed, it is Lite-only now.
* **`perticulars` reason inversion** and **`dokan_get_seller_id_by_order()` falling back to `post_author`** — both real, both pre-existing, neither touched here.

## On scope

Worth being straight: splitting was considered and rejected **before these PRs were opened**. I recommended roughly one PR per finding and specifically argued the refund pair deserved its own revert boundary; the team chose one Lite and one Pro PR. So your objection landed where it was expected to.

That does not make it wrong, and one thing has changed: the refund half has grown a lot during review — proration, idempotency, the filter seam, its own tests. Your revert-boundary argument is stronger now than when the decision was made. My preference is not to split mid-review, but **if you want the refund pair separated before merge, say so and I will do it.** Better you overrule it explicitly than have it stand by default.

---

## Decisions on the two open questions

Both answered by the team, recording them here so nothing waits on us:

**Refund ID in the balance row — not in this PR.** Staying with the per-refund meta stamp. It closes the duplicate-credit gap, which is the one that can fire today. Reversal on deletion needs the column plus all five writers (Lite's `RefundHandler` and `Order/Hooks.php`, Stripe Express, Razorpay, PayPal Marketplace) and the `refunded_payment` condition, or it silently fails for gateway refunds. That is its own piece of work, and it stays listed under *Known limitations* in the description rather than being half-done here.

**Splitting the refund pair — not splitting.** The one-Lite/one-Pro shape was a deliberate decision before these PRs were opened and the team is keeping it. Your revert-boundary argument is noted and fair; it is a scoping trade-off we are accepting knowingly rather than an oversight.


---

## Update — the refund work has been split out of this PR

Acting on your *Scope & process* point: the vendor balance work is moving to **its own pull request**, so the money code gets its own review and its own revert boundary. `includes/Order/RefundHandler.php` is back to its state on develop, and the test file has gone with it.

**This PR is now the header bump only** (`WC tested up to: 10.4.3 → 11.0.1`). Everything discussed above about proration, idempotency, the filter seam and the reversal gap carries over to the refund PR — including the two decisions recorded above and the reproduction of the deleted-refund case.


---

# ⚠️ RETRACTION — the refund leak was not a bug

QA has tested a **real gateway refund** and the vendor balance adjusts correctly. Everything above about a money leak is withdrawn. The refund work has been removed from this PR and the planned follow-up PR is cancelled.

## What was actually wrong

Two accurate observations:

1. Lite's `handle_refund()` bails when Pro is active — true
2. Pro hooks refunds at admin-ajax only — true

And one conclusion that did not follow: *"therefore nothing else handles refunds."*

**Five gateway modules handle their own refunds** through Dokan's refund-request system:

```
modules/stripe-express/     add_action( 'dokan_refund_request_created', ... )
modules/razorpay/           add_action( 'dokan_refund_approve_before_insert', ... )
modules/paypal-marketplace/ add_filter( 'dokan_refund_approve_vendor_refund_amount', ... )
modules/mangopay/
modules/stripe/
```

I never searched for other handlers before concluding there were none.

## Why the "reproductions" looked convincing

Every test — mine and the ones in #2314 — used a **manual** refund:

```json
{ "amount": "15.00", "api_refund": false }
                     ^^^^^^^^^^^^^^^^^^
```

`api_refund: false` means no money leaves the gateway. WooCommerce says so itself on that action: *"Order status set to refunded. To return funds to the customer you will need to issue a refund through your payment gateway."* The refund record carries `refunded_payment = false`.

So the customer was never actually refunded. The vendor keeping their earning was correct behaviour, not a leak. That detail sat in the reproduction steps from the start and neither of us caught it.

Verified identical on WooCommerce 10.4.3 and 11.0.1, with and without WooCommerce Germanized — so it was never a WooCommerce 11 compatibility matter either.

## What this invalidates

* The item-less refund analysis, the proration fix and its money audit — the premise is gone
* The idempotency guard — already flagged as defensive rather than reproduced
* The refund-deletion finding — my reproduction of it depended on the proration fix being active, so it needs re-testing on unmodified code before anyone treats it as real

## What still stands

* **`wc-refunded` is hardcoded as a payable withdraw status** (`Withdraw/functions.php:357`), overriding the admin's own setting. Verified, unexplained, and worth asking about separately — but not a WooCommerce 11 matter.
* This PR itself: the header bump, unaffected by any of the above.

Apologies for the noise on what turned out to be correct behaviour. The lesson worth recording: the analysis was specific — file names, line numbers, real hooks — and that specificity made it read as verified when it was only detailed. What settled it was refunding through a real gateway and watching the money, which is the one thing none of the written analysis did.

akzmoudud:
## Correction to my previous comment — the idempotency claim was overstated

In my reply above I wrote that *"a duplicate `woocommerce_order_refunded` credited twice (80 → 160, driving the balance negative)"*. That reads as observed production behaviour. **It was not.** I produced the 160 by calling `do_action( 'woocommerce_order_refunded', ... )` myself inside a test, then reported the result without checking whether anything real can do that.

It cannot, as far as I can find:

* `wc_create_refund()` at `wc-order-functions.php:743` is the **only** production caller, and it fires once per refund created
* no bundled plugin on the test site re-fires it
* before this PR, Lite bailed out entirely on Pro sites, so its insert never ran there at all

So the accurate description is: **the insert has never had duplicate protection, and I have not demonstrated a path that exploits it.** The guard is defensive hardening on a property the code should have had since 2017, not a fix for a failure anyone has hit. The PR description now says this.

You applied exactly the right standard to your own finding about Pro's process-global flag — *"I could not find a live path that reaches it ... so this is hardening, not a proven leak"* — and I failed to apply it to mine. Thanks for the implicit example; the challenge that surfaced it came from our own QA rather than from me re-checking.

**One thread I could not close.** Pro's approval path fires the balance adjustment at `Refund.php:1064`, and I did not find an explicit pending-status guard preventing the same refund request being approved twice. I have not traced it far enough to say either way. If that path *can* run twice, the duplicate becomes real rather than theoretical — worth a look from someone who knows that flow.

---

## Your point 2 — the deletion half is now reproduced

Since I had been asserting rather than testing, I ran it. Manual refund, no money moved, then deleted:

```
1. order created:   debit=80.00  credit= 0.00  -> vendor owed 80.00
2. manual refund:   debit=80.00  credit=80.00  -> vendor owed  0.00
                    (refunded_payment = false)
3. refund DELETED:  debit=80.00  credit=80.00  -> vendor owed  0.00

WooCommerce: total_refunded = 0.00   (self-corrects)
Dokan:       credit         = 80.00  (stranded)  -> vendor short $80
```

**Confirmed exactly as you described it.** The vendor loses $80 they legitimately earned, silently.

Worth noting *why* WooCommerce is immune: it never stores a derived total. `get_total_refunded()` sums the refund records that currently exist, so a deleted refund simply stops counting. Dokan stores a snapshot instead, which is faster to query but carries the obligation to keep it in step — the obligation that has not been met since the initial release.

I also traced the origin, since I had told you 4.0.0 earlier and that was wrong: the refund ID has **never** been stored in the row, going back to `7dd7e9ad` (2017, v2.5.3). `grep -S "'refund_id'"` across all history in `includes/Order/` returns nothing.

## On doing it properly now

Given the above, my honest read on your *"ideally the refund ID goes into the row"*:

* The schema part is genuinely trivial — the table is created with `dbDelta` (`Installer.php:457`), so adding a nullable `refund_id` column is one line, and there is an established `Upgrades/V_X_Y_Z.php` pattern besides
* The work is that **five writers** create `dokan_refund` rows: Lite's `RefundHandler` and `Order/Hooks.php`, plus Stripe Express, Razorpay and PayPal Marketplace. If only mine populates the column, reversal appears to work and silently fails for gateway refunds — which are exactly the ones where `refunded_payment = true` and reversing would be wrong anyway
* Existing rows would carry `NULL`, so reversal only ever works for refunds created after the upgrade

So I still lean towards a separate PR, but the decision is yours and I will do it here if you prefer. If you want it separate, I would rather you say so on the record than have it become a paragraph nobody actions.

MdAsifHossainNadim (approved):

PR #3394, opened 2026-09-01
fix(product-editor): stop the vendor's first gallery image becoming the featured image

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

Saving a product from the new React product editor with **gallery images but no featured image** moved the vendor's first gallery image into the featured slot and dropped it from the gallery. With a single gallery image the gallery ended up empty. This is the defect behind the recurring report *"vendors are unable to add more than one photo to the product image gallery."*

`PayloadResolver::resolve_images()` flattened `image_id` and `gallery_image_ids` into WooCommerce's positional `images` array and relied on array order alone. `ProductControllerV3` extends `WC_REST_Products_Controller` without overriding `set_product_images()`, so WooCommerce's own implementation runs — and it takes the featured image from **array index 0** (`class-wc-rest-products-controller.php:703`). With no featured image, index 0 is a gallery image. WooCommerce's v3 `images` schema has no `position` field, so the flat array simply cannot express "gallery images, no featured image".

The resolver now tags each entry with the position the vendor chose — the same convention `ProductController` (v1) already splits on — and the v3 controller reapplies that split after WooCommerce has consumed `images`. Payloads Dokan did not resolve carry no positions and are left to WooCommerce untouched.

Emitting `images` even when empty also fixes a second bug in the same method: a vendor who removed **every** image sent no `images` key at all, so the old images survived.

### Closes

* Closes getdokan/plugin-internal-tasks#2343

1. WP Admin → Dokan → Settings → Appearance → **Vendor Product Editor** = *New UI*.
2. Vendor dashboard → Products → **Add new product**. Fill Title, Price, Description. **Leave Feature Image empty.**
3. Gallery Image → add **two** images (pick one → *Use this media*, then repeat for a second).
4. Confirm the editor shows Feature Image empty and 2 gallery thumbnails. Save.
5. Reopen the product.

**Before:** Feature Image shows the vendor's first gallery image; gallery has only the second.
**After:** Feature Image stays empty; gallery keeps both.

Verified against `dokan/v3/products` on WooCommerce 11.0.1:

| Vendor sets | Before | After |
|---|---|---|
| 2 gallery, no featured | `featured=A, gallery=[B]` | `featured=(none), gallery=[A,B]` |
| 1 gallery, no featured | `featured=B, gallery=[]` | `featured=(none), gallery=[B]` |
| featured + gallery | correct | unchanged |
| every image removed | no-op | both cleared |
| partial update (price only) | untouched | unchanged |
| `image_id: 0` alone, no gallery key | featured **kept**, gallery kept | featured cleared, gallery **also cleared** |

Because `images` is now always emitted, the two fields compile into one WooCommerce array and a request naming **either** of them describes the product's whole image state — so clearing the featured image alone also clears the gallery. This matches WooCommerce's own all-or-nothing `images` semantics, and the product editor always sends both keys, but it is a behaviour change for a partial API update and is called out in the changelog below.

### Changelog entry

**Fix — product editor no longer loses a gallery image on save**

Previously, saving a product from the new product editor without a featured image moved the vendor's first gallery image into the featured slot and removed it from the gallery; with one gallery image the gallery was emptied. Removing every image also did nothing. Now the featured image and gallery are saved exactly as the vendor set them, and clearing all images works. Note for API clients: a request that sends either `image_id` or `gallery_image_ids` describes the product's whole image state, so sending only one of them clears the other.

### Notes for the reviewer

Three follow-ups deliberately left out of scope:

* `position` survives only because WooCommerce's v3 images schema sets no `additionalProperties`, so WordPress preserves the unknown key. This controller does not override `get_item_schema()`, while the v1 `ProductController` declares `position` explicitly. If WooCommerce ever tightens that schema the split silently stops being applied and the bug returns with no error. Declaring `position` in a v3 schema override would close it, at the cost of deep-merging into WooCommerce's nested schema.

* WooCommerce still attaches index 0 to the product via `post_parent` before the split is corrected, leaving a gallery image parented to the product. Cosmetic (Media Library "Uploaded to" only). Removing it means overriding `set_product_images()` and owning WooCommerce's URL-upload path — considered and rejected as the riskier trade.
* Pro's variation resolver (`dokan-pro/includes/Product/PayloadResolver.php:41`) guards on `! empty( $data['images'] )`, so the cleared-everything case does not reach variations. Pre-existing; needs a Pro-side change.

The legacy vendor form writes through `Product\Manager::update()` and was never affected, so this is a 5.0 rewrite regression. Variations are unaffected — Pro's variation controller extends `WC_REST_Product_Variations_Controller`, not this one.

### DIFF
--- a/includes/ProductEditor/PayloadResolver.php
+++ b/includes/ProductEditor/PayloadResolver.php
@@ -183,6 +183,9 @@ class PayloadResolver {
      * schema sets no `additionalProperties`. ProductControllerV3::apply_image_positions() is the
      * only consumer.
      *
+     * The two fields compile into one WooCommerce array, so a request carrying either of them
+     * describes the product's whole image state — sending only one clears the other side.
+     *
      * @since 5.0.0
      *
      * @param array $data Request body keyed by schema field id.
--- a/includes/ProductEditor/PayloadResolver.php
+++ b/includes/ProductEditor/PayloadResolver.php
@@ -176,32 +176,57 @@ class PayloadResolver {
     /**
      * Transform featured image and gallery image IDs into the WC REST images array.
      *
+     * Every entry carries a `position` — 0 for the featured image, 1..n for the gallery — because
+     * the flat array on its own cannot express a product that has gallery images but no featured
+     * image. This mirrors the convention ProductController (v1) already splits on; WooCommerce's
+     * own v3 images schema has no `position`, and ignores the key while preserving it, since that
+     * schema sets no `additionalProperties`. ProductControllerV3::apply_image_positions() is the
+     * only consumer.
+     *
      * @since 5.0.0
+     *
+     * @param array $data Request body keyed by schema field id.
+     *
+     * @return array
      */
     public function resolve_images( array $data ): array {
-        $images = [];
+        $has_featured = array_key_exists( Elements::FEATURED_IMAGE_ID, $data );
+        $has_gallery  = isset( $data[ Elements::GALLERY_IMAGE_IDS ] ) && is_array( $data[ Elements::GALLERY_IMAGE_IDS ] );
 
-        if ( ! empty( $data[ Elements::FEATURED_IMAGE_ID ] ) ) {
-            $id = $this->extract_image_id( $data[ Elements::FEATURED_IMAGE_ID ] );
-            if ( $id > 0 ) {
-                $images[] = [ 'id' => $id ];
-            }
-            unset( $data[ Elements::FEATURED_IMAGE_ID ] );
+        if ( ! $has_featured && ! $has_gallery ) {
+            return $data;
         }
 
-        if ( isset( $data[ Elements::GALLERY_IMAGE_IDS ] ) && is_array( $data[ Elements::GALLERY_IMAGE_IDS ] ) ) {
+        $images = [];
+
+        if ( $has_featured ) {
+            $id = $this->extract_image_id( $data[ Elements::FEATURED_IMAGE_ID ] );
+            unset( $data[ Elements::FEATURED_IMAGE_ID ] );
+
+            if ( $id > 0 ) {
+                $images[] = [
+                    'id'       => $id,
+                    'position' => 0,
+                ];
+            }
+        }
+
+        if ( $has_gallery ) {
+            $position = 1;
             foreach ( $data[ Elements::GALLERY_IMAGE_IDS ] as $img ) {
                 $id = $this->extract_image_id( $img );
                 if ( $id > 0 ) {
-                    $images[] = [ 'id' => $id ];
+                    $images[] = [
+                        'id'       => $id,
+                        'position' => $position++,
+                    ];
                 }
             }
             unset( $data[ Elements::GALLERY_IMAGE_IDS ] );
         }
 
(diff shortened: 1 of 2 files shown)


MdAsifHossainNadim:
## Verdict: Approve ✅

Reviewed the code, reproduced the bug on `develop` first, and verified the fix on this branch against a live site (WooCommerce 11.0.1, vendor dashboard + `dokan/v3`).

### What the fix does

`PayloadResolver::resolve_images()` now tags every entry with the position the vendor actually chose — `0` for the featured image, `1..n` for the gallery — and `ProductControllerV3::apply_image_positions()` reapplies that split after WooCommerce has consumed `images`. It also emits `images` even when empty, so "the vendor removed everything" is expressible.

### Why it is correct

This is a Dokan bug, not a WooCommerce one, and the fix is placed accordingly. WooCommerce's v3 `images` array is positional by contract — `class-wc-rest-products-controller.php:703` takes the featured image from **array index 0**:

```php
foreach ( $images as $index => $image ) {
    if ( 0 === $index ) {
        $product->set_image_id( $attachment_id );
    } else {
        $gallery[] = $attachment_id;
    }
}
```

Its schema accepts only `id, date_created, date_modified, src, name, alt` — there is no `position`, so array order is the sole signal and a product with gallery images but no featured image simply cannot be expressed. The old resolver flattened two independent fields into that array and discarded which was which, so with no featured image the vendor's first gallery image *was* index 0 and WooCommerce correctly promoted it.

Two things make the chosen approach the right one:

- **It restores a convention Dokan already had.** The v1 `ProductController` splits on `position` (`ProductController.php:1809`) and declares it in its schema. The 5.0 rewrite lost that; this brings v3 back in line rather than inventing something new.
- **It fails safe.** A payload this plugin did not resolve carries no positions, and the whole array is validated before anything is mutated — a partially-tagged payload is handed back to WooCommerce untouched rather than half-applied.

### Testing scope

Bug reproduced on `develop` first, so the fix is provably causal:

| Case | `develop` | This branch |
|---|---|---|
| 2 gallery, no featured | `featured=1313, gallery=[1314]` ❌ | `featured=(none), gallery=[1313,1314]` ✅ |
| 1 gallery, no featured | `featured=1315, gallery=[]` ❌ | `featured=(none), gallery=[1315]` ✅ |
| featured + gallery | correct | unchanged ✅ |
| clear every image | **no-op** ❌ | both cleared ✅ |
| price-only update | untouched | untouched ✅ |

Also covered:

- **Full UI repro from the issue** — created a product in the new editor with two gallery images and no featured image, saved, reopened: Feature Image empty, both thumbnails present.
- **`image_id: 0` alone** — clears the featured image (`develop` silently ignored it) and now also clears the gallery. Documented in the description and changelog; the editor always sends both keys so it is not reachable from the dashboard.
- **PHPCS** — clean on both changed files; the same files are clean on `develop`, so no new violations.

### Notes

- No CI checks ran on this branch, and Lite has no PHPUnit workflow — this silent-data-loss fix ships without an automated regression guard. Worth a test when a harness exists.
- The `position` key survives only because WooCommerce's v3 schema sets no `additionalProperties`. Recorded as a follow-up in the description; if WooCommerce tightens that schema the split stops applying with no error.

PR #3395, opened 2026-09-01
fix(dashboard): recover the vendor dashboard when the current-user resolver fails

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

The vendor dashboard gated its entire render on the `getCurrentUser` resolution reaching `finished`, with no branch for `error`:

```tsx
if ( ! loading || loading?.status !== 'finished' ) {
    return <Skeleton />;
}
```

`getResolutionState()` returns `resolving | finished | **error**`. Because `error` is a terminal state that never becomes `finished`, any failure of that REST call left the dashboard rendering the spinner **forever** — no message, no retry, no way out. This is what vendors report as the dashboard *"just staying on loading"*.

Nothing throws on this path, so no error boundary would ever have caught it. It is a missing `if`.

An error now renders the existing `InternalError` panel (`src/layout/500.tsx`) with a **Try again** button that invalidates the resolution, so the dashboard recovers in place once the endpoint answers — no full page reload.

### Closes

* Closes getdokan/plugin-internal-tasks#2344

Reproduce a genuine failure with a temporary mu-plugin that blocks the REST users endpoint, the way a security plugin does:

```php
add_filter( 'rest_pre_dispatch', function ( $result, $server, $request ) {
    if ( 0 === strpos( $request->get_route(), '/wp/v2/users' ) ) {
        return new WP_Error( 'rest_forbidden', 'Blocked.', [ 'status' => 403 ] );
    }
    return $result;
}, 1, 3 );
```

Then, on the vendor dashboard console, force the current-user resolution to re-run against the blocked endpoint:

```js
wp.data.dispatch( 'core' ).invalidateResolution( 'getCurrentUser', [] );
wp.data.dispatch( 'dokan/core' ).invalidateResolution( 'getCurrentUser', [] );
```

**Before:** endless spinner; `getResolutionState('getCurrentUser')` is `{status:"error", code:"rest_forbidden", 403}` and the UI never reflects it.
**After:** *"We could not load your dashboard"* with **Try Again**; clicking it returns the resolution to `finished` and the dashboard renders normally.

A synthetic trigger works too, without the mu-plugin:

```js
wp.data.dispatch( 'dokan/core' ).failResolution( 'getCurrentUser', [] );
```

### Changelog entry

**Fix — the vendor dashboard no longer hangs on a spinner when it cannot load your account**

Previously, if the request for the vendor's account details failed — a blocked REST endpoint, a 403, a proxy problem — the dashboard showed a loading spinner indefinitely with no explanation and no way to retry. It now shows a clear message with a Try again button that recovers the dashboard in place.

### Notes for the reviewer

The trigger is narrower than "any failed request": WordPress hydrates the current user into the page, so blocking the endpoint alone does not break a plain page load — the cached value has to actually be re-fetched. Documented on getdokan/plugin-internal-tasks#2344. Still worth fixing, since a permanent unrecoverable spinner with no message is the worst possible failure mode.

ESLint clean. `invalidateResolution` was confirmed present on the live `dokan/core` store dispatch.

### DIFF
--- a/src/dashboard/index.tsx
+++ b/src/dashboard/index.tsx
@@ -6,6 +6,7 @@ import Layout from '../layout';
 import getRoutes, { withRouter } from '../routing';
 import { createHashRouter, RouterProvider } from 'react-router-dom';
 import { useDispatch, useSelect } from '@wordpress/data';
+import { store as coreDataStore } from '@wordpress/core-data';
 import coreStore from '@dokan/stores/core';
 import Skeleton from '@src/layout/Skeleton';
 import InternalError from '@src/layout/500';
@@ -37,6 +38,16 @@ const App = () => {
 
     // @ts-ignore — invalidateResolution is a store metadata action, absent from the store's own types.
     const { invalidateResolution } = useDispatch( coreStore );
+    // @ts-ignore — same metadata action on the WordPress store Dokan's resolver reads through.
+    const { invalidateResolution: invalidateCoreData } =
+        useDispatch( coreDataStore );
+
+    // Dokan's resolver awaits resolveSelect( coreDataStore ).getCurrentUser(), which replays that
+    // store's cached failure, so retrying only Dokan's own resolution would never re-issue the request.
+    const retryCurrentUser = () => {
+        invalidateCoreData( 'getCurrentUser', [] );
+        invalidateResolution( 'getCurrentUser', [] );
+    };
 
     const mapedRoutes = routes.map( ( route ) => {
         const WithRouterComponent = withRouter(
@@ -71,7 +82,7 @@ const App = () => {
                     'Your account details could not be read. Try again, and contact the marketplace owner if this keeps happening.',
                     'dokan-lite'
                 ) }
-                onRefresh={ () => invalidateResolution( 'getCurrentUser', [] ) }
+                onRefresh={ retryCurrentUser }
             />
         );
     }
--- a/src/dashboard/index.tsx
+++ b/src/dashboard/index.tsx
@@ -5,9 +5,10 @@ import { __ } from '@wordpress/i18n';
 import Layout from '../layout';
 import getRoutes, { withRouter } from '../routing';
 import { createHashRouter, RouterProvider } from 'react-router-dom';
-import { useSelect } from '@wordpress/data';
+import { useDispatch, useSelect } from '@wordpress/data';
 import coreStore from '@dokan/stores/core';
 import Skeleton from '@src/layout/Skeleton';
+import InternalError from '@src/layout/500';
 import { generateColorVariants } from '@dokan/utilities';
 
 // `<DataViews>` (via @wordpress/dataviews) renders its Actions column header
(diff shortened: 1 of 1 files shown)


src/dashboard/index.tsx
```diff
@@ -34,6 +36,19 @@ const App = () => {
...
+    // @ts-ignore — invalidateResolution is a store metadata action, absent from the store's own types.
+    const { invalidateResolution } = useDispatch( coreStore );
+    // @ts-ignore — same metadata action on the WordPress store Dokan's resolver reads through.
+    const { invalidateResolution: invalidateCoreData } =
+        useDispatch( coreDataStore );
+
+    // Dokan's resolver awaits resolveSelect( coreDataStore ).getCurrentUser(), which replays that
+    // store's cached failure, so retrying only Dokan's own resolution would never re-issue the request.
+    const retryCurrentUser = () => {
+        invalidateCoreData( 'getCurrentUser', [] );
```
kzamanbd, line 51: `retryCurrentUser` clears only one of the two upstream resolutions this resolver depends on. `src/stores/core/resolvers.ts:14-17` awaits both:

```ts
const user = await registry.resolveSelect( store ).getCurrentUser();
const userResponse = await registry.resolveSelect( store ).getUser( user.id );
```

If the failure happened on the `getUser` leg, invalidating `getCurrentUser` does not touch it. `resolveSelect` rejects immediately off the cached failure (`hasResolutionFailed` is still true for `getUser`/`[id]`), Dokan's resolution fails again without issuing any request, and the panel comes straight back. **Try again** becomes a permanent no-op — the same unrecoverable state this PR removes, one screen later.

That leg is a realistic failure point on its own: the `user` entity is configured with `baseURLParams: { context: 'edit' }`, so `getUser` hits `/wp/v2/users/<id>?context=edit` and needs `edit_user`.

The mu-plugin in the test steps hides this, since it blocks the whole `/wp/v2/users` prefix — `getCurrentUser` fails first and `getUser` is never reached. To reproduce, let `/wp/v2/users/me` through and block only `/wp/v2/users/<id>`.

`invalidateResolutionForStoreSelector` covers it without needing the id:

```tsx
// @ts-ignore — metadata actions are absent from the store's own types.
const {
    invalidateResolution: invalidateCoreData,
    invalidateResolutionForStoreSelector: invalidateCoreDataSelector,
} = useDispatch( coreDataStore );

const retryCurrentUser = () => {
    invalidateCoreDataSelector( 'getUser' );
    invalidateCoreData( 'getCurrentUser', [] );
    invalidateResolution( 'getCurrentUser', [] );
};
```

kzamanbd (approved):
Diagnosis is right and the fix is in the right place. Verified the chain: `resolveSelect` rejects on a failed resolution (`hasFinishedResolution` counts `error` as finished), so Dokan's `getCurrentUser` genuinely lands in `error` and the old `status !== 'finished'` gate could never clear. Hooks sit before the early returns, the `InternalError` usage matches `src/dashboard/product-editor/App.tsx:202`, and importing `@src/layout/500` directly rather than through the `@src/components` barrel is the right call for bundle size. ESLint and `tsc --noEmit` are clean on this branch locally. `src/dashboard/index.tsx:34` is also the only `getResolutionState` gate in `src/`, so there is no sibling instance of this bug left behind.

Approving. One thing worth fixing before merge (inline below): the retry only clears one of the two upstream resolutions, so there is a case where **Try again** is a permanent no-op.

Two smaller, optional notes:

- `loading.error` carries the real `WP_Error` (code + status) and is currently dropped. Nothing reaches the console, so support cannot tell "security plugin blocks REST" from a network blip — which is the distinction the ticket is about. A `console.error( 'Dokan: current user resolution failed', loading.error )` would be enough.
- A Playwright case that dispatches `failResolution( 'getCurrentUser', [] )` on the dashboard and asserts the panel plus the button would lock the regression. Low priority.

PR #3396, opened 2026-09-01
fix(admin): stop the legacy dashboard preference expiring after 15 days

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

Choosing the **legacy** admin dashboard wrote a transient with a 15-day TTL, and reading it never refreshed that TTL:

```php
protected int $transient_expiration = 15 * DAY_IN_SECONDS;   // :30
set_transient( $filtered_legacy_key, $new_legacy_state, $this->transient_expiration );   // :129
```

Fifteen days after an admin last clicked switch, their choice silently reverted and the new React dashboard came back with no warning. This is a **preference the marketplace owner set, not a cache entry**, so it should not expire at all.

The TTL is dropped and the now-unused `$transient_expiration` property goes with it.

Storage stays a transient and the key names are unchanged, so **Dokan Pro's own reads keep working untouched** (`dokan-pro/includes/Admin/ToolsActions.php:76`, `dokan-pro/includes/Admin/Notices/WhatsNew.php:40`).

### Closes

* Closes getdokan/plugin-internal-tasks#2345

1. WP Admin → Dokan → Dashboard → switch to the **legacy** dashboard.
2. Confirm the Dokan → Dashboard menu item points at `admin.php?page=dokan`.
3. Simulate the TTL elapsing — `wp transient delete dokan_legacy_dashboard_page` is the end state the old 15-day expiry produced.

**Before:** after 15 days the preference is gone and the menu silently flips back to `page=dokan-dashboard`.
**After:** the preference has no expiry and persists until the admin switches back.

Round-trip verified through the real key helper:

```
key=dokan_legacy_dashboard_page  read_back=true  after_delete=false
```

phpcs clean on the changed file.

### Changelog entry

**Fix — the legacy admin dashboard preference no longer resets itself**

Previously, choosing the legacy Dokan admin dashboard was stored with a 15-day expiry, so a fortnight later the new dashboard silently came back without the admin changing anything. The preference now persists until it is explicitly changed.

### Notes for the reviewer

Known remaining gap, deliberately **not** addressed here: on a site with a persistent object cache the transient never reaches the options table, so any cache flush still drops the preference — verified with `wp cache flush`, after which the Dokan → Dashboard menu item flips back from `page=dokan` to `page=dokan-dashboard`. Closing that means moving the preference to an option and migrating **seven read sites in Lite plus two in Pro**, which is a larger and riskier change than this bug warrants on its own. Documented on getdokan/plugin-internal-tasks#2345.

### DIFF
--- a/includes/Admin/Dashboard/LegacySwitcher.php
+++ b/includes/Admin/Dashboard/LegacySwitcher.php
@@ -21,15 +21,6 @@ class LegacySwitcher implements Hookable {
     public const PRODUCT_EDITOR_SWITCH_ACTION = 'switch_product_editor';
 
     /**
-     * Default transient expiration time in seconds (15 days)
-     *
-     * @since 4.1.3
-     *
-     * @var int
-     */
-    protected int $transient_expiration = 15 * DAY_IN_SECONDS;
-
-    /**
      * Register hooks for the LegacySwitcher
      *
      * @since 4.1.3
@@ -126,7 +117,8 @@ class LegacySwitcher implements Hookable {
         if ( $current_is_legacy ) {
             delete_transient( $filtered_legacy_key );
         } else {
-            set_transient( $filtered_legacy_key, $new_legacy_state, $this->transient_expiration );
+            // Stored without an expiry: this is a preference the marketplace owner set, not a cache entry.
+            set_transient( $filtered_legacy_key, $new_legacy_state );
         }
 
         // Redirect to the new admin page, if needed.


MdAsifHossainNadim:
## Closing — not needed, the 15-day expiry is by design

This PR removed the 15-day TTL on the legacy admin dashboard preference, on the reading that a user preference should not expire.

That reading was wrong: the expiry is a **deliberate product decision**, so the current behaviour is correct and this change should not ship. Closing unmerged.

Tracking issue getdokan/plugin-internal-tasks#2345 is closed as won't-fix, and the parent getdokan/plugin-internal-tasks#2342 has been updated — item 4 is now marked *not a defect — by design* and removed from the priority list.

The branch `fix/admin-dashboard-preference-persistence` is left in place; happy to delete it if you'd rather keep the remote tidy.

The other three PRs from that investigation are unaffected and still stand on their own:

* getdokan/dokan#3394 — vendor's first gallery image moved into the featured slot on save
* getdokan/dokan#3395 — vendor dashboard hangs on an infinite spinner
* getdokan/dokan#3397 — "Register as a Vendor" not enforced server-side

PR #3179, opened 2026-05-05
enhance: Add controllable "Register as a Vendor" option in registration form.

MdAsifHossainNadim:
### All Submissions:

* [x] My code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)
* [x] My code satisfies feature requirements
* [x] My code is tested
* [x] My code passes the PHPCS tests
* [ ] My code has proper inline documentation
* [ ] I've included related pull request(s) (optional)
* [ ] I've included developer documentation (optional)
* [x] I've added proper labels to this pull request

### Related Pull Request(s)

* Full PR Link

### Closes 

* Closes https://github.com/getdokan/dokan-pro/issues/5604

### How to test the changes in this Pull Request:

* Steps or issue link

### Changelog entry

*Title*

Detailed Description of the pull request. What was previous behaviour
and what will be changed in this PR.

### Before Changes

Describe the issue before changes with screenshots(s).

### After Changes

Describe the issue after changes with screenshot(s).

### Feature Video (optional)

Link of detailed video if this PR is for a feature.

### PR Self Review Checklist:

* Code is not following code style guidelines
* Bad naming: make sure you would understand your code if you read it a few months from now.
* KISS: Keep it simple, Sweetie (not stupid!).
* DRY: Don't Repeat Yourself.
* Code that is not readable: too many nested 'if's are a bad sign.
* Performance issues
* Complicated constructions that need refactoring or comments: code should almost always be self-explanatory.
* Grammar errors.

### FOR PR REVIEWER ONLY:

> As a reviewer, your feedback should be focused on the idea, not the person. Seek to understand, be respectful, and focus on constructive dialog.

> As a contributor, your responsibility is to learn from suggestions and iterate your pull request should it be needed based on feedback. Seek to collaborate and produce the best possible contribution to the greater whole.

* [ ] Correct — Does the change do what it’s supposed to? ie: code 100% fulfilling the requirements?
* [ ] Secure — Would a nefarious party find some way to exploit this change? ie: everything is sanitized/escaped appropriately for any SQL or XSS injection possibilities?
* [ ] Readable — Will your future self be able to understand this change months down the road?
* [ ] Elegant — Does the change fit aesthetically within the overall style and architecture?

## Summary by CodeRabbit

* **New Features**
  * Added admin setting to control whether "Register as Vendor" option appears on the signup form
  * Registration form now conditionally displays vendor role selection based on admin preferences

* **Bug Fixes**
[...]

PR #3397, opened 2026-09-01
fix(registration): enforce the "Register as a Vendor" setting on the server

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

Turning **off** Appearance → *Show "Register as a Vendor" in Sign Up Page* only hid the role toggle in `templates/global/seller-registration-form.php:87-99`. `Registration::validate_registration()` never consulted the setting:

```php
$allowed_roles = apply_filters( 'dokan_register_user_role', [ 'customer', 'seller' ] );
```

`seller` was unconditionally allowed, so a crafted POST to the **My Account** sign-up form carrying `role=seller` still registered a vendor. Both places that build the allowed-role list now share `get_allowed_registration_roles()`, which drops `seller` while the toggle is off. The `dokan_register_user_role` filter still runs **last**, so an integration that opens vendor sign-ups on its own terms can put the role back.

No privilege escalation either way — `seller` is the role the form offers when the toggle is on — so this closes a settings-enforcement gap rather than a vulnerability.

#### Scope: which form this setting governs

**The toggle governs the WooCommerce My Account sign-up form only.** That is what its own description says (`includes/Admin/Settings.php:855`: *"Adds the 'I am a customer / I am a vendor' role toggle to the WooCommerce My Account sign-up form"*), and Dokan's dedicated vendor sign-up pages — `[dokan-vendor-registration]` and the **Vendor Onboarding** page published on every fresh install (`includes/Install/Installer.php:250-255`) — are **deliberately unaffected**. Those pages exist to sign vendors up; a setting naming the My Account form must not silently close them.

Dokan's own vendor forms are told apart from the My Account form by a marker field they stamp. A consequence worth stating plainly: because logged-out nonces derive from uid 0, that marker can be lifted with one `GET /vendor-onboarding/` and replayed against `/my-account/`, which then accepts `role=seller` with the toggle off. **Verified — it does.** The outcome is identical to registering on the vendor onboarding page itself, which is open by design, so this is a scope boundary rather than a bypass. A marketplace that wants vendor sign-ups fully closed must also unpublish or restrict its vendor registration and onboarding pages.

### Closes

* Closes getdokan/plugin-internal-tasks#2346

**Setup:** Dokan → Settings → Appearance → **Show "Register as a Vendor" in Sign Up Page** → off, save.

Confirm the visible half already works — as a logged-out visitor, `/my-account/` renders only `name="role" value="customer"`, with no vendor radio.

**Reproduce.** The registration form only renders for logged-out visitors, so drive it as an anonymous visitor. Grab the nonce from the page, then POST with `role=seller` injected:

```bash
curl -s -c c.jar http://your-site.test/my-account/ -o reg.html
NONCE=$(grep -o 'name="woocommerce-register-nonce" value="[^"]*"' reg.html | sed 's/.*value="//;s/"//')

curl -s -b c.jar -c c.jar http://your-site.test/my-account/ \
  --data-urlencode "email=an-existing-account@example.com" \
  --data-urlencode "password=Str0ngTestPass!123" \
  --data-urlencode "role=seller" \
  --data-urlencode "fname=Test" --data-urlencode "lname=Vendor" \
  --data-urlencode "phone=0123456789" \
  --data-urlencode "shopname=Crafted Shop" --data-urlencode "shopurl=crafted-shop-probe" \
  --data-urlencode "woocommerce-register-nonce=$NONCE" \
  --data-urlencode "register=Register" | grep -o 'woocommerce-error.*</ul>'
```

Use an **already-registered email** so no account is created in either run. That also makes the result unambiguous — reaching WooCommerce's duplicate-email check at all is only possible if Dokan let the role through. Do **not** send the vendor-form marker here; lifting it from `/vendor-onboarding/` first is the documented scope boundary above, not the case under test.

| Branch | Response | Meaning |
|---|---|---|
| `develop` | `Error: An account is already registered with …` | Dokan **accepted** `seller`, handed off to account creation |
| this PR | `Error: Cheating, eh?` | Dokan **rejected** the role first |

With a fresh email, `develop` creates a vendor through a My Account form that had closed vendor sign-ups.

### Testing scope

**Verified — regression matrix, re-run on the current head against both sign-up pages:**

| Setting | Page | Posted role | Result |
|---|---|---|---|
| **off** | My Account | `seller` (crafted) | `Cheating, eh?` — blocked |
| **off** | My Account | `customer` | reaches account creation |
| **off** | Vendor onboarding | `seller` (with marker) | reaches account creation — page stays open |
| **off** | Vendor onboarding | `seller`, marker stripped | `Cheating, eh?` — blocked |
| **on** | My Account | `seller` | reaches account creation |
| **on** | My Account | `customer` | reaches account creation |
| **on** | Vendor onboarding | `seller` | reaches account creation |

Blocks exactly the cases it should, nothing else. No account was created in any run — every probe used an existing email.

**Verified — theme template overrides.** An override of `account/vendor-registration.php` that drops `do_action( 'dokan_vendor_reg_form_start' )` never prints the marker, which used to turn that site's own vendor page into `Cheating, eh?` the moment the toggle went off, with nothing to explain why. Dokan has no outdated-template notice to hang a version bump on (`dokan_get_template_part()` performs no version check at all), so the marker is now also stamped over the rendered form through the existing `dokan_vendor_reg_form` filter, which an override cannot drop. Reproduced both sides by removing the action: before the safety net the sign-up is refused; after it, it succeeds. Confirmed the marker lands in the form carrying the `role` field — not the onboarding page's login form — and is never stamped twice.

**Verified — `dokan_register_nonce_check`.** `validate_nonce()` lets a site bypass the register-nonce check for headless and custom sign-up forms. The new check ignored it, so such a site lost vendor registration entirely once the toggle was off. It is honoured now: with the filter returning false, toggle off and no marker, `role=seller` is accepted.

**Verified — filter escape hatch.** Toggle off, plus a filter adding `seller` back, returns `[customer, seller]`.

**Verified — PHPUnit.** `tests/php/src/RegistrationTest.php`, 11 tests / 11 assertions, green locally against WP 7.1 + WooCommerce with HPOS.

**Verified — adjacent paths:**

* **Checkout / WP-CLI / REST** — `validate_registration()` early-returns for `is_checkout()`, `WP_CLI` and `REST_REQUEST`, so the gate never runs there. `set_new_vendor_names()` still normalises a crafted `role=seller` down to `customer` while the toggle is off, closing the same hole on the checkout path as a side benefit. Worth a QA eye if any site legitimately creates vendors at checkout with the toggle off — unusual, but it is a behaviour change on that path.
* **Pro — Germanized module.** `dokan-pro/modules/germanized/includes/CustomFields/Registration.php:216` keeps its **own copy** of `apply_filters( 'dokan_register_user_role', [ 'customer', 'seller' ] )`. It runs on `woocommerce_new_customer_data` at priority 99, **downstream of the validation gate**, so with the toggle off the request is already rejected before it runs — the fix holds with the module active. That duplicate should be aligned with the shared helper in a Pro follow-up; it does not bypass this fix.

**Not covered:**

* The form was driven over **HTTP rather than clicked in a browser** — the registration form only renders logged out, and the available browser session was a logged-in admin. The HTTP path is what a crafted request actually uses, so the reproduction is faithful, but there is no screenshot of the submission. The admin toggle itself was exercised in the browser.
* Pro's **customer → vendor account-migration** flow (`dokan_after_seller_migration_fields`) is a separate form and was not exercised. Whether migration should also respect this setting is a product question, not part of this fix.
* No multisite run.

**Static checks:** phpcs on `includes/Registration.php` reports the same **22 pre-existing** nonce false positives as `develop`; none added. `tests/php/src/RegistrationTest.php` is clean. No new `$_POST` handling introduced.

### Changelog entry

**Fix — "Show Register as a Vendor" is now enforced on the server**

Previously, turning the setting off only hid the vendor option on the WooCommerce My Account sign-up form; a request that asked for the vendor role anyway was still accepted, so a marketplace that had closed vendor sign-ups there could still gain vendors. The setting is now enforced when the registration is processed. Dokan's own vendor registration and Vendor Onboarding pages are unaffected — they exist to sign vendors up, and the setting has never governed them.

**Upgrade note for customised themes.** Sites with their own copy of `dokan/account/vendor-registration.php` keep working: Dokan now adds the field that identifies its vendor sign-up form to the rendered output as well, so an override that does not fire `dokan_vendor_reg_form_start` no longer breaks vendor registration while the setting is off.

### Before Changes

Setting off, crafted `role=seller` POST to **My Account** → passes Dokan validation, reaches WooCommerce account creation:

```
Error: An account is already registered with … . Please log in or use a different email address.
```

### After Changes

Same request, same setting → rejected by Dokan before account creation:

```
Error: Cheating, eh?
```

### Notes for the reviewer

The setting itself shipped from getdokan/dokan-pro#5604 via #3179. This is a follow-up gap on that work, not a regression.

### DIFF
--- a/includes/Registration.php
+++ b/includes/Registration.php
@@ -11,22 +11,10 @@ use WP_Error;
  */
 class Registration {
 
-    /**
-     * Nonce action marking a request as coming from Dokan's own vendor sign-up form.
-     *
-     * @since DOKAN_SINCE
-     *
-     * @var string
-     */
+    /** Nonce action marking a request as coming from Dokan's own vendor sign-up form. @since DOKAN_SINCE */
     const VENDOR_FORM_NONCE_ACTION = 'dokan_vendor_registration_form';
 
-    /**
-     * Field carrying the vendor sign-up form marker.
-     *
-     * @since DOKAN_SINCE
-     *
-     * @var string
-     */
+    /** Field carrying the vendor sign-up form marker. @since DOKAN_SINCE */
     const VENDOR_FORM_NONCE_FIELD = 'dokan_vendor_registration_form_nonce';
 
     public function __construct() {
@@ -45,8 +33,8 @@ class Registration {
     /**
      * Print the marker identifying Dokan's dedicated vendor sign-up form.
      *
-     * Hooked to an action only that form fires, so the marker travels with the submission even when
-     * a theme overrides the template.
+     * Hooked to an action only that template fires. A theme override keeping the action keeps the
+     * marker; one that drops it falls back to the appearance setting like any other form.
      *
      * @since DOKAN_SINCE
      *
@@ -59,41 +47,22 @@ class Registration {
     /**
      * Roles a visitor may register as through the request being processed.
      *
-     * `seller` drops out when the WooCommerce sign-up form hides the "I am a vendor" toggle, because
-     * a role that form never offered has to be rejected server side too — the template alone only
-     * hides it. Dokan's own vendor sign-up pages are exempt, see is_vendor_role_on_offer(). The
-     * filter still runs last, so an integration that opens vendor sign-ups by its own rules can put
-     * `seller` back.
+     * The appearance setting only hides the vendor option in the WooCommerce form's template, so a
+     * role that form never offered has to be refused server side too.
      *
      * @since DOKAN_SINCE
      *
      * @return array
      */
     public function get_allowed_registration_roles() {
-        $roles = [ 'customer', 'seller' ];
+        $roles = [ 'customer' ];
 
-        if ( ! $this->is_vendor_role_on_offer() ) {
-            $roles = array_values( array_diff( $roles, [ 'seller' ] ) );
+        // Dokan's dedicated vendor pages exist to sign vendors up, so a setting naming the WooCommerce form must not close them.
+        if ( $this->is_vendor_form_request() || dokan_is_vendor_registration_offered() ) {
+            $roles[] = 'seller';
         }
 
         return apply_filters( 'dokan_register_user_role', $roles );
-    }
-
-    /**
-     * Whether the vendor role is on offer for the request being processed.
-     *
-     * @since DOKAN_SINCE
-     *
-     * @return bool
-     */
-    protected function is_vendor_role_on_offer() {
-        // Dokan's dedicated vendor pages exist to sign vendors up, so a setting that names the
-        // WooCommerce sign-up form must not close them.
-        if ( $this->is_vendor_form_request() ) {
-            return true;
-        }
-
-        return 'off' !== dokan_get_option( 'show_register_as_vendor', 'dokan_appearance', 'on' );
     }
 
     /**
@@ -107,7 +76,7 @@ class Registration {
         // Only that form prints this marker, so a request that never rendered it cannot claim to be one.
         $nonce = isset( $_POST[ self::VENDOR_FORM_NONCE_FIELD ] ) ? sanitize_key( wp_unslash( $_POST[ self::VENDOR_FORM_NONCE_FIELD ] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing -- this is the nonce check.
 
-        return ! empty( $nonce ) && wp_verify_nonce( $nonce, self::VENDOR_FORM_NONCE_ACTION );
+        return (bool) wp_verify_nonce( $nonce, self::VENDOR_FORM_NONCE_ACTION );
     }
 
     /**
--- a/includes/Registration.php
+++ b/includes/Registration.php
@@ -40,20 +40,58 @@ class Registration {
 
         // stamp the dedicated vendor form so the two sign-up forms can be told apart on submit
         add_action( 'dokan_vendor_reg_form_start', [ $this, 'render_vendor_form_marker' ] );
+        add_filter( 'dokan_vendor_reg_form', [ $this, 'inject_vendor_form_marker' ] );
+    }
+
+    /**
+     * Build the marker identifying Dokan's dedicated vendor sign-up form.
+     *
+     * @since DOKAN_SINCE
+     *
+     * @return string
+     */
+    protected function get_vendor_form_marker() {
+        return wp_nonce_field( self::VENDOR_FORM_NONCE_ACTION, self::VENDOR_FORM_NONCE_FIELD, false, false );
     }
 
     /**
      * Print the marker identifying Dokan's dedicated vendor sign-up form.
-     *
-     * Hooked to an action only that template fires. A theme override keeping the action keeps the
-     * marker; one that drops it falls back to the appearance setting like any other form.
      *
      * @since DOKAN_SINCE
      *
      * @return void
      */
     public function render_vendor_form_marker() {
-        wp_nonce_field( self::VENDOR_FORM_NONCE_ACTION, self::VENDOR_FORM_NONCE_FIELD, false );
+        echo $this->get_vendor_form_marker(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- wp_nonce_field() builds its own escaped markup.
+    }
+
+    /**
+     * Add the marker to a rendered vendor sign-up form that did not print it itself.
+     *
+     * A theme override copied before the marker existed never fires the action, and the toggle would
+     * then close that site's own vendor page with no way to tell why. Stamping the rendered markup
+     * keeps such an override working, whatever it did to the form.
+     *
+     * @since DOKAN_SINCE
+     *
+     * @param string $content Rendered vendor sign-up markup.
+     *
+     * @return string
+     */
+    public function inject_vendor_form_marker( $content ) {
+        if ( ! is_string( $content ) || false !== strpos( $content, self::VENDOR_FORM_NONCE_FIELD ) ) {
+            return $content;
+        }
+
+        $closing_tag = strrpos( $content, '</form>' );
+
+        // Onboarding renders the login form too, so only stamp markup that actually asks for a role.
+        if ( false === $closing_tag || ! preg_match( '/<input[^>]+name=["\']?role["\'\s>]/i', $content ) ) {
+            return $content;
+        }
+
+        // The sign-up form is the last one on both the shortcode and the onboarding layout.
+        return substr_replace( $content, $this->get_vendor_form_marker(), $closing_tag, 0 );
     }
 
     /**
@@ -85,6 +123,11 @@ class Registration {
      * @return bool
      */
     protected function is_vendor_form_request() {
+        // A site that turned the register-nonce check off drives its own sign-up forms; do not strand them.
+        if ( ! apply_filters( 'dokan_register_nonce_check', true ) ) {
+            return true;
+        }
+
         // Only that form prints this marker, so a request that never rendered it cannot claim to be one.
         $nonce = isset( $_POST[ self::VENDOR_FORM_NONCE_FIELD ] ) ? sanitize_key( wp_unslash( $_POST[ self::VENDOR_FORM_NONCE_FIELD ] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing -- this is the nonce check.
 
--- a/includes/Registration.php
+++ b/includes/Registration.php
@@ -22,6 +22,28 @@ class Registration {
     }
 
     /**
+     * Roles a visitor may register as through the WooCommerce sign-up form.
+     *
+     * `seller` drops out when the marketplace hides the "I am a vendor" toggle, because a role the
+     * form never offered has to be rejected server side too — the template alone only hides it.
+     * The filter still runs last, so an integration that opens vendor sign-ups by its own rules can
+     * put `seller` back.
+     *
+     * @since DOKAN_SINCE
+     *
+     * @return array
+     */
+    protected function get_allowed_registration_roles() {
+        $roles = [ 'customer', 'seller' ];
+
+        if ( 'off' === dokan_get_option( 'show_register_as_vendor', 'dokan_appearance', 'on' ) ) {
+            $roles = array_values( array_diff( $roles, [ 'seller' ] ) );
+        }
+
+        return apply_filters( 'dokan_register_user_role', $roles );
+    }
+
+    /**
      * Validate vendor registration
      *
      * @param \WP_Error $error
@@ -41,7 +63,7 @@ class Registration {
             return new WP_Error( 'nonce_verification_failed', __( 'Nonce verification failed', 'dokan-lite' ) );
         }
 
-        $allowed_roles = apply_filters( 'dokan_register_user_role', [ 'customer', 'seller' ] );
+        $allowed_roles = $this->get_allowed_registration_roles();
 
         // is the role name allowed or user is trying to manipulate?
         if ( empty( $_POST['role'] ) || ( ! in_array( $_POST['role'], $allowed_roles, true ) ) ) {
@@ -89,7 +111,7 @@ class Registration {
             return $data;
         }
 
-        $allowed_roles = apply_filters( 'dokan_register_user_role', [ 'customer', 'seller' ] );
+        $allowed_roles = $this->get_allowed_registration_roles();
         $role          = ( isset( $_POST['role'] ) && in_array( $_POST['role'], $allowed_roles, true ) ) ? sanitize_text_field( wp_unslash( $_POST['role'] ) ) : 'customer';
 
         $data['role'] = $role;
--- a/includes/Registration.php
+++ b/includes/Registration.php
@@ -11,6 +11,24 @@ use WP_Error;
  */
 class Registration {
 
+    /**
+     * Nonce action marking a request as coming from Dokan's own vendor sign-up form.
+     *
+     * @since DOKAN_SINCE
+     *
+     * @var string
+     */
+    const VENDOR_FORM_NONCE_ACTION = 'dokan_vendor_registration_form';
+
+    /**
+     * Field carrying the vendor sign-up form marker.
+     *
+     * @since DOKAN_SINCE
+     *
+     * @var string
+     */
+    const VENDOR_FORM_NONCE_FIELD = 'dokan_vendor_registration_form_nonce';
+
     public function __construct() {
         // validate registration
         add_filter( 'woocommerce_process_registration_errors', [ $this, 'validate_registration' ] );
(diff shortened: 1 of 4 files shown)


MdAsifHossainNadim:
## Verdict: Approve ✅

### The bug

Turning off Appearance → *Show "Register as a Vendor" in Sign Up Page* only hid the role radio in the My Account template. `Registration::validate_registration()` never consulted the setting, so a POST carrying `role=seller` still registered a vendor.

### How it's fixed

Both places that build the allowed-role list now share `get_allowed_registration_roles()`, which withholds `seller` while the toggle is off.

The subtlety is that Dokan ships **two** sign-up forms that post to the same WooCommerce handler with the same nonce:

| Form | Template | Role posted |
|---|---|---|
| WooCommerce My Account | `templates/global/seller-registration-form.php` | radio, or hidden `customer` when the toggle is off |
| Vendor Onboarding / `[dokan-vendor-registration]` | `templates/account/vendor-registration.php` | hidden `seller`, always |

A blanket check closes the second form too — and the page carrying it is created on **every** install (`Installer.php:250`, the 5.0.0 upgrade, and Dokan → Tools). So the gate is scoped to the form the setting actually names: `Registration` prints a nonce marker on `dokan_vendor_reg_form_start`, an action only the dedicated template fires, and a request carrying a valid marker keeps `seller`.

```php
$roles = [ 'customer' ];

// Dokan's dedicated vendor pages exist to sign vendors up, so a setting naming the WooCommerce form must not close them.
if ( $this->is_vendor_form_request() || 'off' !== dokan_get_option( 'show_register_as_vendor', 'dokan_appearance', 'on' ) ) {
    $roles[] = 'seller';
}

return apply_filters( 'dokan_register_user_role', $roles );
```

### Why it's correct

* **It matches what the setting promises.** The label and description both scope it to the WooCommerce sign-up form. Enforcing it there and nowhere else changes no working configuration — a marketplace running vendor sign-up on the dedicated page keeps working with the toggle off, which is exactly the configuration that toggle invites.
* **The marker is a nonce, not a magic string.** Stripping it or forging it does not buy the exemption; both cases are covered by tests and were confirmed live.
* **It survives theme overrides.** The marker is hooked to the action rather than written into the template, so an override that keeps `do_action( 'dokan_vendor_reg_form_start' )` keeps the marker. An override that drops the action degrades to the appearance setting like any other form — no fatal, no silent grant.
* **The escape hatch is intact.** `dokan_register_user_role` still runs **last**, so an integration can reopen vendor sign-ups on its own terms.
* **Validation and assignment cannot disagree.** `validate_registration()` and `set_new_vendor_names()` read the same accessor within the same request.

### Testing scope

**Live, real submissions on WooCommerce 11.0.1**, fresh emails so accounts genuinely get created (both probe accounts deleted afterwards):

| Toggle | Path | `develop` | this branch |
|---|---|---|---|
| off | My Account, crafted `role=seller` | accepted 🐛 | `Cheating, eh?` — **no account created** |
| off | My Account, `role=customer` | registers | registers |
| off | Vendor Onboarding, real submit | registers | **registers a vendor** (`roles=seller`) |
| off | Vendor Onboarding, marker stripped | — | `Cheating, eh?` |
| on | every path above | registers | registers |

**Browser.** With the toggle off, the anonymous render of Vendor Onboarding shows the complete form and carries the marker; My Account shows a plain Register box with no vendor option and no marker. Marker count per page: `/my-account/` 0, `/vendor-onboarding/` 1, `/onboarding-new/` 1.

**Unit.** 7 cases in `tests/php/src/RegistrationTest.php`, all green in 0.27s — toggle on/off/never-saved via data provider, the marker exemption, a forged marker rejected, the filter escape hatch, and the marker asserted by **rendering `account/vendor-registration`** rather than by firing the action directly, so the test fails if the template ever stops firing it.

**Static.** PHPCS reports the same 22 pre-existing findings as `develop`, none added; the test file is clean.

**Not covered.** No multisite run. The customer → vendor migration flow (`includes/Frontend/MyAccount/BecomeAVendor.php` — Lite, not Pro) is a separate surface that never passes through `validate_registration()` and is deliberately untouched.

### Follow-ups, none blocking

1. The setting check is now stated twice — here and at `templates/global/seller-registration-form.php:87`. Worth one shared predicate, but adding a global to `includes/functions.php` does not belong in a bugfix PR.
2. Pro's `modules/germanized/includes/CustomFields/Registration.php:216` keeps its own hard-coded copy of the role list and overwrites `$data['role']` at priority 99, which undoes the downgrade on the checkout path. `get_allowed_registration_roles()` is public so Pro can retire that duplicate.
3. `BecomeAVendor` is ungated by this setting — a customer can still self-promote in one click with the toggle off.
4. Longer term, the honest shape is a distinct "vendor registration open" policy setting, leaving `show_register_as_vendor` purely about the My Account radio. Then no marker is needed at all.

> **The description above is stale** — it predates the last three commits. It still describes the original global behaviour, says no automated test was added, and attributes the migration flow to Pro. Worth refreshing before merge.

includes/Registration.php
```diff
@@ -19,6 +37,58 @@ public function __construct() {
         // after registration
         add_filter( 'woocommerce_new_customer_data', [ $this, 'set_new_vendor_names' ] );
         add_action( 'woocommerce_created_customer', [ $this, 'save_vendor_info' ], 10, 2 );
+
+        // stamp the dedicated vendor form so the two sign-up forms can be told apart on submit
+        add_action( 'dokan_vendor_reg_form_start', [ $this, 'render_vendor_form_marker' ] );
```
kzamanbd, line 42: **Blocker — theme template overrides break silently.**

Any site carrying `yourtheme/dokan/account/vendor-registration.php` copied before this change will not print the marker. With the toggle off, that site's vendor sign-up page starts returning `Cheating, eh?`, with nothing to indicate why.

The docblock below acknowledges the fallback, but "falls back to the appearance setting" means "stops working" for exactly the sites that customised the form, and the changelog entry does not mention it at all.

Please bump the template version so `dokan_get_template_part()`'s outdated-template notice fires, and add an upgrade note to the changelog naming the template. If you would rather not depend on the override keeping the action, stamping the marker from a filter over the rendered form output would survive a stale copy.

includes/Registration.php
```diff
@@ -19,6 +37,58 @@ public function __construct() {
...
+     *
+     * @since DOKAN_SINCE
+     *
+     * @return bool
+     */
+    protected function is_vendor_form_request() {
+        // Only that form prints this marker, so a request that never rendered it cannot claim to be one.
+        $nonce = isset( $_POST[ self::VENDOR_FORM_NONCE_FIELD ] ) ? sanitize_key( wp_unslash( $_POST[ self::VENDOR_FORM_NONCE_FIELD ] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing -- this is the nonce check.
+
+        return (bool) wp_verify_nonce( $nonce, self::VENDOR_FORM_NONCE_ACTION );
```
kzamanbd, line 134: **Blocker — `dokan_register_nonce_check` is not honoured here.**

`validate_nonce()` (line 300) lets a site bypass the WooCommerce register-nonce check through `apply_filters( 'dokan_register_nonce_check', true )`, which exists for headless and custom sign-up forms. This new check has no equivalent escape, so a site using that documented bypass with the toggle off loses vendor registration entirely — the bypass no longer covers the whole path it used to.

Either respect the same filter:

```php
if ( ! apply_filters( 'dokan_register_nonce_check', true ) ) {
    return true;
}
```

or state explicitly in the `get_allowed_registration_roles()` docblock that `dokan_register_user_role` is the only supported hatch for those setups.

kzamanbd (changes requested):
### Review: request changes

The core fix is right — routing both call sites through a single `get_allowed_registration_roles()` closes the real gap, `@since DOKAN_SINCE` is used correctly, and strict `in_array()` is preserved. PHPCS shows **22** errors on `includes/Registration.php` on both `develop` and this branch, so nothing new. Three blockers below.

---

#### 1. The description claims more than the code delivers

`includes/Install/Installer.php:250-255` publishes a **public "Vendor Onboarding" page** (`[dokan-vendor-onboarding-registration]`) on every fresh install. That shortcode renders `account/vendor-registration`, which now prints the marker. Nonces for logged-out visitors derive from uid 0 and an empty session token, so every anonymous visitor within the tick gets the same value.

The consequence: on a **default install**, the crafted POST in your "Before Changes" section still succeeds. It needs one extra `GET /vendor-onboarding/` to lift the marker first, and then `/my-account/` accepts `role=seller` with the toggle off.

The behaviour itself is defensible — the setting's own description (`includes/Admin/Settings.php:855`) already scopes it to "the WooCommerce My Account sign-up form", and the outcome is identical to registering on the vendor page, which is intentionally open. What is not defensible is the framing. The body reads as a general claim — *"a marketplace owner who believed they had closed vendor sign-ups had not"* — and the After Changes table presents `Cheating, eh?` as the outcome for that request. That holds only on a site where no page renders the vendor form.

Please state the scope in both the description and the changelog entry: the toggle governs the My Account sign-up form; Dokan's dedicated vendor registration and onboarding pages are deliberately unaffected. Also drop "No automated test added" from **Testing scope** — tests exist now — and re-run the regression matrix, which predates the marker commits and covers neither the dedicated form nor the onboarding page.

#### 2. Theme template overrides break silently

See the inline comment on the constructor.

#### 3. `dokan_register_nonce_check` is not honoured by the new check

See the inline comment on `is_vendor_form_request()`.

---

**Not verified:** PHPUnit did not run here — no Docker daemon available. The new test class needs a green run before merge.

**Labels:** please add `Type: Bug`, `focus: vendor management`, `focus: security`. Moving `Needs: Dev Review` to `Needs: Author Reply`.

includes/Registration.php
MdAsifHossainNadim: Valid — reproduced it. Removing `dokan_vendor_reg_form_start` and loading `/vendor-onboarding/` with the toggle off, the site's own vendor page answers `Cheating, eh?`, exactly as you describe. Fixed in 872a620c6.

Two corrections to the prescribed remedy, though.

**The template-version route does not exist.** `dokan_get_template_part()` performs no version check — it resolves a path through `locate_template()`, applies `dokan_get_template_part`, and `include`s it. There is no `get_file_data()` call, no `@version` read, and no outdated-template notice anywhere in Lite. Bumping a version header would have been a silent no-op.

**The blast radius is narrower than the comment implies.** `do_action( 'dokan_vendor_reg_form_start' )` and `id="dokan-vendor-register"` both arrived in the same commit, b9da8a827 (2018-03-28). An override that lacks the action therefore predates March 2018 and is already broken against the current form in other ways. The realistic case is a deliberately rewritten form, not a merely stale copy — still worth closing, just not the wide exposure it first reads as.

So I took your **second** suggestion, which is the right one: the marker is now also stamped over the rendered output through the existing `dokan_vendor_reg_form` filter, which both sign-up shortcodes already apply and no override can drop. It:

- skips markup that already carries the marker, so the normal path is untouched and never double-stamped;
- requires a `role` input before stamping, so the onboarding page's **login** form is left alone;
- injects before the last `</form>`, which is the sign-up form on both the shortcode and the onboarding layout.

Verified over HTTP with the action removed: before the safety net the marker is absent and the sign-up is refused; after it, the marker is present once, lands in the form carrying `role` (form index 2 of 4, same index as the role field), and the sign-up succeeds.

Covered by `test_a_template_override_that_dropped_the_action_still_gets_the_marker`, `test_the_marker_is_not_stamped_twice` and `test_a_login_only_onboarding_page_is_left_alone`.

The changelog now carries an upgrade note naming the template.

includes/Registration.php
MdAsifHossainNadim: Valid. Fixed in 872a620c6 by respecting the same filter, as suggested:

```php
protected function is_vendor_form_request() {
    // A site that turned the register-nonce check off drives its own sign-up forms; do not strand them.
    if ( ! apply_filters( 'dokan_register_nonce_check', true ) ) {
        return true;
    }
    ...
}
```

Agreed on the reasoning: a site that switched the register-nonce check off has taken over its own sign-up forms, so Dokan cannot expect its marker on those requests, and `dokan_register_user_role` remains the hook for actually controlling the role there.

Verified over HTTP — filter returning false, toggle off, no marker on the request: `role=seller` now reaches account creation, where it was refused with `Cheating, eh?` before. The same request with the filter left alone is still blocked, so the bypass is the only thing that changed.

Covered by `test_disabling_the_register_nonce_check_does_not_strand_custom_forms`.

MdAsifHossainNadim:
Thanks — all three were valid. Addressed in 872a620c6 plus a description rewrite; the two inline threads are answered and resolved.

### 1. The description claimed more than the code delivers

Confirmed, and reproduced rather than taken on trust: `Installer.php:250-255` does publish the **Vendor Onboarding** page on every fresh install, and lifting the marker with a single `GET /vendor-onboarding/` then replaying it against `/my-account/` with `role=seller` and the toggle off **is accepted**. The old *Before / After* table was therefore not true on a default install.

The description and changelog now state the scope outright: the toggle governs the **WooCommerce My Account sign-up form**, Dokan's dedicated vendor registration and Vendor Onboarding pages are deliberately unaffected, and — stated plainly rather than left implicit — a marketplace that wants vendor sign-ups fully closed has to unpublish or restrict those pages too. The reproduction steps now say not to send the marker, since that is the scope boundary and not the case under test.

"No automated test added" is gone, and the regression matrix has been re-run on the current head across **both** sign-up pages and both toggle states (7 rows, all as expected). No account was created in any run — every probe reused an existing email.

### 2 & 3

Answered inline. Short version: the theme-override breakage is real and is now closed through the `dokan_vendor_reg_form` filter — your second suggestion — because the first one cannot work: `dokan_get_template_part()` performs no version check and Lite has no outdated-template notice to trigger. Worth noting the exposure is also narrower than it first reads, since the action and the form id landed in the same 2018 commit. `dokan_register_nonce_check` is honoured now.

### The two open items from your review

**PHPUnit** — ran it: `tests/php/src/RegistrationTest.php`, **11 tests / 11 assertions, green** (WP 7.1, WooCommerce with HPOS). Four cases are new, covering the override safety net, the no-double-stamp guarantee, the login-only onboarding page, and the nonce-check bypass.

**PHPCS** — your count was right and the PR body was wrong: **22**, not 23, and identical on `develop` and this branch. Corrected in the body.

**Labels** — added `Type: Bug`; `focus: security` and `focus: vendor management` were already on. Moving `Needs: Author Reply` back to `Needs: Dev Review`.

One thing I did **not** do, since it looks like a product call rather than part of this fix: Pro's customer → vendor migration flow is still untouched, and the Germanized module's duplicate role list still wants aligning with the shared helper in a Pro follow-up.

kzamanbd (approved):
### Approved

All three blockers resolved in 872a620c6. Re-checked each against the head.

**Scope** — description and changelog now state the boundary; stale claims and the old matrix are gone.

**Theme overrides** — you were right on the version bump: `dokan_get_template_part()` does no version check, so that suggestion wasn't actionable. Stamping through `dokan_vendor_reg_form` is the better fix. Verified independently: the filter has only two call sites (`VendorRegistration.php:33`, `VendorOnboardingRegistration.php:55`), neither renders the My Account form, so the gated form can't be stamped by accident; the `role` regex rejects `roles` / `user_role` / `role_id` / login-only markup; and the registration form is last in `vendor-onboarding.php`, so `strrpos( '</form>' )` targets it.

**`dokan_register_nonce_check`** — honoured now.

PHPCS: 22 errors on both `develop` and this branch; test file clean. I couldn't run PHPUnit (no Docker here), so QA's run is the one that counts.

Two notes, not blocking: setting `dokan_register_nonce_check` false now disables the toggle everywhere including My Account — wider than the thread implied, though the description says so; and `inject_vendor_form_marker()` matches `role` across the whole string but inserts at the last `</form>`, which only diverges under a reordered onboarding override.

Follow-up rather than another round: `dokan_register_user_role` has no PHPDoc at its new call site and its input contract changed (it can now receive `[ 'customer' ]`) — worth documenting for integrators. `array_values( array_unique( ... ) )` on the result, tests driving `validate_registration()` and `set_new_vendor_names()`, and `@group security` would all help.

Passing to QA.

Issue #3321, opened 2026-07-16
Dokan 5.0.9 admin Vendors list shows zero vendors while REST API returns valid stores


Vomcraft:
### Bug Description

On a test marketplace site, the Dokan admin Vendors screen shows no vendors in any filter: All (0), Approved (0), and Pending (0).

A test vendor exists in WordPress Users, has the internal role slug seller, has selling enabled, and has Dokan store profile fields configured. The issue affects the admin listing only.

### Step To Reproduce

1. Install and activate Dokan Lite 5.0.9 with WooCommerce.
2. Create or use a test-only vendor through Dokan.
3. Confirm the account has role slug seller, selling enabled, and a store name and store URL.
4. Open Dokan > Vendors as an administrator.
5. Reload the screen and repeat the check in a separate browser tab.
6. Request /wp-json/dokan/v1/stores and inspect the response count only.

### Expected Behaviour

The Dokan admin Vendors screen should list the existing test vendor(s), and the All, Approved, and Pending counters should reflect the returned store records.

### Actual Behaviour

The Vendors screen consistently shows All (0), Approved (0), Pending (0), and No data found. The direct Dokan stores REST endpoint returns HTTP 200 and a valid JSON array containing two store records, including the affected test vendor.

### Additional Information

REST and browser observations

- No browser JavaScript console errors were observed on the Vendors screen.
- No 4xx, 5xx, or nonce error was observed for the checked REST endpoint.
- The Dokan REST namespace and /dokan/v1/stores endpoint are available.

Source-code observation (not a conclusion)

In the installed Dokan Lite 5.0.9 admin asset vendors-store.js, the vendor resolver appears to request /dokan/v1/stores. The compiled asset defines a setVendors(...) action/reducer, but this read-only investigation did not identify a clear step that dispatches the REST response into the list state. This may indicate a client-side state-handling or build issue; please verify against the original source and build pipeline.

Separate cron observation

Two failed Action Scheduler entries were observed for dokan_daily_midnight_cron, reporting that the callback is not registered. A direct causal relationship with the zero-vendor list has not been established.

Privacy

This report intentionally excludes site URLs, email addresses, usernames, user IDs, database identifiers, internal paths, full API JSON, and real store details. The two records are referred to only as Test Vendor 1 and Test Vendor 2.

Redacted screenshots can be provided on request. No screenshots are attached here because the original captures require redaction before public sharing.

[...]

PR #3399, opened 2026-09-02
fix(admin): stop stacking the pending-vendor meta query onto every user query

MdAsifHossainNadim:

`UserList::filter_pending_vendors()` hooks `pre_get_users`, which fires for **every** `WP_User_Query` built while `users.php?role=pending_vendor` renders — including the one `dokan_get_pending_vendor_count()` runs through `Vendor\Manager::get_vendors()`, which already builds its own `dokan_enable_selling` group for `status => pending`.

Stacked, WordPress compiles four `LEFT JOIN wp_usermeta`, three of them with no `meta_key` in the `ON` clause, so each user fans out across its own meta rows before the `WHERE` filters anything.

The reporter measured **141,238,050 rows examined to return a single value**, 193–221 seconds per request, on a site with 319 users and a 24k-row `wp_usermeta`.

Only the list table asks for that role — WordPress passes `'role' => $_REQUEST['role']` (`wp-admin/includes/class-wp-users-list-table.php:107`) — while `get_vendors()` scopes with `role__in` and never sets `role`. Gating on that leaves the list query working and the badge query alone.

### Closes

* Closes #

1. On a site with a few hundred sellers and no persistent object cache, open **wp-admin → Users → Pending Vendor**.
2. Watch the slow query log / `SHOW PROCESSLIST`.

**Expected:** no `WP_User_Query` with four `LEFT JOIN wp_usermeta` and three `dokan_enable_selling` clauses. The Pending Vendor list itself, and the pending badge on Dokan → Vendors, are unchanged.

### Test results

Measured on a real admin request with a temporary `pre_user_query` probe, `wp-admin → Users → Pending Vendor`:

```
before (develop)              after (this PR)
joins=3 clauses=2             joins=3 clauses=2   <- list table query, unchanged
joins=4 clauses=3   <-- !!    joins=0 clauses=1   <- badge query, no longer stacked
joins=3 clauses=2
```

The `joins=4 / clauses=3` query is the one from the issue's slow log; it is gone.

Independently, building both shapes and running `EXPLAIN` on a 50-user site:

```
get_vendors group only    joins=0   est. rows examined     1,485
both groups stacked       joins=4   est. rows examined 2,205,225
```

Same combinatorial shape the reporter hit at 319 users.

### Notes

The pending **count** and the pending **list** still disagree about vendors whose `dokan_enable_selling` meta is absent — `get_vendors()` matches only `= 'no'` while this filter and `dokan_get_seller_status_count()` also treat a missing key as pending. That is #3321 and is fixed separately in getdokan/dokan#3400.

### DIFF
--- a/includes/Admin/UserList.php
+++ b/includes/Admin/UserList.php
@@ -71,6 +71,12 @@ class UserList {
             return $query;
         }
 
+        // Only the list table asks for this role; every other user query on the screen -- the pending
+        // count badge among them -- already carries its own criteria and must not be stacked on to.
+        if ( 'pending_vendor' !== $query->get( 'role' ) ) {
+            return $query;
+        }
+
         // Unset the role parameter for fetch all users and set our custom role query.
         $query->set( 'role', '' );
         $query->set( 'role__in', [ 'seller', 'administrator' ] );
--- a/includes/Admin/UserList.php
+++ b/includes/Admin/UserList.php
@@ -71,8 +71,11 @@ class UserList {
             return $query;
         }
 
+        // Core copies $_REQUEST['role'] into the query var raw, so normalise it the same way the guard above does.
+        $query_role = sanitize_text_field( wp_unslash( $query->get( 'role' ) ) );
+
         // Only the list table asks for this role; the pending-count badge query on the same screen already carries its own criteria.
-        if ( 'pending_vendor' !== $query->get( 'role' ) ) {
+        if ( 'pending_vendor' !== $query_role ) {
             return $query;
         }
 
--- a/includes/Admin/UserList.php
+++ b/includes/Admin/UserList.php
@@ -71,8 +71,7 @@ class UserList {
             return $query;
         }
 
-        // Only the list table asks for this role; every other user query on the screen -- the pending
-        // count badge among them -- already carries its own criteria and must not be stacked on to.
+        // Only the list table asks for this role; the pending-count badge query on the same screen already carries its own criteria.
         if ( 'pending_vendor' !== $query->get( 'role' ) ) {
             return $query;
         }


MdAsifHossainNadim:
## Verdict: Approve ✅

### What this fixes

`UserList::filter_pending_vendors()` hooks `pre_get_users`, which fires for **every** `WP_User_Query` built while `users.php?role=pending_vendor` renders — not just the list table's. It then appends its `dokan_enable_selling` group with a bare `$meta_query[] = …` and never sets a relation.

That last detail is the real root cause. `Manager::get_vendors()` builds `[ 'relation' => 'OR', [ dokan_enable_selling = 'no' ] ]` (`Manager.php:63`), so the injected group gets **OR'd** into it rather than AND'd — which both explodes the joins and *widens* the result set. One mechanism, two symptoms.

Measured live on a single render of that page (59 users, 2,725 `wp_usermeta` rows, Redis + OPcache cleared between runs). The screen issues **four** `WP_User_Query` calls, and the reporter's `4 LEFT JOIN / 3 unkeyed / 3 clauses` shape appears **twice**:

| # | Caller | before | after |
|---|---|---|---|
| 1 | `Manager::get_vendors` (pending badge) | 4 LEFT JOINs (3 unkeyed), 3 clauses | 0 LEFT / 2 INNER, 1 clause |
| 2 | `WP_Users_List_Table::prepare_items` | 3 LEFT JOINs, 2 clauses | **identical** |
| 3 | `dokan_get_seller_status_count` (active) | 4 LEFT JOINs (3 unkeyed), 3 clauses | 0 LEFT / 2 INNER, 1 clause |
| 4 | `dokan_get_seller_status_count` (all) | 3 LEFT JOINs, 2 clauses | 0 LEFT / 1 INNER, 0 clauses |

It is also a **correctness** fix, not only a performance one. Values returned during that same request:

| | before | after | correct |
|---|---|---|---|
| `dokan_get_pending_vendor_count()` | 10 | **1** | 1 |
| `dokan_get_seller_status_count()` | `{total:10, active:0, inactive:10}` | **`{total:43, active:33, inactive:10}`** | the latter |

`dokan_get_pending_vendor_count()` writes into the **persistent** `vendors` cache group, so that wrong `10` was being served to the Dokan → Vendors sidebar badge site-wide. The approve flow is the reachable trigger: `make_active()` invalidates the group, and the redirect straight back to `users.php?role=pending_vendor` re-warms it from the polluted query. Verified in the browser — badge read `10` before, `1` after.

### Why the fix is correct

`WP_Users_List_Table::prepare_items()` always puts the requested role on its query — `$role = $_REQUEST['role'] ?? '';` (`class-wp-users-list-table.php:88`) then `'role' => $role` (`:107`) — in every branch except `'none' === $role`, which can never coincide with `pending_vendor`. The badge and status-count queries scope with `role__in` and never set `role` at all. So keying on that one query var separates exactly the query that needs the filter from the three that don't, with no allowlist to keep in sync as callers change.

`pre_get_users` fires at `class-wp-user-query.php:283` via `do_action_ref_array`, *after* `fill_query_vars()`, so `role` is always populated when the callback reads it.

And structurally the gate sits after both pre-existing guards, so the set of queries it mutates is a strict subset of what it was before: it can only stop the filter from firing, never start it. The only failure mode it can introduce is under-application — which is what the line below closes.

### One line on top of the original fix

Live A/B turned up one edge the first version missed. The pre-existing guard normalises its input (`sanitize_text_field( wp_unslash( $_GET['role'] ) )`), but the new gate compared the **raw** query var, which core copies out of `$_REQUEST['role']` unsanitised. Since `sanitize_text_field()` collapses `[\r\n\t ]+` and trims, a padded value passed the guard and failed the gate: `?role=pending_vendor%09` rendered a fully-chromed Pending Vendor screen — tab marked current, count badge, Approve Vendors offered — over **“No users found.”**, where `develop` correctly listed the 10 rows.

```php
+ // Core copies $_REQUEST['role'] into the query var raw, so normalise it the same way the guard above does.
+ $query_role = sanitize_text_field( wp_unslash( $query->get( 'role' ) ) );
+
  // Only the list table asks for this role; the pending-count badge query on the same screen already carries its own criteria.
- if ( 'pending_vendor' !== $query->get( 'role' ) ) {
+ if ( 'pending_vendor' !== $query_role ) {
      return $query;
  }
```

Re-verified after the change: the padded URL renders the same 10 rows as the clean one, the badge query is still gated (`meta_query` untouched, 0 LEFT JOINs), and PHPCS is clean.

### Testing scope

**Needs verification**

* **Performance** — on a site with a few hundred sellers and no persistent object cache, open **Users → Pending Vendor** and watch the slow query log / `SHOW PROCESSLIST`. No `WP_User_Query` with four `LEFT JOIN wp_usermeta` should appear. The issue reported 141,238,050 rows examined at 193–221 s per request.
* **The list is unchanged** — same rows, same count, same order as before the patch. Confirm search, column sorting and pagination still compose: `?role=pending_vendor&s=…&orderby=login&order=desc`.
* **Approve Vendors** — select pending vendors, run the bulk action, confirm they are approved, the admin notice shows, the redirect lands back on the filtered tab, and the sidebar badge updates.
* **Both counters** — the Users-page “Pending Vendor (N)” tab and the Dokan → Vendors sidebar badge. **They legitimately differ**: the tab counts users whose `dokan_enable_selling` is missing *or* `'no'`, the badge only `'no'` (deliberate — see `functions.php:3488-3496`). Before this PR the poisoned cache made them agree by accident, so on a site with sellers that carry no meta row QA will see the badge drop after the update. That is #3321 / #3400 territory, not a regression here.
* **Padded URL** — `?role=pending_vendor%09` and `?role=%20pending_vendor` should render the same list as the clean URL.

**Should be unaffected**

* Every other Users-screen filter (`All`, `Administrator`, `Customer`, `Vendor`, `Vendor Staff`, `No role`) and `users.php` with no `role` at all.
* Vendor create / edit / approve away from this screen, the Dokan → Vendors list page, and the REST vendor endpoints.
* Dokan Pro — no module references `dokan_get_pending_vendor_count`, `dokan_get_seller_status_count`, either filter, or the string `pending_vendor`; and no Pro or third-party `pre_get_users` callback reads, sets or unsets `role`.

**One pass if you support multisite** (not covered here — no multisite install available): `$pagenow` is also `users.php` on `/wp-admin/network/users.php`, and `WP_MS_Users_List_Table` never sets `role`, so the gate bails there. Note `develop`’s output on that URL was already wrong — the table passes `blog_id => 0` and `class-wp-user-query.php:557` drops all role clauses on a falsy blog id — and the URL has no UI entry point, since Dokan hooks `views_users` while that screen fires `views_users-network`.

### For the changelog

`filter_pending_vendors` has existed since 4.2.4, but it only became pathological in **5.0.13**, when `dokan_get_pending_vendor_count()` was added and started running a `get_vendors()` query on every admin menu build. That scopes the affected versions for support: 5.0.13 → 5.0.18.

MdAsifHossainNadim:
PR Goes [Here](https://github.com/getdokan/dokan/pull/3400).

PR #3400, opened 2026-09-02
fix(vendor): treat a missing selling flag as pending when listing vendors

MdAsifHossainNadim:

Three code paths defined "pending" three different ways:

| Path | Definition |
|---|---|
| `dokan_get_seller_status_count()` | `all − active`, so a **missing** key counts as pending |
| `UserList::filter_pending_vendors()` | `dokan_enable_selling = 'no'` **OR** `NOT EXISTS` |
| `Vendor\Manager::get_vendors()` | `dokan_enable_selling = 'no'` only |

A vendor that never had the meta written — created before the flag existed, or through a path that doesn't set it — is therefore **counted** as pending but can never be **listed**. The Vendors screen shows a non-zero Pending tab that opens on "No data found".

`get_vendors()` is the odd one out, so it now agrees with the other two. Approved is untouched and still matches `'yes'` exactly.

The `status` argument also gained an explicit whitelist while the two were being reconciled. Only `all`, `approved` and `pending` are understood; asking for both halves (or for `all` outright) means everyone; anything else names no status this listing can filter on and falls back to the listing's own default, `approved`, rather than widening the result. That is the coercion `Abilities\Definitions\VendorsQuery::resolve_status()` already applies to the same three statuses.

### Related Pull Request(s)

* None. `UserList::filter_pending_vendors()` touches the same meta on `pre_get_users` and is fixed separately in #3399.

### Refs

* Refs #3321

Not `Closes`: #3321 reports **All (0), Approved (0), Pending (0)** with "No data found" while a direct REST call returns two stores. This PR fixes a **non-zero** Pending counter that opens onto an empty list — a different failure. The tab counters come from the `X-Status-*` response headers, i.e. `dokan_get_seller_status_count()` over `role__in = [ seller, administrator ]`, so **All can never be 0** while an administrator is looking at the page — the viewing admin is in the result set themselves. The reported screen matches `src/admin/dashboard/pages/vendors.tsx` instead, where `counts` initialises to `{ all: 0, approved: 0, pending: 0 }` and the `catch` in `fetchVendors` resets `data` and `totalItems` but never `counts`: a failed or unauthorised **first** fetch renders exactly three zero tabs and "No data found". That also explains the reporter's healthy-looking manual REST check — an unauthenticated `GET /dokan/v1/stores` returns `200` with rows but **without** the `X-Status-*` headers. Worth asking the reporter for the Vendors screen's own Network entry (status code plus response headers) before closing that issue.

1. Make sure at least one seller has **no** `dokan_enable_selling` user meta at all:
   ```sql
   SELECT u.ID FROM wp_users u
     JOIN wp_usermeta c ON c.user_id = u.ID AND c.meta_key = 'wp_capabilities' AND c.meta_value LIKE '%"seller"%'
     LEFT JOIN wp_usermeta e ON e.user_id = u.ID AND e.meta_key = 'dokan_enable_selling'
    WHERE e.umeta_id IS NULL;
   ```
2. Open **Dokan → Vendors** and click the **Pending** tab.

**Expected:** the tab lists exactly as many vendors as its counter claims. Check the **Approved** and **All** tabs are unchanged.

Worth testing on **both** admin surfaces — the legacy screen (`admin.php?page=dokan#/vendors`) and the new dashboard (`admin.php?page=dokan-dashboard#/vendors`). Both consume the same REST endpoint and both were affected.

For the whitelist, `GET /dokan/v1/stores?status=<anything else>` should now return the approved directory instead of the pending set.

### Changelog entry

**Fixed the Vendors screen's Pending tab counting vendors it could not list**

The Pending tab counted every vendor that was not approved, but listed only the ones explicitly disabled. A vendor carrying no selling flag at all — created before the flag existed, or through a path that never wrote it — was counted but never shown, so the tab opened on "No data found" with a non-zero counter next to it. The listing now reads a missing flag as pending, the same way the counters, the Users screen's "Pending Vendors" filter and `dokan_is_seller_enabled()` already did, so the tab and its counter always agree. An unrecognised `status` on the store listing now returns the approved directory instead of a wider set.

### Test results

On a site with 33 vendors carrying `dokan_enable_selling = 'yes'`, **none** carrying `'no'`, and **9** sellers with no key at all:

```
                                     rows   X-WP-Total   All  Approved  Pending
before  status=(none)                 40        40        40     31        9
        status=approved               31        31        40     31        9
        status=pending                 0         0        40     31        9   <-- counter says 9, list is empty

after   status=(none)                 40        40        40     31        9
        status=approved               31        31        40     31        9
        status=pending                 9         9        40     31        9   <-- agrees
```

Verified in the browser on both surfaces: the Pending tab went from "No data found" / "No vendors found" (0 items) to listing all 9, with All and Approved unchanged. Re-verified after the whitelist change — All (43) / Approved (34) / Pending (9) on both screens, Pending listing 9 rows.

PHPCS on the changed files: 0 errors. `Manager.php` keeps one pre-existing `slow_db_query_meta_query` warning; `functions.php` is at its `develop` warning count.

`VendorStatusFilterTest`: 16 tests, 60 assertions. The nested-listing test fails against `9a5229238`, and the three whitelist cases fail against `75df70e93`.

### Behaviour notes

* **Flagless administrators appear under Pending.** `role__in` defaults to `[ seller, administrator ]` and `dokan_admin_user_register()` only writes `dokan_enable_selling` for the `seller` role, so an administrator who never touched the seller fields has no flag. `dokan_get_seller_status_count()` has counted them as inactive since 2.9.23 and the Users-screen "Pending Vendors" filter (`UserList::filter_pending_vendors()`) already lists them; `get_vendors()` was the one place that did not. This PR makes the Vendors screen agree with the other two rather than carving administrators out of all three — that is a separate decision if anyone wants it.
* **Pro fallout, accepted deliberately:** dokan-pro's `Announcement\Manager` sends the `disabled_seller` announcement to `status => [ 'pending' ]` with no `role__in` override, so a flagless administrator now receives it. Measured on a site with 9 flagless sellers: `disabled_seller` goes from 0 recipients to 10. The same account is already in the `all_seller` audience today (that type asks for `status => [ 'all' ]`, which has always included administrators), so this makes the announcement audiences consistent rather than introducing admins to them. Scoping the vendor announcement types to `role__in => [ 'seller' ]` is a Pro-side change and belongs to all four types, not just this one, so it is left to a Pro follow-up. Labelled `Dependency With Pro`.
* **An unknown status is no longer read as pending.** Before, anything that was not `approved` or `all` meant `dokan_enable_selling = 'no'`; the first cut of this PR carried that forward onto the much wider "everyone not approved" set, and a mix such as `[ 'approved', 'rejected' ]` collapsed to `all`. Both now resolve to `approved`. A genuine mix of the two halves — `[ 'approved', 'pending' ]` — still means everyone, as it did before this PR through the OR meta group. `[ 'all', 'approved' ]` used to collapse to approved-only and now means everyone, which is what it asks for. No caller in Lite or Pro passes a mix or an unknown status.
* **`dokan_get_seller_count()` agrees too.** The admin dashboard's vendor widget takes its inactive figure from `status => 'pending'`, so it now matches the badge as well. That listing is capped at one row, since only the total is read and the pending set is larger than it used to be.
* **The `pre_user_query` callback stays registered** after the first pending listing. It is a no-op for any query without the `dokan_pending_only` var; unhooking it around the query let a nested `get_vendors()` (from inside `pre_get_users`) disarm the outer one. Covered by `test_nested_listing_does_not_disarm_the_pending_filter`.

### Notes for the reviewer

Two adjacent things found while investigating, both left out of this PR:

1. `src/stores/vendors/resolvers.ts` — the `getVendors` resolver fetches `/dokan/v1/stores` and then never dispatches `setVendors()` or clears `setLoading( true )`. This matches the "Source-code observation" in #3321, but that store is not what drives either vendors list (`pages/vendors.tsx` does its own `apiFetch`), so it is dead-code rot rather than the cause. Worth its own issue.
2. The `status` collection param on `GET /dokan/v1/stores` still declares no `enum`. With the whitelist in place an unknown value is coerced to `approved` rather than silently widening, so this is now cosmetic — an `enum` would turn it into a `400` instead. Note the endpoint is `permission_callback => '__return_true'` while `Abilities\Definitions\VendorsQuery::resolve_status()` forces non-Store-Admins to `approved`; the REST controller arguably should do the same. Worth its own issue.

### DIFF
--- a/includes/Vendor/Manager.php
+++ b/includes/Vendor/Manager.php
@@ -70,7 +70,7 @@ class Manager {
         }
 
         // if featured
-        if ( 'yes' == $args['featured'] ) {
+        if ( 'yes' === $args['featured'] ) {
             $args['meta_query']['relation'] = 'AND';
             $args['meta_query'][] = [
                 'key'     => 'dokan_feature_seller',
@@ -210,26 +210,28 @@ class Manager {
         /**
          * @since 3.2.7 added $data parameter
          */
-        $store_data = apply_filters( 'dokan_vendor_create_data', [
-            'store_name'              => ! empty( $data['store_name'] ) ? $data['store_name'] : '',
-            'social'                  => ! empty( $data['social'] ) ? $data['social'] : [],
-            'payment'                 => ! empty( $data['payment'] ) ? $data['payment'] : [
-                'paypal' => [ 'email' ],
-                'bank'   => [],
-            ],
-            'phone'                   => ! empty( $data['phone'] ) ? $data['phone'] : '',
-            'show_email'              => ! empty( $data['show_email'] ) ? $data['show_email'] : 'no',
-            'address'                 => ! empty( $data['address'] ) ? $data['address'] : [],
-            'location'                => ! empty( $data['location'] ) ? $data['location'] : '',
-            'banner'                  => ! empty( $data['banner_id'] ) ? $data['banner_id'] : 0,
-            'icon'                    => ! empty( $data['icon'] ) ? $data['icon'] : '',
-            'gravatar'                => ! empty( $data['gravatar_id'] ) ? $data['gravatar_id'] : 0,
-            'enable_tnc'              => ! empty( $data['enable_tnc'] ) ? $data['enable_tnc'] : 'off',
-            'store_tnc'               => ! empty( $data['store_tnc'] ) ? $data['store_tnc'] : '',
-            'show_min_order_discount' => ! empty( $data['show_min_order_discount'] ) ? $data['show_min_order_discount'] : 'no',
-            'store_seo'               => ! empty( $data['store_seo'] ) ? $data['store_seo'] : [],
-            'dokan_store_time'        => ! empty( $data['store_open_close'] ) ? $data['store_open_close'] : [],
-        ], $data );
+        $store_data = apply_filters(
+            'dokan_vendor_create_data', [
+				'store_name'              => ! empty( $data['store_name'] ) ? $data['store_name'] : '',
+				'social'                  => ! empty( $data['social'] ) ? $data['social'] : [],
+				'payment'                 => ! empty( $data['payment'] ) ? $data['payment'] : [
+					'paypal' => [ 'email' ],
+					'bank'   => [],
+				],
+				'phone'                   => ! empty( $data['phone'] ) ? $data['phone'] : '',
+				'show_email'              => ! empty( $data['show_email'] ) ? $data['show_email'] : 'no',
+				'address'                 => ! empty( $data['address'] ) ? $data['address'] : [],
+				'location'                => ! empty( $data['location'] ) ? $data['location'] : '',
+				'banner'                  => ! empty( $data['banner_id'] ) ? $data['banner_id'] : 0,
+				'icon'                    => ! empty( $data['icon'] ) ? $data['icon'] : '',
+				'gravatar'                => ! empty( $data['gravatar_id'] ) ? $data['gravatar_id'] : 0,
+				'enable_tnc'              => ! empty( $data['enable_tnc'] ) ? $data['enable_tnc'] : 'off',
+				'store_tnc'               => ! empty( $data['store_tnc'] ) ? $data['store_tnc'] : '',
+				'show_min_order_discount' => ! empty( $data['show_min_order_discount'] ) ? $data['show_min_order_discount'] : 'no',
+				'store_seo'               => ! empty( $data['store_seo'] ) ? $data['store_seo'] : [],
+				'dokan_store_time'        => ! empty( $data['store_open_close'] ) ? $data['store_open_close'] : [],
+			], $data
+        );
 
         $vendor = dokan()->vendor->get( $vendor_id );
 
@@ -425,7 +427,7 @@ class Manager {
 
         // for backward compatibility we'll allow both `enable_tnc` and `toc_enabled` to set store trams and condition settings
         if ( ( isset( $data['enable_tnc'] ) && dokan_validate_boolean( $data['enable_tnc'] ) )
-             || ( isset( $data['toc_enabled'] ) && dokan_validate_boolean( $data['toc_enabled'] ) ) ) {
+            || ( isset( $data['toc_enabled'] ) && dokan_validate_boolean( $data['toc_enabled'] ) ) ) {
             $vendor->set_enable_tnc( 'on' );
         } else {
             $vendor->set_enable_tnc( 'off' );
--- a/includes/Vendor/Manager.php
+++ b/includes/Vendor/Manager.php
@@ -77,9 +77,7 @@ class Manager {
                 continue;
             }
 
-            // A vendor that never had the meta written carries no key at all, and both the status
-            // counters and the Users list read that as pending -- so the listing has to agree, or the
-            // tab reports a count it cannot show.
+            // The status counters and the Users list both read a missing key as pending, so the listing has to agree or the tab shows a count it cannot list.
             $meta_query[] = [
                 'relation' => 'OR',
                 [
--- a/includes/Vendor/Manager.php
+++ b/includes/Vendor/Manager.php
@@ -67,10 +67,30 @@ class Manager {
                 continue;
             }
 
+            if ( 'approved' === $stat ) {
+                $meta_query[] = [
+                    'key'     => 'dokan_enable_selling',
+                    'value'   => 'yes',
+                    'compare' => '=',
+                ];
+
+                continue;
+            }
+
+            // A vendor that never had the meta written carries no key at all, and both the status
+            // counters and the Users list read that as pending -- so the listing has to agree, or the
+            // tab reports a count it cannot show.
             $meta_query[] = [
-                'key'     => 'dokan_enable_selling',
-                'value'   => ( $stat == 'approved' ) ? 'yes' : 'no',
-                'compare' => '=',
+                'relation' => 'OR',
+                [
+                    'key'     => 'dokan_enable_selling',
+                    'value'   => 'no',
+                    'compare' => '=',
+                ],
+                [
+                    'key'     => 'dokan_enable_selling',
+                    'compare' => 'NOT EXISTS',
+                ],
             ];
         }
 
--- a/includes/Vendor/Manager.php
+++ b/includes/Vendor/Manager.php
@@ -58,28 +58,15 @@ class Manager {
 
         $args = wp_parse_args( $args, $defaults );
 
-        $status         = (array) $args['status'];
-        $filter_status  = ! in_array( 'all', $status, true );
-        $wants_approved = $filter_status && in_array( 'approved', $status, true );
-        $wants_pending  = $filter_status && (bool) array_diff( $status, [ 'approved' ] );
+        $status = $this->resolve_status( $args['status'] );
 
-        // Asking for approved and pending together is asking for everyone.
-        if ( $wants_approved && ! $wants_pending ) {
-            $meta_query = [
-                'relation' => 'OR',
-                [
-                    'key'     => 'dokan_enable_selling',
-                    'value'   => 'yes',
-                    'compare' => '=',
-                ],
+        if ( 'approved' === $status ) {
+            $args['meta_query']['relation'] = 'AND';
+            $args['meta_query'][]           = [
+                'key'     => 'dokan_enable_selling',
+                'value'   => 'yes',
+                'compare' => '=',
             ];
-
-            if ( ! empty( $args['meta_query'] ) ) {
-                $args['meta_query']['relation'] = 'AND';
-                $args['meta_query'][]           = $meta_query;
-            } else {
-                $args['meta_query'] = $meta_query;
-            }
         }
 
         // if featured
@@ -95,8 +82,8 @@ class Manager {
         unset( $args['status'] );
         unset( $args['featured'] );
 
-        // Pending is everyone not approved. One keyed subquery costs no usermeta joins; an OR group with NOT EXISTS costs three.
-        if ( $wants_pending && ! $wants_approved ) {
+        // Pending is everyone not approved, a missing flag included, exactly as the status counters and dokan_is_seller_enabled() read it.
+        if ( 'pending' === $status ) {
             add_action( 'pre_user_query', [ $this, 'exclude_approved_vendors' ] );
         }
 
@@ -119,10 +106,35 @@ class Manager {
     }
 
     /**
-     * Drop every vendor whose selling flag is approved from a user query.
+     * Collapse the requested statuses into the single filter the query applies.
      *
-     * The status counters and the Users list both read a missing dokan_enable_selling key as pending,
-     * so the listing has to agree or the Pending tab reports a count it cannot show.
+     * Anything other than 'approved' or 'all' has always been read as pending, and asking
+     * for approved and pending together is asking for everyone.
+     *
+     * @since DOKAN_SINCE
+     *
+     * @param string|string[] $status
+     *
+     * @return string One of 'all', 'approved' or 'pending'.
+     */
+    private function resolve_status( $status ): string {
+        $statuses = array_unique(
+            array_map(
+                static function ( $item ) {
+                    return in_array( $item, [ 'all', 'approved' ], true ) ? $item : 'pending';
+                },
+                (array) $status
+            )
+        );
+
+        return 1 === count( $statuses ) ? reset( $statuses ) : 'all';
+    }
+
+    /**
+     * Drop every approved vendor from a user query, leaving the pending ones.
+     *
+     * A keyed subquery adds no usermeta join, where the equivalent OR / NOT EXISTS meta
+     * query adds two and turns every join on the query into a LEFT JOIN.
      *
      * @since DOKAN_SINCE
      *
--- a/includes/Vendor/Manager.php
+++ b/includes/Vendor/Manager.php
@@ -58,45 +58,28 @@ class Manager {
 
         $args = wp_parse_args( $args, $defaults );
 
-        $status = (array) $args['status'];
+        $status         = (array) $args['status'];
+        $filter_status  = ! in_array( 'all', $status, true );
+        $wants_approved = $filter_status && in_array( 'approved', $status, true );
+        $wants_pending  = $filter_status && (bool) array_diff( $status, [ 'approved' ] );
 
-        $meta_query = [ 'relation' => 'OR' ];
-
-        foreach ( $status as $stat ) {
-            if ( $stat === 'all' ) {
-                continue;
-            }
-
-            if ( 'approved' === $stat ) {
-                $meta_query[] = [
-                    'key'     => 'dokan_enable_selling',
-                    'value'   => 'yes',
-                    'compare' => '=',
-                ];
-
-                continue;
-            }
-
-            // The status counters and the Users list both read a missing key as pending, so the listing has to agree or the tab shows a count it cannot list.
-            $meta_query[] = [
+        // Asking for approved and pending together is asking for everyone.
+        if ( $wants_approved && ! $wants_pending ) {
+            $meta_query = [
                 'relation' => 'OR',
                 [
                     'key'     => 'dokan_enable_selling',
-                    'value'   => 'no',
+                    'value'   => 'yes',
                     'compare' => '=',
                 ],
-                [
-                    'key'     => 'dokan_enable_selling',
-                    'compare' => 'NOT EXISTS',
-                ],
             ];
-        }
 
-        if ( ! empty( $args['meta_query'] ) ) {
-            $args['meta_query']['relation'] = 'AND';
-            $args['meta_query'][]           = $meta_query;
-        } else {
-            $args['meta_query'] = $meta_query;
+            if ( ! empty( $args['meta_query'] ) ) {
+                $args['meta_query']['relation'] = 'AND';
+                $args['meta_query'][]           = $meta_query;
+            } else {
+                $args['meta_query'] = $meta_query;
+            }
         }
 
         // if featured
@@ -112,8 +95,15 @@ class Manager {
         unset( $args['status'] );
         unset( $args['featured'] );
 
+        // Pending is everyone not approved. One keyed subquery costs no usermeta joins; an OR group with NOT EXISTS costs three.
+        if ( $wants_pending && ! $wants_approved ) {
+            add_action( 'pre_user_query', [ $this, 'exclude_approved_vendors' ] );
+        }
+
         $user_query = new WP_User_Query( $args );
         $results    = $user_query->get_results();
+
+        remove_action( 'pre_user_query', [ $this, 'exclude_approved_vendors' ] );
 
         $this->total_users = $user_query->total_users;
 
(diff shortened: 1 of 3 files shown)


MdAsifHossainNadim (commented):
_Posted as a comment rather than a formal approval: GitHub does not allow the author of a pull request to stamp `Approve` on it. Treat this as an approving review — it needs a second pair of eyes for the green check._

---

Reviewed at `75df70e93`. **No blockers in the code — this is good to merge.** One change I would make before it goes in is in the description rather than the diff: the issue link should be `Refs #3321`, not `Closes #3321`.

## What the change gets right

Three code paths defined "pending" three different ways and the vendor listing was the odd one out. Reading a missing `dokan_enable_selling` flag as pending brings `get_vendors()` in line with `dokan_get_seller_status_count()`, `UserList::filter_pending_vendors()` and `dokan_is_seller_enabled()`. That is the right direction to reconcile in — widening the listing to match the counters, rather than narrowing three counters to match one listing.

I re-verified the mechanics against the current tree:

* Core builds the user-query cache key and the `FOUND_ROWS()` total from the final SQL, **after** `pre_user_query` has run. The injected `NOT EXISTS` is therefore honoured by the rows, the total and the object cache alike — no risk of a count that disagrees with the list it is counting.
* `Vendor\Manager` only ever exists as the `dokan()->vendor` singleton, so the `has_action()` hook-once guard cannot double-append the clause.
* The correlated `NOT EXISTS` is the right shape here. It rides the `usermeta` index and stops at the first match per row, where the previous OR / meta-query form forced a full `wp_users` scan with three joins under `DISTINCT`.
* Keeping the callback registered and gating it on the `dokan_pending_only` query var is the correct fix for the nested-query problem. Unhooking around the query let a nested listing disarm the outer one, and `test_nested_listing_does_not_disarm_the_pending_filter` pins exactly that.

Branch is even with `develop` and mergeable. PHPCS and PHPUnit 8.3 are green. The 14 tests cover the status matrix, counter-versus-list agreement, both hook-leak directions, caller-supplied meta queries and the featured filter.

## The one thing to change: `Closes #3321` → `Refs #3321`

#3321 reports **All (0), Approved (0), Pending (0)** with "No data found", while a direct REST call returns two stores. This PR fixes a **non-zero** Pending counter that opens onto an empty list. Those are different failures, and merging this will not close the reported one.

The tab counters come from the `X-Status-*` response headers, i.e. `dokan_get_seller_status_count()` over `role__in = [ seller, administrator ]`. **All can never be `0`** while an administrator is looking at the page — the viewing admin is themselves in the result set. So the reported screen cannot be produced by any server-side status filter.

It matches the front end instead. In `src/admin/dashboard/pages/vendors.tsx`, `counts` initialises to `{ all: 0, approved: 0, pending: 0 }` and the `catch` in `fetchVendors` resets `data` and `totalItems` but never touches `counts`. A failed or unauthorised **first** fetch therefore renders exactly the reported screen: three zero tabs and "No data found".

That also explains why the reporter's manual REST check looked healthy — an unauthenticated `GET /dokan/v1/stores` returns `200` with rows but **without** the `X-Status-*` headers.

Suggest keeping #3321 open and asking the reporter for the Vendors screen's own Network entry (status code plus response headers), since the console does log `Failed to fetch vendors` on that path.

## Follow-ups, none blocking this PR

1. **Flagless administrators now list as pending.** `dokan_admin_user_register()` only writes the flag for the `seller` role, so a fresh install will show a `Vendors (1)` badge until the admin toggles their own store. The behaviour is consistent with the other two paths, so it belongs here — but the principled fix is for the Installer to write `yes` for administrators alongside the `dokandar` capability. Worth its own issue.
2. **The stores `status` collection param has no `enum`.** Any unrecognised string now resolves to the wider pending set instead of the old narrower one. See the inline note.
3. **`UserList::filter_pending_vendors()` still hand-builds its own OR / `NOT EXISTS` group** and could delegate to the new query var. That file is in flight in #3399, so leave it alone until that lands.
4. The reformat of the `dokan_vendor_create_data` filter call reads as churn but is load-bearing — CI runs PHPCS over the whole changed file and `develop`'s copy carries pre-existing errors.

## CI

The **PHPUnit (PHP 7.4)** job is red for infrastructure reasons, not for this PR. It fails after ~2m in `Start WordPress environment`, at `RUN apt-get -qy install sudo` in the wp-env image build, against a Debian mirror that no longer serves those paths. It never reaches a test. The same job fails identically on other PRs from the same day.

https://claude.ai/code/session_01JVRcXVe4nsfSJCZ65AbEKC

includes/Vendor/Manager.php
```diff
@@ -110,6 +113,57 @@ public function get_vendors( $args = [] ) {
...
+        $statuses = array_unique(
+            array_map(
+                static function ( $item ) {
+                    return in_array( $item, [ 'all', 'approved' ], true ) ? $item : 'pending';
+                },
+                (array) $status
+            )
+        );
+
+        return 1 === count( $statuses ) ? reset( $statuses ) : 'all';
```
MdAsifHossainNadim: Worth a follow-up: `resolve_status()` maps anything that is not `all` or `approved` onto `pending`, which preserves the historical behaviour and is fine in itself. The wrinkle is that `StoreController::get_stores()` passes `$params['status']` straight through and the collection param declares no `enum`, so a typo in the query string used to land on the narrow `dokan_enable_selling = 'no'` set and now lands on the much wider "everyone not approved" set.

Not a problem for any caller in Lite or Pro, and not something to fix in this PR. An `enum` of `all` / `approved` / `pending` on the collection param would close it properly.

includes/functions.php
```diff
@@ -1500,7 +1500,7 @@ function dokan_get_seller_count( $from = null, $to = null ) {
     $now              = dokan_current_datetime();
     $inactive_sellers = dokan_get_sellers(
         [
-            'number' => - 1,
+            'number' => 1, // Only the total is read; one row keeps this from building a Vendor object per pending seller.
```
MdAsifHossainNadim, line 1503: Good catch capping this — only `count` is read out of the return, and the pending set is materially larger after this change, so the old `number => -1` would have built a `Vendor` object per pending seller for nothing.

For the record: the `$active_sellers` call just below still uses `number => -1` with full objects. Pre-existing and correctly left out of scope here, but it is the same waste and worth sweeping up separately.

kzamanbd (changes requested):
Fix is correct and the mechanism checks out — `pre_user_query` lands before the request is built, the cache key hashes the full SQL, the `has_action` guard is safe because `vendor` is `addShared`, and `number => 1` still yields a real `count`. A few things to resolve first.

**1. `includes/Vendor/Manager.php:135` — `resolve_status()` fails open.**
Any mix resolves to `all`, so a typo like `[ 'approved', 'rejected' ]` returns every vendor on a listing that feeds the public `/dokan/v1/stores`. Please use an explicit whitelist and default the ambiguous case to the narrowest set.

**2. `includes/Vendor/Manager.php:92` — Pro fallout.**
`pending` now includes administrators (`role__in` default is `[ seller, administrator ]`). dokan-pro `includes/Announcement/Manager.php:382` passes `status => [ 'pending' ]` for `disabled_seller`, so that announcement will now reach admin accounts. Needs a Pro-side `role__in => [ 'seller' ]` or an explicit accept. Please add the `Dependency With Pro` label.

**3. `tests/php/src/Vendor/VendorStatusFilterTest.php:118` — name contradicts the design.**
`75df70e` deliberately leaves the callback hooked; the test asserts it is a no-op for other queries. Rename to something like `test_hooked_filter_is_a_noop_for_other_queries`.

**4. PR description — one row of the behaviour table is wrong.**
`[ 'approved', 'pending' ]` did not collapse to approved-only; the old OR meta group returned the union. `[ 'all', 'approved' ]` is the pair that collapsed.

**Notes (not blocking):**
- `PHPUnit (PHP 7.4)` is red at "Start WordPress environment" — `Run PHPUnit` was skipped. Infra, please re-run job `101633730661`.
- Follow-up worth filing: `GET /dokan/v1/stores` is `permission_callback => '__return_true'` with a public `status` param. `Abilities\Definitions\VendorsQuery::resolve_status()` already forces non-Store-Admins to `approved`; the REST controller should do the same.
- Needs a changelog entry.

kzamanbd (commented):
Anchoring the three items from my review to the lines they apply to.

includes/Vendor/Manager.php
kzamanbd: Fails open: any mix resolves to `all`, so a typo like `[ 'approved', 'rejected' ]` returns **every** vendor — on a listing that feeds the public `/dokan/v1/stores`. Prefer an explicit whitelist, and default the ambiguous case to the narrowest set rather than the widest.

Note the old code only collapsed `[ 'all', 'approved' ]` to approved-only; `[ 'approved', 'pending' ]` already returned the union via the OR meta group. The behaviour table in the PR description has that row backwards.

includes/Vendor/Manager.php
```diff
@@ -94,6 +86,17 @@ public function get_vendors( $args = [] ) {
         unset( $args['status'] );
         unset( $args['featured'] );
 
+        // Pending is everyone not approved, a missing flag included, exactly as the status counters and dokan_is_seller_enabled() read it.
+        if ( 'pending' === $status ) {
+            // Carried into the query so the callback can tell our query from any other one running inside the same hook.
+            $args['dokan_pending_only'] = true;
```
kzamanbd, line 101: `pending` now includes administrators, since `role__in` defaults to `[ seller, administrator ]`.

dokan-pro `includes/Announcement/Manager.php:382` passes `status => [ 'pending' ]` for the `disabled_seller` announcement type with no `role__in` override, so that announcement will now be delivered to admin accounts. Needs a Pro-side `'role__in' => [ 'seller' ]` or an explicit accept — and the `Dependency With Pro` label here.

tests/php/src/Vendor/VendorStatusFilterTest.php
```diff
@@ -0,0 +1,213 @@
...
+        $counts = dokan_get_seller_status_count();
+
+        $this->assertSame( count( $listed ), $counts['inactive'] );
+        $this->assertSame( count( $listed ), dokan_get_pending_vendor_count() );
+    }
+
+    /**
+     * The callback stays hooked after the listing, so it must sit out every query that did not ask for pending.
+     */
+    public function test_pending_filter_does_not_outlive_the_query() {
```
kzamanbd: Name contradicts the design: `75df70e` deliberately leaves the callback hooked, so the filter *does* outlive the query. What this actually asserts is that it is a no-op for queries that did not ask for pending. Suggest `test_hooked_filter_is_a_noop_for_other_queries`.

includes/Vendor/Manager.php
MdAsifHossainNadim: Fixed in efdb7dac8. `resolve_status()` now intersects the input against an explicit `Manager::STATUSES` whitelist, drops what it does not recognise, and falls back to `approved` — the listing's own default — when nothing recognisable is left. `[ 'approved', 'rejected' ]` resolves to `approved`, `[ 'pending', 'rejected' ]` to `pending`, and a lone `'rejected'`, `''` or `[]` to `approved`. `[ 'approved', 'pending' ]`, and anything containing `all`, still mean everyone. It is the same coercion `Abilities\Definitions\VendorsQuery::resolve_status()` has applied since 5.0.15.

One correction on reachability: the array form cannot come from the public endpoint. `status` is declared `'type' => 'string'` in `get_store_collection_params()`, so `rest_validate_request_arg` rejects it before `get_stores()` runs:

```
GET /wp-json/dokan/v1/stores?status[]=approved&status[]=rejected
400  rest_invalid_param — Invalid parameter(s): status
```

What *was* publicly reachable is a single bogus string, and it widened the same way — measured on a site with 34 approved and 9 flagless sellers:

```
?status=rejected     develop 0 rows  |  75df70e93 9 rows (the whole pending set)  |  now 34 rows (approved)
```

The mix stays reachable from PHP and through the `dokan_seller_listing_args` filter, so the whitelist is the right fix either way. Worth noting for the severity read: the endpoint already returns all 43 anonymously by default (`status` defaults to `all`), so nothing was newly *disclosed* — the rows just landed in the wrong bucket.

You are right about the behaviour table too; the description now says `[ 'approved', 'pending' ]` was already a union through the OR meta group and `[ 'all', 'approved' ]` was the pair that collapsed.

includes/Vendor/Manager.php
MdAsifHossainNadim: Confirmed, and measured rather than reasoned. Deleting the administrator's `dokan_enable_selling` on a site with 9 flagless sellers and running exactly what `Announcement\Manager::create_announcement()` runs — `dokan()->vendor->all( [ 'fields' => 'ID', 'number' => -1, 'status' => [ 'pending' ] ] )` — gives:

```
                  develop                  this branch
disabled_seller   0 recipients             10 recipients  (9 flagless sellers + the flagless admin)
enabled_seller    33, admin not included   33, admin not included
all_seller        43, admin included       43, admin included
```

So the admin account is already in the `all_seller` audience today: that type asks for `status => [ 'all' ]`, which has always resolved through the same `role__in => [ seller, administrator ]` default. This branch adds them to `disabled_seller` as well, which makes the four audiences consistent rather than introducing admins to announcements.

Accepting it explicitly here, and the `Dependency With Pro` label is on. Keeping vendor announcements to real vendors is a Pro-side change and it belongs to all four types, not just `disabled_seller`, so I would rather not special-case one of them from this PR. Happy to file it against dokan-pro so it is tracked, or to push `role__in => [ 'seller' ]` there before this merges — your call, so I am leaving this thread open.

tests/php/src/Vendor/VendorStatusFilterTest.php
MdAsifHossainNadim: Renamed to `test_hooked_filter_is_a_noop_for_other_queries` in eecdb9d33 — the old name described the mechanism 75df70e93 removed. Thanks.

includes/Vendor/Manager.php
MdAsifHossainNadim: Half-closed by efdb7dac8: an unrecognised `status` now resolves to `approved` instead of the wider pending set, so a typo no longer widens anything — `?status=rejected` returns the 34-row approved directory. An `enum` on the collection param would still be better (a `400` beats a silent coercion), so it stays a follow-up, alongside the `permission_callback => '__return_true'` point.

MdAsifHossainNadim:
Thanks — three of the four are fixed, the fourth is verified and accepted. Head is now `eecdb9d33`.

**1. `resolve_status()` failing open — fixed** (`efdb7dac8`). It intersects against an explicit `Manager::STATUSES` whitelist, drops what it does not recognise, and falls back to `approved` — the listing's own default — when nothing recognisable is left, matching what `Abilities\Definitions\VendorsQuery::resolve_status()` has done since 5.0.15.

```
[ 'approved', 'rejected' ]  ->  approved      (was: all)
[ 'pending', 'rejected' ]   ->  pending
'rejected' / '' / []        ->  approved      (was: the full pending set)
[ 'approved', 'pending' ]   ->  all           (unchanged)
[ 'all', 'approved' ]       ->  all           (unchanged)
```

One correction on the reachability, in the thread: the array form never reaches `get_vendors()` from the public endpoint — `status` is declared `'type' => 'string'`, so `?status[]=…` is a `400` from `rest_validate_request_arg`. The single bogus string was reachable, and it widened the same way, so the whitelist is the right fix regardless. `?status=rejected` returned 0 rows on `develop`, 9 (the whole pending set) on `75df70e93`, and 34 (the approved directory) now.

**2. Pro fallout — real, measured, accepted; label added.** With the admin's flag deleted on a site carrying 9 flagless sellers, `disabled_seller` goes from 0 recipients on `develop` to 10 here. The same account is already in the `all_seller` audience today, since that type asks for `status => [ 'all' ]` through the same `role__in => [ seller, administrator ]` default — so this makes the four audiences consistent rather than putting admins on announcements for the first time. Scoping vendor announcements to real vendors is a Pro-side change and belongs to all four types, so I have not special-cased `disabled_seller` from here. Numbers and the offer to file it (or push it) are in the thread, which I have left open for you.

**3. Test name — fixed** (`eecdb9d33`): `test_hooked_filter_is_a_noop_for_other_queries`.

**4. Behaviour table — fixed** in the description. `[ 'approved', 'pending' ]` was already a union through the OR meta group; `[ 'all', 'approved' ]` was the pair that collapsed to approved-only.

**Changelog entry** — added as its own section in the description. `readme.txt` itself only changes in the `chore: Release version …` commits (last three: `7564e22bc`, `1bae8afe2`, `3e1f1a2e3`), so I have left the file alone rather than adding an entry ahead of the release.

**CI** — re-run on the new head. PHPCS and PHPUnit (PHP 8.3) are green. **PHPUnit (PHP 7.4) is not worth re-running**: it dies in `Start WordPress environment`, building the wp-env image, before any test runs —

```
#25 [wordpress 12/23] RUN apt-get -qy update
E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease
   is expired (invalid since 6h 24min 54s)
```

Debian bullseye-security stopped being refreshed, so the release file is permanently expired and every re-run hits it. It is not specific to this PR: every `PHPUnit Tests` run in the repo since 2026-09-07 fails the same way, `develop`'s own runs included (`34119783229`, `34103827451`; last green was `33750097135` on 2026-09-03). Fixing it means pinning or updating the wp-env base image, which is its own PR.

Locally, against `$TMPDIR/wordpress-tests-lib`: the full Lite suite is **453 tests / 1657 assertions, green**, and `VendorStatusFilterTest` is 16 tests / 60 assertions — the three whitelist cases fail against `75df70e93`. PHPCS reports 0 errors on the changed files. Re-verified in the browser on both admin surfaces after the whitelist change: All (43) / Approved (34) / Pending (9), with the Pending tab listing all 9.

includes/Vendor/Manager.php
MdAsifHossainNadim: Two clarifications on my reply above, one a correction and one it left out.

**Correction:** "makes the four audiences consistent" overstates it — `enabled_seller` still (correctly) excludes admins, since they are not `dokan_enable_selling = 'yes'`. The accurate point is narrower: admins were already in the `all_seller` audience (`status => [ 'all' ]`), so this does not introduce admins to the announcement system for the first time.

**The bigger half of that 0 → 10, which I under-sold as fallout:** 9 of the 10 are **flagless sellers**, and including them is the intended behaviour, not a side effect. `dokan_is_seller_enabled()` reads a missing flag as *disabled* (`'yes' === get_user_meta(...)` → false), and a normally-registered vendor is never flagless — registration always writes `'yes'`/`'no'` via `make_active()`/`make_inactive()`. So a flagless seller is a genuinely disabled vendor that the old `= 'no'` test silently skipped; the `disabled_seller` announcement was *meant* to reach them and now does. Only the **1 admin** is the debatable recipient.

That also means the two things do not fight: a Pro-side `role__in => [ 'seller' ]` drops the admin while **keeping** the 9 flagless sellers — so the accept here and the Pro follow-up are compatible, not either/or. Net on `disabled_seller`:

```
                              admins   flagless sellers   explicit-'no' sellers
develop                         out          out                 in
this branch (accept)            in           in                  in
+ Pro role__in => ['seller']    out          in                  in   <- the clean end state
```

Still leaving this open for your call on whether the Pro change lands with this or as a tracked follow-up.

kzamanbd (approved):

PR #3401, opened 2026-09-02
fix(admin): show the new-vendor fields on the legacy Add Vendor form

### LINKED ISSUE #3392: Admin Add New Vendor flow doesn't show required fields
### Bug Description

When adding a vendor from the Vendors admin screen (Dokan -> Vendors -> "Add Vendor"), a number of required fields are not shown, including the username, password, vendor picture, and store URL fields. This makes it impossible to add a vendor via the admin UI, as there is no way to see that the fields are required or enter their values.

Image

### Step To Reproduce

1. Go to wp-admin
2. Navigate to _Dokan_ -> _Vendors_ and then click on "Add Vendor"
3. Scan available fields and note that they are missing, most especially the "Vendor Picture", "Store URL", "Username", and "Password" fields.

### Expected Behaviour

All required fields should be visible so admins can enter the relevant values.

### Actual Behaviour

Many required fields are not visible, and submitting the "Next" button in the footer doesn't seem to do anything.

### Additional Information

Based on a quick code inspection of [VendorAccountFields.vue](https://github.com/getdokan/dokan/blob/8d1e0015550df197d110721005c1aa6c7630c2ba/src/admin/pages/VendorAccountFields.vue), it looks like these hidden fields are all behind `v-if="! getId()"` conditions. It then looks like https://github.com/getdokan/dokan/pull/3290 updated the implementation of multiple `getId()` functions to return `String( <id_from_route> || 0 )`, which will be `"0"` for cases where no route exists. However, `! "0"` is false in vanilla Javascript, so I suspect that is a possible cause for this issue.

```shell
1. Dokan Version: 5.0.17
2. Dokan Pro Version: 5.0.16
3. WooCommerce Version: 11.0.1
4. Browser: Firefox
```

MdAsifHossainNadim:

**1. The hidden fields (the reported bug)**

`getId()` returned `String( parseInt( this.$route.params.id, 10 ) || 0 )`. On the add-vendor route that is the string `"0"` — and `"0"` is **truthy** in JavaScript, so every template condition that reads it as *"are we editing an existing vendor?"* took the edit branch and hid the fields that only exist when creating one.

Seven conditions across two components were affected:

| Component | Hidden by the bug |
|---|---|
| `VendorAccountFields.vue` | vendor picture, banner upload, store URL, username, password, and both required-field markers |
| `VendorPaymentFields.vue` | the `edit-mode` class and the `column` / `checkbox-group` layout |

Returning the number restores the intended semantics and keeps the numeric pin from #3290 — `parseInt` is what stops a crafted route param traversing the REST path; the `String()` wrapper was incidental to that fix.

```
add route (no :id)      before "0"  -> ! getId() === false   fields hidden
                        after   0   -> ! getId() === true    fields shown
edit route (:id = 42)   before "42" -> ! getId() === false   fields hidden
                        after   42  -> ! getId() === false   unchanged
```

Neither of these two components ever concatenates the route id into a REST path — their only `dokan.api` calls are the three `/stores/check` lookups — so `getId()` there is purely an "am I editing?" flag and the `String()` pin bought nothing. The components that **are** the sinks #3290 was written for keep their pin untouched: `VendorSingle.vue:457` and `ReverseWithdrawalTransactions.vue:229`.

**2. Password is required**

The form always POSTs `user_pass`, so the key is present-but-empty when nobody clicks *Generate Password*. `wp_parse_args` in `Vendor\Manager::create()` therefore never applies its `wp_generate_password()` default, and `wp_insert_user` takes its *"force the password reset flow"* branch (`wp-includes/user.php:2287`) — **every vendor created this way got a blank password plus a WP warning.**

`user_pass` is now in `requiredFields`, matching the React dashboard, whose `dokan-create-vendor-required-fields` has listed it all along. The input is always rendered rather than appearing only after *Generate Password* is clicked — a required field the admin cannot see is the same dead end this PR set out to fix. *Cancel* clears the value instead of hiding the input, which also stops a cancelled password still shipping in the payload.

**3. Form rhythm around the slot fields**

The slot wrapper Pro fills (`.form-group`) is a plain block, so its floated cells staggered instead of pairing — one cell taller than its partner pushed the next out of the left lane and left a hole. It now gets the same flex row model as `.dokan-form-group`, so unequal heights can never open a gap.

The email toggle and the password generator stand in for inputs at the end of their columns but carried no bottom margin, leaving 2–3px before the next row against the form's ~17px rhythm. They now carry the same `margin-bottom` an input does.

### Related Pull Request(s)

* getdokan/dokan-pro#6127 — removes the inline `style="margin-top: 10px"` in Germanized's `VendorTaxFields.vue` that triggered the stagger in the first place. This PR makes the row model resilient; that one removes the trigger.

### Closes

* Closes #3392

1. **Dokan → Vendors → Add New** on the legacy admin screen (`admin.php?page=dokan#/vendors`).
2. Confirm **Vendor Picture**, **Upload Banner**, **Store URL**, **Username** and **Password** are all present, with the markers reading `Store Name *`, `Email *`, `Username *`, `Password *`.
3. Press **Next** with the form empty — `Password is required` should appear in red alongside the other required fields.
4. Click **Generate Password**, then **Cancel** — the input stays and is emptied.
5. Fill the form and create the vendor, then confirm the typed password works (`wp_check_password` against the stored hash).
6. Check the company fields below: **Company Name** and **Company ID/EUID Number** on one baseline, no empty gap, and ~17px of breathing room under the email toggle and the Generate Password button.
7. Open an existing vendor for editing and confirm the create-only fields are still hidden.

### Test results

Verified in the browser on the legacy screen.

* Add: vendor picture with Gravatar link, banner upload, store URL with live `http://…/store/` preview, `Username *`, `Password *` with an always-visible input, the account-email toggle, and `payment-info` without `edit-mode` (PayPal in `checkbox-group`).
* Edit (`#/vendors/2`): create-only fields absent, no required markers, `payment-info edit-mode`, PayPal in `column`.
* End to end: created a vendor with a typed password; `wp_check_password` returns true against the stored hash, so the forced-reset branch is no longer taken.
* Spacing measured in the DOM — gaps went from 3px / 2px to 19px / 18px against a 17px in-form reference.

The **new React dashboard** (`admin.php?page=dokan-dashboard#/vendors/create`) has its own implementation, was checked, and was never affected — it already renders Username, Password and the banner upload. This PR is legacy-only.

### Notes for the reviewer

I also tried moving the "Send the vendor an email" toggle out of the Username cell to close the vertical offset between the two columns, and **backed it out**: `.dokan-form-group` is `display:flex; flex-wrap:wrap` with `.column { width: 50% }`, so cells pair strictly by DOM order and inserting one shifts every pairing below it without fixing the offset. Fixing the missing bottom margins turned out to be the real cause.

The flex row model deliberately lives here rather than in a `<style scoped>` block in the Pro component: `modules/germanized/module.php:183` registers only `script-admin.js` and never a stylesheet, so CSS emitted from that SFC would never be enqueued. Lite owns the grid its slot children sit in.

ESLint's parse errors on these files are pre-existing tooling config, not this change — `develop`'s own copies fail identically.

https://claude.ai/code/session_016XwKvaSv3z32pSu8tBhDQr

### DIFF
--- a/src/admin/pages/AddVendor.vue
+++ b/src/admin/pages/AddVendor.vue
@@ -129,7 +129,8 @@ export default {
             requiredFields: [
                 'store_name',
                 'user_login',
-                'email'
+                'email',
+                'user_pass'
             ],
             errors: [],
             storeAvailable: false,
--- a/src/admin/pages/AddVendor.vue
+++ b/src/admin/pages/AddVendor.vue
@@ -436,6 +436,41 @@ export default {
                         }
                     }
 
+                    .password-field {
+                        position: relative;
+
+                        // Keep the value clear of the reveal control.
+                        .dokan-form-input {
+                            padding-right: 36px;
+                        }
+
+                        .password-toggle {
+                            position: absolute;
+                            top: 6px;
+                            right: 0;
+                            height: 40px;
+                            width: 36px;
+                            padding: 0;
+                            border: 0;
+                            background: none;
+                            cursor: pointer;
+                            color: #82878c;
+
+                            &:hover,
+                            &:focus {
+                                color: #23282d;
+                            }
+
+                            .dashicons {
+                                width: 20px;
+                                height: 20px;
+                                font-size: 20px;
+                                line-height: 20px;
+                                vertical-align: middle;
+                            }
+                        }
+                    }
+
                     .password-generator {
                         margin-top: 6px;
                         // Stands in for an input, so it carries the same bottom rhythm.
--- a/src/admin/pages/AddVendor.vue
+++ b/src/admin/pages/AddVendor.vue
@@ -478,9 +478,16 @@ export default {
 
                         .regen-button {
                             margin-right: 5px;
+                            // Centre the icon against the label instead of guessing a line-height.
+                            display: inline-flex;
+                            align-items: center;
+                            gap: 4px;
 
-                            span {
-                                line-height: 26px;
+                            .dashicons {
+                                width: 20px;
+                                height: 20px;
+                                font-size: 20px;
+                                line-height: 20px;
                             }
                         }
                     }
--- a/src/admin/pages/AddVendor.vue
+++ b/src/admin/pages/AddVendor.vue
@@ -402,6 +402,12 @@ export default {
                     content: " ";
                 }
 
+                // Slot-filled groups (Pro fields) share the row model; floats stagger and leave holes.
+                > .form-group {
+                    display: flex;
+                    flex-wrap: wrap;
+                }
+
                 .column {
                     float: left;
                     width: 50%;
@@ -432,6 +438,8 @@ export default {
 
                     .password-generator {
                         margin-top: 6px;
+                        // Stands in for an input, so it carries the same bottom rhythm.
+                        margin-bottom: 16px;
 
                         .regen-button {
                             margin-right: 5px;
@@ -444,6 +452,7 @@ export default {
 
                     .checkbox-left.notify-vendor {
                         margin-top: 6px;
+                        margin-bottom: 16px;
                     }
 
                     .multiselect {
--- a/src/admin/pages/VendorAccountFields.vue
+++ b/src/admin/pages/VendorAccountFields.vue
@@ -96,7 +96,9 @@
 
                 <template v-if="! getId()">
                     <div class="column">
-                        <label for="user-login">{{ __( 'Username', 'dokan-lite') }}</label><span class="required-field">*</span>
+                        <label for="user-login">{{ __( 'Username', 'dokan-lite') }}</label>
+                        <span class="required-field">*</span>
+
                         <input type="text" class="dokan-form-input"
                             id="user-login"
                             v-model="vendorInfo.user_login"
--- a/src/admin/pages/VendorAccountFields.vue
+++ b/src/admin/pages/VendorAccountFields.vue
@@ -258,8 +258,7 @@ export default {
 
         // getId function has been used to identify whether is it vendor edit page or not
         getId() {
-            // Numeric, not a string: the templates read this as "are we editing?", and "0" is truthy.
-            // parseInt still pins it so a crafted route param can't traverse the REST path (XSS, #3290).
+            // Templates read this as "are we editing?" and the string "0" is truthy; parseInt keeps the numeric pin from #3290.
             return parseInt( this.$route.params.id, 10 ) || 0;
         },
 
--- a/src/admin/pages/VendorAccountFields.vue
+++ b/src/admin/pages/VendorAccountFields.vue
@@ -124,13 +124,23 @@
                         <label for="store-password">{{ __( 'Password', 'dokan-lite') }}</label>
                         <span class="required-field">*</span>
 
-                        <input
-                            id="store-password"
-                            type="text"
-                            v-model="vendorInfo.user_pass"
-                            :class="{'dokan-form-input': true, 'has-error': getError('user_pass')}"
-                            :placeholder="getError( 'user_pass' ) ? __( 'Password is required', 'dokan-lite' ) : __( 'Enter password or Generate', 'dokan-lite' )"
-                        >
+                        <div class="password-field">
+                            <input
+                                id="store-password"
+                                :type="passwordVisible ? 'text' : 'password'"
+                                v-model="vendorInfo.user_pass"
+                                :class="{'dokan-form-input': true, 'has-error': getError('user_pass')}"
+                                :placeholder="getError( 'user_pass' ) ? __( 'Password is required', 'dokan-lite' ) : __( 'Enter password or Generate', 'dokan-lite' )"
+                            >
+
+                            <button type="button"
+                                class="password-toggle"
+                                :aria-pressed="passwordVisible ? 'true' : 'false'"
+                                :aria-label="passwordVisible ? __( 'Hide password', 'dokan-lite' ) : __( 'Show password', 'dokan-lite' )"
+                                @click.prevent="passwordVisible = ! passwordVisible">
+                                <span class="dashicons" :class="passwordVisible ? 'dashicons-hidden' : 'dashicons-visibility'"></span>
+                            </button>
+                        </div>
 
                         <password-generator
                             @passwordGenerated="setPassword"
@@ -183,6 +193,8 @@ export default {
     data() {
         return {
             showStoreUrl: true,
+            // The generated password is readable only on request; masking is the resting state.
+            passwordVisible: false,
             otherStoreUrl: null,
             banner: '',
             defaultUrl: dokan.urls.siteUrl + dokan.urls.storePrefix + '/',
@@ -242,6 +254,7 @@ export default {
         // Cancel only drops the generated value; the input stays so a required field is always fillable.
         this.$root.$on( 'passwordCancelled', () => {
             this.vendorInfo.user_pass = '';
+            this.passwordVisible = false;
         } );
     },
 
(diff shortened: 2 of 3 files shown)


MdAsifHossainNadim:
## Verdict: Approve ✅

### The exact issue

`getId()` returned `String( parseInt( this.$route.params.id, 10 ) || 0 )`. The add-vendor route has no `:id`, so that expression evaluates to the **string** `"0"` — and `"0"` is truthy in JavaScript. Every template condition written as `v-if="! getId()"` (meaning *"are we creating a new vendor?"*) therefore took the **edit** branch and hid the fields that only exist during creation: vendor picture, banner upload, store URL, username, password and both required markers in `VendorAccountFields.vue`, plus the `edit-mode` class and the `column` / `checkbox-group` layout in `VendorPaymentFields.vue`.

A second bug sat downstream of the first. Because the password input never rendered, the form always POSTed `user_pass: ''`. The `wp_parse_args` default of `wp_generate_password()` in `Vendor\Manager::create()` only applies when the key is **absent**, so an empty string went straight into `wp_insert_user`, which takes its *"force the password reset flow"* branch on an empty password (`wp-includes/user.php:2287`). Every vendor created from this form ended up with a blank password and a WP warning.

### Why the fix is correct

**Dropping the `String()` wrapper.** `parseInt` is what actually does the security work in #3290 — it pins the value to a number so a crafted route param can't traverse a REST path. The `String()` wrapper was incidental to that, and it is what broke the truthiness contract. Returning the number restores the intended semantics with the pin fully intact.

This is safe **in these two files specifically**: neither ever concatenates the route id into a REST path — their only `dokan.api` calls are the three `/stores/check` availability lookups, so `getId()` here is purely an *"am I editing?"* flag. The components that genuinely are the sinks #3290 was written for keep their pin untouched — `VendorSingle.vue:457` and `ReverseWithdrawalTransactions.vue:229`.

**Making the password required.** `user_pass` joins `requiredFields`, matching what the React dashboard has enforced all along via `dokan-create-vendor-required-fields`. The input is now always rendered instead of appearing only after *Generate Password* — a required field the admin cannot see is the same dead end this PR set out to fix — and *Cancel* clears the value rather than hiding the input, which also stops a cancelled password still shipping in the payload.

**The layout half.** Pro's Germanized `VendorTaxFields` drops five fields into Lite's account-fields slot inside a bare `<div class="form-group">`, whose `.column` children inherit `float: left; width: 50%` from Lite's descendant rule. Floats place by available space, so one cell taller than its partner staggered every pair below it. Giving that slot wrapper the same flex row model as `.dokan-form-group` means unequal heights can no longer open a hole. The rule deliberately lives in Lite and not in a `<style scoped>` block in the Pro SFC: `modules/germanized/module.php:183` registers only `script-admin.js` and never a stylesheet, so CSS emitted from that SFC would never be enqueued. Lite owns the grid its slot children sit in.

### Testing scope

**Needs verification:**

- **Add Vendor (legacy)** — `admin.php?page=dokan#/vendors` → *Add New Vendor*. All five create-only fields present; markers read `Store Name *`, `Email *`, `Username *`, `Password *`.
- **Required validation** — press *Next* on an empty form: `Password is required` appears in red alongside the others and the form does not advance.
- **Password flow** — type a password, create the vendor, then log in as that vendor (this is the regression that mattered most). Also check *Generate Password* → *Cancel* leaves the input present and empty, *Regenerate* replaces the value, and the eye toggle masks/reveals.
- **Edit Vendor (legacy)** — `#/vendors/<id>`: create-only fields still absent, no required markers, `payment-info edit-mode` present, PayPal in `column` rather than `checkbox-group`.
- **Germanized layout** — module active with the company/VAT/bank fields on: Company Name and Company ID/EUID share a baseline, no gap under Company Name, VAT not stranded in the right lane. Best paired with getdokan/dokan-pro#6127 on the same site.
- **Spacing** — roughly 17px of breathing room under the email toggle and the *Generate Password* button, matching the rest of the form.

**Should be unaffected, worth a smoke pass:**

- The **new React dashboard** (`page=dokan-dashboard#/vendors/create`) has its own implementation and already rendered username, password and banner. It is untouched here, but both paths post to `/dokan/v1/stores`.
- Frontend vendor registration, and vendor creation over the REST API directly.

**Automation gap worth closing:** `tests/pw/tests/e2e/stores/stores.spec.ts:31` — `test.skip('admin can add vendor')` — is precisely the test that would have caught this, and its page object (`storesPage.ts:588`) already drives this form end to end. Unskipping it, minus the now-unnecessary *Generate* click, would turn the manual pass above into a standing regression guard.

### Follow-ups (non-blocking)

1. `#store-password` is now a real `type="password"` sitting immediately after `#user-login` inside the same `<form>` — the shape browsers and password managers classify as a sign-in form. Adding `autocomplete="new-password"` (and `autocomplete="off"` on `#user-login`), as WP core's `user-new.php` does, keeps a saved admin credential from autofilling into a new vendor.
2. The empty-`user_pass` hole still exists server-side for every other caller. `Vendor\Manager::create()` could `unset( $data['user_pass'] )` when it is empty, so the `wp_generate_password()` default applies for all callers rather than just this form.

rajpratham1 (approved):
Looks good. The legacy Add Vendor form now correctly treats the password as required, adds a masked password field with an accessible visibility toggle, properly resets the password state when generation is cancelled, and keeps the vendor ID handling numeric. The styling changes are also scoped to the affected form controls. Approve.

dev-shahed:
This PR does two things that, combined, produce a textbook credential-autofill target inside
**Dokan → Vendors → Add New**:

1. It un-hides `#user-login` (a plain text input) — that field was never rendered before, because
   the `! getId()` bug this PR fixes kept the whole create-only block out of the DOM.
2. It changes the password input from `v-if="showPassword"` + `type="text"` to always-rendered
   `:type="passwordVisible ? 'text' : 'password'"`, whose resting state is `password`.

A text input followed by a `type="password"` input inside one `<form>` is exactly the shape
Chrome's password manager looks for. Neither field carries an `autocomplete` attribute, and
neither does the form.

## Reproduction (clean tab, no prior typing)

1. Log into wp-admin as an administrator whose credentials Chrome has saved for the site.
2. Open **Dokan → Vendors → Add New** (`admin.php?page=dokan#/vendors`).
3. Click the **Username** field.

Observed — the admin's own credentials, filled and highlighted:

```
Username *  [ dokanpwtest         ]   <- the logged-in admin's login, autofill tint
Password *  [ ••••••••••• ]  👁        <- the admin's own saved password, masked
```

Verified in the DOM on both fields:

```
#user-login   :autofill  -> true
#store-password :autofill -> true
```

and once the field is interacted with, the values reach the Vue form state:

```
store.user_login = "dokanpwtest"
store.user_pass  = "dokanpwtest"   (11 chars, never typed by the tester)
```

`store` is the exact object POSTed to `/dokan/v1/stores`.

Chrome keeps autofilled values hidden from scripts until the user interacts, so the first read
can show `value === ""` while the field is visibly filled — read `:autofill`, not `.value`,
when reproducing.

## Why it matters

- The password field is **masked**, so the admin cannot see what is about to be submitted. They
  never typed it and have no prompt telling them it was filled.
- In the observed case the submit is blocked, but only by luck: the autofilled username already
  exists, so `userNameAvailable` is false. **Change the username to a free one and the autofilled
  password stays behind.** The vendor is then created with the administrator's real password —
  and whoever holds that vendor account holds a password that also opens the admin account.
- Any other credential the manager has saved for the domain can be filled instead, including one
  whose username is free on this site, in which case nothing blocks the submit at all.

## Root cause

The PR adopted `type="password"` without the guard WordPress core puts on its own equivalent
form. `wp-admin/user-new.php`:

```php
line 544: <input name="user_login" ... autocomplete="off" ... />
line 603: <input type="password" name="pass1" ... autocomplete="new-password" ... />
line 616: <input type="password" name="pass2" ... autocomplete="new-password" ... />
```

Dokan's own React dashboard (`src/admin/dashboard/pages/vendor-create-edit/form/General.tsx`)
is **not** affected — it renders `type: 'text'` and generates a password on focus, so no
credential pair ever forms. The exposure is specific to this PR.

PR #3406, opened 2026-09-07
fix(product-editor): show the category the vendor picked, not its parent

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

The new product editor discarded a vendor's sub-category and showed the top-level parent instead. Two things were wrong, both in Lite.

**1. The field read the wrong source (`includes/ProductEditor/FormSchema.php`)**

A Dokan product deliberately carries *every ancestor* of the picked category as a `product_cat` term, so the storefront can browse by parent. `resolve_field_value()` populated the Categories field from that raw term list:

```php
$category_options = self::terms_to_async_options( $product->get_category_ids(), 'product_cat' );

if ( ProductCategoryHelper::product_category_selection_is_single() ) {
    return array_slice( $category_options, 0, 1 );
}
```

`terms_to_async_options()` sorts by `name ASC`, so the slice lands on whichever *ancestor* sorts first — never reliably the category the vendor picked. That is why the reporter saw `gadgets` even when the child was listed first in the term array.

The storage was never wrong: `chosen_product_cat` held the picked category correctly the whole time. So the field now reads that instead, and falls back to `generate_chosen_categories()` (deepest term per branch) for products saved outside Dokan that never recorded a selection.

This also corrects **multiple**-select stores, which were listing ancestors the vendor never chose.

**2. Create and update disagreed (`includes/Product/Hooks.php`)**

`sync_category_data_for_commission()` was hooked only on `dokan_product_updated`. A product created through `POST /dokan/v3/products` therefore recorded no chosen category and got no ancestor chain, while the very same payload sent as an update did both:

```
POST /dokan/v3/products  category_ids: [smartwatches]  -> terms [smartwatches],                    chosen_product_cat: (unset)
PUT  /dokan/v3/products/N category_ids: [smartwatches]  -> terms [gadgets, wearables, smartwatches], chosen_product_cat: [smartwatches]
```

Hooking `dokan_new_product_added` as well makes creation behave like an edit. This is the concrete half of the issue's "make the routes agree"; the ancestor-chain storage itself is intentional and is left alone.

**3. Single mode read the wrong term out of the chosen set (`includes/ProductEditor/FormSchema.php`)**

The single-select branch sliced the option list *after* `terms_to_async_options()` re-sorts it by name, so a product carrying more than one chosen category — a store switched from multiple to single, or terms spanning two branches — surfaced the alphabetically-first category rather than the one the vendor picked. That is the same name-sort defect as (1), one layer down. The chosen ids are now narrowed *before* the options are built, matching the legacy editor's `[ reset( $chosen_cat ) ]`.

### Related Pull Request(s)

* None — Pro needs no change. Nothing in `dokan-pro` overrides this field or its save path.

### Closes

* Closes getdokan/dokan-pro#6129

1. Create a nested tree, e.g. `gadgets -> wearables -> smartwatches`.
2. Dokan -> Settings -> Appearance -> **Vendor Product Editor = Latest**; Selling -> **Category style = single**.
3. As a vendor, edit a product in the new editor, pick the grandchild `smartwatches`, fill Title/Price/Description, click **Update Product**.
4. Reload the product.
   * **Expected (this PR):** the field still shows `smartwatches`.
   * **On `develop`:** it shows `gadgets`.
5. Click **Update Product** again without changing anything, reload once more — the sub-category must survive (on `develop` it is now permanently `gadgets`).
6. Set **Category style = multiple** and confirm the field lists only the picked categories, not their ancestors.
7. Multi-chosen single mode: give a product two chosen categories (e.g. save `smartwatches` **and** a top-level `kitchen` under **Category style = multiple**), switch back to **single**, reload the editor. The field must show `smartwatches` (the first picked), not `kitchen` (which merely sorts first by name).

Automated coverage: `tests/php/src/REST/ProductEditorCategoryFieldTest.php` (7 tests, incl. a regression test for the single-mode name-sort in (3) that fails on the first commit). Full Lite suite: `OK (443 tests)` on PHP 8.3.

### Changelog entry

*Fix: the new product editor reverted a selected sub-category to its top-level parent*

**Previous behaviour** — after a vendor picked a child category and saved, reopening the product in the 5.0 editor showed the top-level ancestor instead. Because the field is single-select, the next save wrote that parent back, so a vendor who reopened the product and edited anything unrelated silently lost the sub-category. Marketplaces organised by sub-category ended up with products collected at the top level, affecting storefront browsing and any category-based commission or restriction rules. With multiple selection the field also listed ancestors the vendor never chose. Separately, a product created through `dokan/v3` recorded no chosen category at all, so create and update produced different data for the same vendor action.

**New behaviour** — the Categories field shows exactly what the vendor picked, in both single and multiple mode, and re-saving the form no longer overwrites it. Products saved outside Dokan resolve to the most specific category of their branch. Creating a product now records its chosen category and ancestor chain the same way editing one does.

### Before Changes

Product stores `chosen_product_cat = smartwatches`; terms are the full chain `[gadgets, wearables, smartwatches]`. The editor's Categories field renders **`gadgets`**.

### After Changes

Same product, same data — the field renders **`smartwatches`**, with only `smartwatches` ticked in the category tree. Verified in the browser by swapping only `FormSchema.php` between the two page loads.

https://claude.ai/code/session_01JFhWbDeWfYzgPjgXkGNCv8

### DIFF
--- a/includes/Product/Hooks.php
+++ b/includes/Product/Hooks.php
@@ -36,6 +36,8 @@ class Hooks {
         // Add WooCommerce product brands support.
         add_action( 'dokan_new_product_added', [ $this, 'update_product_brands_by_id' ], 10, 2 );
         add_action( 'dokan_product_updated', [ $this, 'update_product_brands_by_id' ], 10, 2 );
+        // Creating a product has to record its chosen category too, or nothing downstream knows what was picked.
+        add_action( 'dokan_new_product_added', [ $this, 'sync_category_data_for_commission' ] );
         add_action( 'dokan_product_updated', [ $this, 'sync_category_data_for_commission' ] );
         add_action( 'dokan_product_edit_after_pricing_fields', [ $this, 'add_product_brand_template_in_edit_product' ] );
         add_action( 'dokan_new_product_after_product_category', [ $this, 'add_product_brand_template_in_add_product' ] );
@@ -636,7 +638,7 @@ class Hooks {
      *
      * @since 5.0.6
      *
-     * @param int   $product_id   The ID of the product being updated.
+     * @param int   $product_id   The ID of the product being created or updated.
      *
      * @return void
      */
--- a/includes/ProductEditor/FormSchema.php
+++ b/includes/ProductEditor/FormSchema.php
@@ -661,8 +661,7 @@ class FormSchema {
                 'variant'          => 'async_select',
                 'placeholder'      => __( 'Select product categories', 'dokan-lite' ),
                 'value'            => [],
-                // Loaded on demand as a nested tree instead of embedding the whole
-                // category hierarchy in the schema, which bloats memory on large catalogs.
+                // Loaded on demand as a nested tree so the whole category hierarchy never bloats the schema on large catalogs.
                 'api_endpoint'     => '/dokan/v1/products/categories/tree',
                 'tree'             => true,
                 // Honor the admin "single vs. multiple" category selection setting.
@@ -679,8 +678,7 @@ class FormSchema {
                 'variant'          => 'async_select',
                 'placeholder'      => 'on' === $can_create_tags ? __( 'Select tags/Add tags', 'dokan-lite' ) : __( 'Select product tags', 'dokan-lite' ),
                 'value'            => [],
-                // Tags load on demand from WooCommerce core (searchable/paginated) instead of
-                // embedding the whole tag taxonomy in the schema, which can exhaust memory on large stores.
+                // Tags load on demand from WooCommerce core (searchable/paginated) so the whole tag taxonomy never bloats the schema on large stores.
                 'api_endpoint'     => '/wc/v3/products/tags',
                 'creatable'        => 'on' === $can_create_tags,
                 'visibility'       => true,
@@ -1170,12 +1168,10 @@ class FormSchema {
                     $chosen_categories = array_slice( $chosen_categories, 0, 1 );
                 }
 
-                // Async select expects [ { value, label }, ... ] so selected categories render
-                // without the full category tree being embedded in the schema.
+                // Async select expects [ { value, label }, ... ] so selections render without embedding the whole tree.
                 return self::terms_to_async_options( $chosen_categories, 'product_cat' );
             case Elements::TAGS:
-                // Async select expects [ { value, label }, ... ] so selected tags render
-                // without the full tag list being embedded in the schema.
+                // Async select expects [ { value, label }, ... ] so selections render without embedding the whole tag list.
                 return self::terms_to_async_options( $product->get_tag_ids(), 'product_tag' );
             case Elements::BRANDS:
                 if ( method_exists( $product, 'get_brand_ids' ) ) {
--- a/includes/ProductEditor/FormSchema.php
+++ b/includes/ProductEditor/FormSchema.php
@@ -1161,20 +1161,18 @@ class FormSchema {
                 $chosen_categories = ProductCategoryHelper::get_product_chosen_category( $product );
 
                 if ( empty( $chosen_categories ) ) {
-                    // Products saved outside Dokan recorded no selection, so fall back to the deepest term of each branch.
+                    // Products saved outside Dokan recorded no selection, so fall back to the deepest term of each branch (not get_saved_products_category(), which self-heals by writing terms and a read must not).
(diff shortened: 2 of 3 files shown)


MdAsifHossainNadim:
Follow-up commit `f3c5079` from a review round — one real gap plus test/nit fixes. The original fix is sound; the create-parity hook and ancestor-chain storage are correct and untouched.

**Fixed: single mode still let the name sort decide the value.** `resolve_field_value()` sliced the option list *after* `terms_to_async_options()` re-sorts by name, so a product carrying more than one chosen category (store switched multiple→single, or terms spanning two branches) surfaced the alphabetically-first category, not the vendor's pick — the same defect as the reported bug, one layer down. The chosen ids are now narrowed *before* options are built, matching the legacy editor's `[ reset( $chosen_cat ) ]`.

**Also:** documented why `get_saved_products_category()` isn't reused (it self-heals by writing terms — a REST read must not); the re-save test now sends the async-select option objects (exercising `map_ids_to_objects`' unwrap branch, previously untested); fixed a tab-indented array.

**Verification (live, on `dokan-core.test`, served bytecode with opcache reset):**
- Reported bug — save child `smartwatches`, reload editor → shows `smartwatches` (both via the `/dokan/v3/products/N/fields` round trip and visually in the editor UI). Re-saving unchanged keeps it (the progressive silent loss is gone).
- Single-mode name-sort — A/B by reverting only `FormSchema.php`: **with the fix → `smartwatches`** (first picked), **without → `kitchen`** (name-first). The new PHPUnit regression test reproduces this and fails on the first commit.
- No regression — multiple mode lists only the picked categories, not their ancestors; the ancestor chain is still stored (`gadgets → wearables → smartwatches`) and the product is still browsable at `/product-category/gadgets/`.
- Create parity — `POST /dokan/v3/products` records `chosen_product_cat = [smartwatches]` and the full chain, matching update.

**Note on the earlier finding about stale `chosen_product_cat` after a non-Dokan edit:** narrower than I first thought — wp-admin/quick-edit/bulk-edit saves already re-derive the chosen meta through the `is_admin()`-guarded `woocommerce_update_product → update_category_data_for_new_and_update_product` hook. Only non-admin core writes (`wc/v3`, WP-CLI) could drift, and a live scan found 0 drifted products, so I did **not** add a read-time subset guard here — it's outside this issue's scope. Happy to file it as a separate hardening follow-up if wanted.

PHP 8.3 CI is green (443 tests); the PHP 7.4 job failure is the repo-wide wp-env image build (`apt-get install $PHPIZE_DEPS` exit 100), same as `develop`.

MdAsifHossainNadim:
## Verdict: ✅ Approve

The fix is correct end-to-end, the reported bug is gone, and no existing behaviour regresses. The two review items are addressed.

**Checked/fixed in this round**
- **Single-line comments** — collapsed every multiline `//` block the PR touches (both `resolve_field_value` cases, the two async-select field builders, and the three test comments) to single lines, per the project convention. Comment text only, no behaviour change (commit `445df9d`).
- **Tests pass** — PHP 8.3 CI on the fix commit is green: `OK (444 tests, 1608 assertions)`. Locally the class is 7/7 and phpcs is clean on all three files. The regression test genuinely discriminates the fix: it returns the name-first category on the pre-fix code and the picked one after. The PHP 7.4 job is the repo-wide wp-env image build failure (`apt-get install $PHPIZE_DEPS`, exit 100), not a test failure — `develop` fails it identically.

**Why it's correct (verified live on served bytecode)**
- Reported bug: save child `smartwatches`, reload → editor shows `smartwatches`, not `gadgets` (via the `/fields` round trip and visually in the UI); re-saving unchanged keeps it, so the silent progressive loss is gone.
- Single-mode name-sort (the follow-up fix): with the fix → first picked category; A/B by reverting only `FormSchema.php` → the name-first one. Narrowing the ids before `terms_to_async_options()` matches the legacy editor's `[ reset( $chosen_cat ) ]`.
- No regression: multiple mode lists only the picked categories, the ancestor chain is still stored, and the product stays browsable at `/product-category/gadgets/`.
- Create parity: `POST /dokan/v3/products` records `chosen_product_cat` + the full chain, matching update — the `dokan_new_product_added` hook is load-bearing because the admin-guarded `woocommerce_update_product` sync bails on the REST path.

One optional follow-up, out of scope here: a read-time subset guard for `chosen_product_cat` that drifts via a non-admin core write (`wc/v3`/WP-CLI) — near-zero incidence (0 found live), file separately if wanted.

kzamanbd (approved):

PR #3412, opened 2026-09-17
feat(extensions): recommend FlyCRM and drop Texty from the setup wizard

MdAsifHossainNadim:
### All Submissions:

* [x] My code is tested

Promotes [FlyCRM](https://wordpress.org/plugins/flycrm/) in the two places Dokan Lite recommends plugins, and retires Texty from the onboarding wizard.

* **Setup Wizard → Recommended Add-ons** (`includes/Admin/RecommendedPlugins.php`): FlyCRM is added as the **first** entry and **Texty is removed**. Array order is what both the React wizard (`?page=dokan-setup`) and the legacy PHP wizard render, so first-in-array = first card.
* **Admin → Dokan → Extensions** (`includes/Admin/Dashboard/Pages/Extensions.php`): FlyCRM is added with `'position' => 45`, which places it **directly after Dokan PDF Invoice** (40) and before wePos (50). Texty and every other card on this page are untouched.
* **New asset** `assets/images/extensions/addons/flycrm.svg` — the FlyCRM brand mark, sized 53×53 like the sibling add-on icons and shared by both surfaces.
* **Removed asset** `assets/images/texty-logo.png` — the deleted wizard entry held its only reference repo-wide (the surviving Extensions-page Texty card uses `addons/texty.svg`, which stays).
* **E2E** `tests/pw/tests/e2e/tools/toolsPage.ts` — the legacy-wizard helper switched the Texty toggle off by id; that id no longer exists, so the selector now targets `dokan_recommended_flycrm`. (`tools.spec.ts:32` is `test.skip`, so nothing is red today — this just keeps the mirror honest and stops the helper from queueing a real FlyCRM install if the case is ever re-enabled.)

### Related Pull Request(s)

* N/A

### Closes

* Closes getdokan/plugin-internal-tasks#2384

1. `wp option delete dokan_onboarding` and open `wp-admin/admin.php?page=dokan-setup`. Step through to **Recommended Add-ons** → FlyCRM is the first card with its logo and description; Texty is gone.
2. Open `wp-admin/admin.php?page=dokan-dashboard#/extensions` → FlyCRM sits directly after **Dokan PDF Invoice**, with an **Install Free** button; Texty is still listed further down, unchanged.
3. Click **Install Free** on FlyCRM → it installs `flycrm/flycrm.php` from wordpress.org and the button flips to **Installed**; reload the page and the badge persists (server-side `installed` detection).
4. Activate FlyCRM and replay the wizard → the FlyCRM card is filtered out of the wizard, as with every other active recommended plugin.

### Changelog entry

*Add FlyCRM to the recommended add-ons and remove Texty from the setup wizard*

Previously the setup wizard's "Recommended Add-ons" step offered wePos, weMail, WooCommerce Conversion Tracking, Texty, StoreGrowth and weDocs, and the admin Extensions page had no FlyCRM card. Now FlyCRM is the first recommendation in the setup wizard, Texty is no longer recommended there, and the Extensions page lists FlyCRM right after Dokan PDF Invoice. Texty remains available on the Extensions page.

### Before Changes

Setup wizard listed **weMail, WooCommerce Conversion Tracking, Texty, weDocs** (wePos/StoreGrowth hidden because they were already active). The Extensions page went **Dokan PDF Invoice → wePos → Texty → …** with no FlyCRM anywhere.

### After Changes

Setup wizard lists **FlyCRM, weMail, WooCommerce Conversion Tracking, weDocs** — FlyCRM first, no Texty. The Extensions page goes **Dokan PDF Invoice → FlyCRM → wePos → … → Texty → …**, verified live including a real install from wordpress.org.

### Note for reviewers

FlyCRM requires **PHP 8.1+ / WP 6.5+**, while Dokan Lite still supports PHP 7.4. On a PHP < 8.1 site `Plugin_Upgrader` rejects the package, so the first wizard recommendation would fail to install there. This PR deliberately does not add a `requires_php` guard — flagging it in case we want one as a follow-up.

---


MdAsifHossainNadim:
## Verdict: ✅ Approve

Reviewed from `develop`. No blocker — nothing here is a security, data-loss or broken-functionality issue, and both surfaces were exercised against a live site before the PR went up.

### How the change is wired

The two lists are ordered by different mechanisms, so each one is nudged with its own lever rather than with any `flycrm` special case in render code:

* **Setup wizard** renders in plain array order — `SetupWizard::dokan_setup_recommended()` foreach's `RecommendedPlugins::get()`, and `AdminOnboardingController::get_onboarding()` wraps it in `array_values()` before `AddonsScreen.tsx` maps it. So "first card" is simply "first element", and that is where the FlyCRM entry was placed. Texty's entry was deleted from the same array.
* **Extensions page** sorts by `position` (`sortByPosition()` in `RecommendedAddons.tsx`), so FlyCRM carries `'position' => 45`, landing between Dokan PDF Invoice (40) and wePos (50). Half-steps are already in use on this list (weDocs is 75), so the convention holds. Texty and every other card on this page are untouched, as requested.
* **Plugin identity** is `slug`/`wp_org_slug` `flycrm` with `basename` `flycrm/flycrm.php`. Confirmed against the real wordpress.org package rather than assumed — see testing below.
* **Icon**: one `assets/images/extensions/addons/flycrm.svg` (53×53, matching its siblings) feeds both surfaces, instead of the older pattern where weMail/StoreGrowth/Conversion-Tracking each carry a separate root-level PNG.
* **Cleanup**: deleting the wizard entry orphaned `assets/images/texty-logo.png` — it had no other reference anywhere in Lite or Pro — so it goes with it. The Extensions card keeps using `addons/texty.svg`, which stays.
* **E2E**: `toolsPage.ts` mirrored the wizard catalog by checkbox id and switched Texty off by name; that id no longer exists, so it now points at `dokan_recommended_flycrm`. Housekeeping only — `tools.spec.ts:32` is `test.skip`, and the legacy PHP wizard markup does not render in admin at all (`set_setup_wizard_template()` gates the step content on `! is_admin()` while `setup_wizard()` runs on `admin_init`).

### Testing scope

Verified manually against a live WordPress install (PHP 8.4, Dokan Pro active), not just by reading the arrays:

1. **Wizard, replayed from scratch** (`wp option delete dokan_onboarding` → `?page=dokan-setup` → Welcome → Basic Information → Marketplace Goal → Recommended Add-ons): FlyCRM renders **first** with its logo and description; **Texty is gone**. Final order — FlyCRM, weMail, WooCommerce Conversion Tracking, weDocs (wePos and StoreGrowth correctly hidden because they were already active).
2. **Extensions page** (`?page=dokan-dashboard#/extensions`): FlyCRM sits **directly after Dokan PDF Invoice** and before wePos, with the green "Free" badge and an **Install Free** button. Texty is still listed further down, unchanged, alongside every other card.
3. **Real install, not a mock**: clicked **Install Free** → `POST /dokan/v1/admin/extensions/install` with `slug: flycrm` → the plugin downloaded from wordpress.org into `flycrm/flycrm.php` and the button flipped to **Installed**. Reloaded the page and the badge persisted, which exercises the server-side `isset( $installed_plugins['flycrm/flycrm.php'] )` path, so both the slug and the basename are confirmed correct.
4. **Active-plugin filtering** still behaves: the Extensions card keys off *installed*, the wizard keys off *active*, so an installed-but-inactive FlyCRM correctly shows "Installed" on one surface while remaining offered on the other.
5. **Static checks**: `phpcs` clean on both changed PHP files (CI's PHPCS job agrees); `tsc` clean on the Playwright suite. The PHPUnit 7.4 job is red at the **"Start WordPress environment"** step — the known repo-wide 7.4 runner breakage, unrelated to this PR.

Not covered: no automated test was added. The Extensions-page specs assert ordering via the lowest-`position` card (Dokan WPML at 10), so inserting at 45 does not disturb them, and the wizard's own spec is skipped.

### Non-blocking follow-ups

* **FlyCRM requires PHP 8.1**, while Dokan Lite supports 7.4 (`readme.txt`) and every other recommended plugin is at or below that floor. On a 7.4–8.0 site the first, pre-checked recommendation fails at `Plugin_Upgrader::check_package()` and the user sees nothing: the wizard installs through `WC_Install::background_installer()` on `shutdown` after the connection closes, and WC only surfaces a notice for *activation* failures — the install feedback is discarded. Worth either gating the entry on `PHP_VERSION` or teaching the wizard to report per-plugin install errors.
* `RecommendedPlugins::get_enlisted_plugins()` calls `is_callable( $value ) ? $value() : $value` on **every** value, and FlyCRM ships a global `flycrm()` accessor (as do `wepos()` and `wemail()`), so `type`/`slug` hydrate to an object whenever that plugin is active. Harmless in the UI — `get()` drops active plugins immediately after — but consumers of the `dokan_recommended_plugins_list` filter would see an object where a string is documented. Pre-existing.
* The addon catalog is hand-mirrored across `RecommendedPlugins.php`, `Extensions.php` and the Playwright selector map, so every addon change is a two-to-three file edit. A single registry with the wizard reading a projection of it would be the real fix, but that is a refactor of its own.

anik-fahmid:
I've tested the onboarding and extension page; it works as intended.

PR #3414, opened 2026-09-17
fix(product-editor): allow reordering gallery images by drag and drop

akzmoudud:
### All Submissions:

* [ ] My code is tested

In the legacy vendor dashboard product editor, gallery images can be dragged to change their order (jQuery UI sortable in `assets/src/js/product-editor.js`). The new React product editor dropped this: `GalleryImages` could only add and remove images.

This PR adds drag-and-drop reordering to the new editor using the existing shared `SortableList` component (`@dnd-kit`), so no new dependency is added.

- **`ImagePreview.tsx`** takes an optional `onSort` prop. When it is set and there are 2+ images, the previews are rendered through `SortableList` (grid strategy, no extra wrapper element, so the upload "+" button stays in the same row). The featured image (`ImageEdit`) does not pass `onSort` and renders as before.
- **`GalleryImages.tsx`** passes `onSort`, which updates local state and sends the image IDs in the new order — the same way `onRemove` already does. The backend already keeps the order of `gallery_image_ids`, so no PHP change is needed.
- **Input handling:** `ImagePreview` passes its own sensors instead of `SortableList`'s default pointer sensor:
  - Mouse: drag starts after 5px of movement, so the remove (X) button stays clickable.
  - Touch: drag starts after a 250ms long press, so swiping over the images still scrolls the page.
  - Keyboard: arrow keys move the focused image.
- **`index.scss`:** `SortableItem` sets an inline `touch-action: none`, which blocks page scrolling on touch devices. A `.dokan-sortable-image` rule overrides it with `touch-action: manipulation !important` and disables the iOS image callout / text selection so a long press does not open them. The shared `SortableList` / `SortableItem` components are unchanged.
- The preview key now uses the image ID instead of the URL.

### Closes

* Closes getdokan/dokan-pro#6150

1. Log in as a vendor → Vendor Dashboard → Products → edit a product in the **new** product editor.
2. Add 3+ gallery images.
3. Drag the images into a new order, save, and reload — the order should be kept in the editor and on the single product page.
4. After reordering, click X on an image — the correct image should be removed.
5. Tab to an image, press Space, move it with the arrow keys, press Space again — the order should change.
6. On a phone (or touch emulation): swiping over the images scrolls the page; long-pressing an image and dragging reorders it; tapping X removes it.
7. Featured image upload/remove should behave as before.

### Changelog entry

*Gallery images can now be reordered in the new product editor*

Previously, vendors could add and remove gallery images in the new product editor but could not change their order, unlike the legacy editor. Gallery images can now be reordered by drag and drop (mouse, touch long-press, or keyboard).

### Before Changes

Gallery images in the new product editor could not be dragged or reordered.

### After Changes

Gallery images can be dragged into a new order, and the order is saved with the product.

---

Checked: ESLint passes on the changed files and `npm run build` succeeds. Not yet tested manually in a browser or on a touch device. Stylelint reports space indentation on the new SCSS lines; the whole existing file uses spaces, so the new rule matches it.


MdAsifHossainNadim (approved):
## Verdict: ✅ Approve

### The exact issue
The legacy vendor product form lets vendors drag gallery images into a new order (jQuery UI sortable). The new React product editor dropped that: `GalleryImages` → `ImagePreview` only rendered a static list with add/remove, so the order a vendor uploaded in was the order they were stuck with (getdokan/dokan-pro#6150).

### Why the fix is correct
- The previews are wrapped in the existing shared `SortableList` (`@dnd-kit`, already a Lite dependency), so nothing new is pulled in. `SortableList` spreads its props onto `DndContext` *after* its own defaults, so the `sensors` passed from `ImagePreview` genuinely replace the default `PointerSensor` — mouse needs 5px of movement (so the remove button stays clickable), touch needs a 250ms long press (so swiping still scrolls), keyboard uses the sortable coordinate getter.
- `onSort` mirrors the existing `onRemove`: it updates local state and pushes the image IDs in the new order through `onChange`. On the PHP side `PayloadResolver::resolve_images()` assigns positions in array order and `ProductControllerV3` feeds them to `set_gallery_image_ids()` in that same order, so no backend change is needed and the order reaches `_product_image_gallery` intact.
- `isSortable` only kicks in with `onSort` **and** 2+ images, so the featured image (`ImageEdit`) and single-image galleries render through the unchanged plain path. The `.dokan-sortable-image` rule's `!important` correctly beats `SortableItem`'s inline `touch-action: none` (verified via computed style).

### Testing scope
Verified locally (WC 11.0.1, Lite + Pro 5.1.2, PR applied on `develop`, `npm run build`) on a scratch product with three gallery images:

| Check | Result |
|---|---|
| Sortable markers on gallery tiles (`role="button"`, `tabindex="0"`, `aria-roledescription="sortable"`, `cursor-move`, `img[draggable=false]`) | ✅ present; featured image tile has none of them |
| Computed `touch-action` on tiles | ✅ `manipulation` (inline `none` overridden) |
| Mouse drag reorder (tile 1 → tile 3) | ✅ reorders |
| Keyboard: focus tile → Space → ArrowRight → Space | ✅ moves one slot, focus ring visible |
| Save → `POST /dokan/v3/products/{id}` | ✅ 200, `_product_image_gallery` = new order |
| Reload editor / single product page gallery | ✅ new order kept in both |
| Real mouse hover + click on X after a reorder | ✅ removes the tile at its *current* position; remaining tiles still sortable |
| 2 → 1 image | ✅ falls back to the plain (non-sortable) path, X still works |
| Uploader "+" placement, console | ✅ stays in the flex row; no React/dnd warnings |
| ESLint (changed files) / `tsc` | ✅ clean / 964 → 964 (no new errors) |

### Not verified here → QA
**Touch** could not be exercised in this environment (and the PR body says it wasn't tested on a device). `touch-action: manipulation` + long-press activation is a documented dnd-kit pattern, but please confirm on **iOS Safari** and **Android Chrome**: swiping over the tiles scrolls the page, long-press then drag reorders, tapping X removes, and no image callout / text selection appears on long-press.

### Non-blocking notes (follow-ups, not required for this PR)
- `SortableItem` clones `role="button"` + `tabindex` onto a container that holds a real `<button>` (nested interactive, axe will flag it). Shared-component constraint, not introduced here.
- dnd-kit's screen-reader announcements are English-only; `SortableList` already exposes an `accessibility` prop for translated strings.
- A Playwright case (reorder → save → reload) under `tests/pw` would lock this in.
- Stylelint 179 → 182 on `index.scss`, same tab-vs-space class as the rest of the file.

PR #3415, opened 2026-09-17
fix(vendor-dashboard): show prices with the store's number of decimals

akzmoudud:
### All Submissions:

* [x] My code is tested

When WooCommerce → Settings → Products → **Number of decimals** is set to `0`, prices rendered by React on the vendor dashboard (e.g. Subscription Packs) still showed two decimals (`50,000.00$` instead of `50,000$`).

**Root cause** — `formatPrice()` in `src/utilities/Accounting.ts`:

```ts
if ( ! precision ) {
    precision =
        window?.dokanFrontend?.currency.precision ||      // 0 → falsy
        window?.dokanAdminDashboard?.currency.precision;  // undefined on vendor dashboard
}
```

The localized precision (`wc_get_price_decimals()`) is `0`, but `0 || undefined` resolves to `undefined`. `accounting.formatMoney()` then uses its default precision of `2`. Admin React pages weren't affected because there the expression is `undefined || 0`.

**Fix**
- `formatPrice`: fall back only when `precision` is `null`/`undefined`, and use `??` for the localized value.
- `formatNumber`: use `??` for the localized precision as well.

Because `formatPrice`/`PriceHtml` is shared, this fixes every React-rendered price on the frontend vendor dashboard, not only Subscription Packs.

### Related Pull Request(s)

* N/A

### Closes

* Closes getdokan/dokan-pro#6149

1. WooCommerce → Settings → Products → Currency options → set **Number of decimals** to `0` and save.
2. Rebuild assets (`npm run build`).
3. As a vendor, open Vendor Dashboard → Subscription → Subscription Packs.
4. Pack prices show without decimals (e.g. `5$`, `50,000$`).
5. Set **Number of decimals** back to `2` and confirm prices show `5.00$` again.
6. Spot-check other React price displays (e.g. subscription orders, withdraw) on the vendor dashboard and admin dashboard.

### Changelog entry

*Fixed zero-decimal currency setting ignored on vendor dashboard prices*

Previously, prices rendered with React on the vendor dashboard (such as subscription packs) always showed two decimal places when WooCommerce's "Number of decimals" was set to `0`. They now respect the configured number of decimals.

### Before Changes

Subscription pack prices show `5.00$`, `50,000.00$` with "Number of decimals" = 0.

### After Changes

Subscription pack prices show `5$`, `50,000$`, matching the rest of the store.

### PR Self Review Checklist:

* Verified against WooCommerce's `accounting.js` for precision `0` and `2` on both vendor-dashboard and admin-dashboard localized data.
* No caller passes `precision` explicitly, so the only behaviour change is that a configured `0` is honoured.
* `npx eslint` reports only pre-existing `no-console` errors in this file (not on changed lines).


MdAsifHossainNadim (approved):
## Verdict: ✅ Approve

### The exact issue
When WooCommerce → Settings → Products → **Number of decimals** is `0`, every price rendered through React on the vendor dashboard still showed two decimals (`৳ 100.00` instead of `৳ 100`). The report in getdokan/dokan-pro#6149 blames the Subscription Packs page only, but the blast radius is wider: withdraw balance, reverse withdrawal, orders list and products list all had it, because they all go through the shared `formatPrice()` / `<PriceHtml>` in `src/utilities/Accounting.ts`.

### Why the fix is correct
`formatPrice()` resolved the precision with `window.dokanFrontend.currency.precision || window.dokanAdminDashboard.currency.precision`. On the vendor dashboard the first operand is the integer `0` (falsy) and the second is `undefined`, so `undefined` reached WooCommerce's `accounting.formatMoney()`. Its `defaults()` helper replaces any `== null` option with the library default of **2** decimals, which is exactly the symptom.

The PR changes the guard to `precision === null || precision === undefined` and the chain to `??`, so a localized `0` survives and `accounting.checkPrecision(0)` returns `0`. For any precision ≥ 1, `||` and `??` behave identically, so the change is confined to the zero-decimal case. No caller passes `precision` explicitly, `<PriceHtml>` defaults it to `null`, and the `formatNumber()` hunk is behaviour-neutral (accounting.js already defaults plain numbers to 0 decimals), so it's a consistency change only. Pro's `PricingCard` consumes `PriceHtml` via the `@dokan/components` external, so a Lite rebuild alone fixes the Pro page.

### Testing scope
Verified locally on WC 11.0.1 / Dokan Lite + Pro 5.1.2 with the PR diff applied on top of `develop`, `npm run build`, both as an admin-owned store and as a real vendor (`dummy_store_1`):

| Surface (decimals = 0) | Before | After |
|---|---|---|
| Vendor → `/dashboard/new/#/subscription` (packs) | ৳ 100.00 | ৳ 100 |
| Vendor → Withdraw balance | ৳ 745.80 | ৳ 746 |
| Vendor → Reverse Withdrawal | ৳ 153.95 | ৳ 154 |
| Vendor → Orders list | ৳ 71.00 | ৳ 71 |
| Vendor → Products list | ৳ 10.00 | ৳ 10 |
| Admin → Dokan Dashboard | ৳ 374 | ৳ 374 (unchanged, already correct) |

- Decimals back to **2**: every surface shows two decimals again, including real cents (`৳ 47.82`, `৳ 745.80`) — no regression.
- `formatPrice(1234.56)` → `৳ 1,235` at 0 decimals, `৳ 1,234.56` at 2; `formatNumber(50000.5)` → `50,001` / `50,000.50`.
- ESLint: the same 4 pre-existing `no-console` errors, none on changed lines. `tsc` error count unchanged (964 → 964, identical set for this file).

### Non-blocking notes
- The same falsy-fallback pattern still exists for the **thousand separator** in this function (`thousand ||`). WooCommerce allows an empty separator; with it set, PHP renders `৳ 50000.00` while React renders `৳ 50,000.00`. Fine to fold in here or handle as a follow-up.
- Two twins outside this file use `currency?.precision || 2`: `src/components/commission/FixedCommissionInput.tsx:53` and `src/admin/dashboard/pages/setup-guide/Elements/Fields/Commission/CombineInput.tsx:80` (`PriceInput.tsx` already uses `?? 2`). Follow-up candidates.
- Test step 3 should point QA to the React route `/dashboard/new/#/subscription`; the legacy `/dashboard/subscription/` page formats prices with PHP `wc_price()` and never showed the bug.

